PR #19, opened 2017-06-29
alter docker pull tests to use alpine:3.5

j-ogas:


j-ogas:
update: split the docker pull builds test into two.

PR #32, opened 2017-10-11
Make tests pass without Docker

reidpr:


test/chtest/Build
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
+
+# Build an Alpine Linux image roughly following the chroot(2) instructions:
+# https://wiki.alpinelinux.org/wiki/Installing_Alpine_Linux_in_a_chroot
+#
+# We deliberately do not sudo. It's a little rough around the edges, because
+# apk expects root, but it better follows the principle of least principle. We
```
trandles-lanl: "principle of least privilege"

reidpr:
Read-only on the default binds is because there's no need to change them, so declare that fact explicitly.

User's home directory isn't one of the default binds. It's handled separately.

trandles-lanl:
Overall, looks sane to me.  Are there any remaining known-to-fail tests?

reidpr:
No. I try to avoid known-to-fail tests, and if any do crop up, they will `skip`.

I'll make the requested changes, thanks.

PR #38, opened 2017-10-25
Check if a sys_file could be found

mcv21:
Previously, if none of the things tried for sys_file exist, it remains
as None, which results in the rather unhelpful error message:

     File "/test/dev_proc_sys.py", line 20, in <module>
       open(f, "rb").read(1)
   TypeError: expected str, bytes or os.PathLike object, not NoneType

Now we only try non-None paths, and emit a warning if sys_file is None

### DIFF
--- a/test/chtest/dev_proc_sys.py
+++ b/test/chtest/dev_proc_sys.py
@@ -14,16 +14,20 @@ for f in ("/sys/devices/cpu/rdpmc",
       sys_file = f
       break
 
+if sys_file is None:
+   print("WARNING\t No sys_file found to test")
+
 problem_ct = 0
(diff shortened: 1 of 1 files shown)


reidpr:
Thanks Matthew! You're exactly right.

I adapted your patch, which is why it shows as closed instead of merged. Part of that reasoning is that `ERROR`, `RISK`, `SAFE` is a machine-parseable vocabulary, and I wanted the test to fail (`ERROR` and exit non-zero) if the file wasn't found. Skipping the file would result in a pass because `problem_ct` isn't incremented, and the warning would be silent because Bats only prints output for failed tests.

PR #44, opened 2017-10-26
Document the need for perm-test user to be in at least 2 groups; handle the failing case better

mcv21:
Some users (e.g. the default user on a number of "cloud" images) are only in one group. In that case, make-perms-test will fail with the message:

  File "./make-perms-test", line 79, in <module>
    ''' % locals())
TypeError: %d format: a number is required, not NoneType

Which isn't very clear. This PR checks that a second group was found, and errors out if not; it also documents the need to be a member of at least 2 groups

### DIFF
--- a/doc-src/install.rst
+++ b/doc-src/install.rst
@@ -313,7 +313,9 @@ Run
 ---
 
 The run tests require the contents of :code:`$CH_TEST_TARDIR` produced by a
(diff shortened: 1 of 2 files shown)


reidpr:
Interestingly we did check for this condition already but it was too late and the error message was bad.

reidpr:
Merged, thank you!

PR #57, opened 2017-10-28
Use /dev instead of /oldroot for the pivot_root dance.

olifre:
The dance only requires an existing directory in
the image, and /dev is present in any image.
Considered alternatives: /bin might be a symlink to /usr/bin,
other images may miss /usr altogether.

### DIFF
--- a/test/run_uidgid.bats
+++ b/test/run_uidgid.bats
@@ -93,7 +93,7 @@ setup () {
     type=$(fgrep ' / ' /proc/mounts | cut -d' ' -f3)
     opts=$(fgrep ' / ' /proc/mounts | cut -d' ' -f4)
     run ch-run $UID_ARGS $GID_ARGS $CHTEST_IMG -- \
-               /bin/mount -n -o $opts -t $type $dev /oldroot
+               /bin/mount -n -o $opts -t $type $dev /dev
     echo "$output"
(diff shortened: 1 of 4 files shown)


test/run_uidgid.bats
```diff
@@ -93,7 +93,7 @@ setup () {
     type=$(fgrep ' / ' /proc/mounts | cut -d' ' -f3)
     opts=$(fgrep ' / ' /proc/mounts | cut -d' ' -f4)
     run ch-run $UID_ARGS $GID_ARGS $CHTEST_IMG -- \
-               /bin/mount -n -o $opts -t $type $dev /oldroot
```
reidpr: Here /oldroot is used for a different purpose: convenient directory to use for a test. I'll change it to `/mnt/0`.

olifre: Ah, ok, thanks. Yes, I was unsure at this point, but then was convinced that `/dev` is also suitable for testing since the tests did not fail. 
But I agree that `/mnt/0` certainly seems to be the better option here.

reidpr: I had to think a bit too. Normally we wouldn't want to depend on `/mnt/0` existing because only certain `--bind` arguments need it, but in this case we control the image (it's the `chtest` image).

reidpr:
Merged, thank you!

PR #58, opened 2017-10-28
Fix documentation building with >= sphinx 1.3.5.

olifre:
Newer versions of sphinx throw an error:
Could not lex literal_block as "ini". Highlighting skipped.
Older versions just don't highlight the full block.


olifre:
Also see:
https://github.com/sphinx-doc/sphinx/issues/1565
And for even more detail:
https://github.com/sphinx-doc/sphinx/issues/2264

Issue #33, opened 2017-10-19
Zip files and tarballs downloaded from GitHub don't build or pass tests


reidpr:
This is for two reasons:

1. GitHub doesn't include submodules when it makes those `.zip`s, so Bats is missing.

2. `VERSION.full` is missing and can't be built because it's not a Git checkout.

Expected behavior: After unpacking, the `.zip` files build (`make` at top level) and pass tests (`cd test && make test`).

reidpr:
Possible workaround: Attach `.tar.gz` files we build with `make export` to releases using the web interface. But, this is kind of clunky and only solves part of the problem.

reidpr:
OK, here are a couple of options:

**Option A: Manually publish tarballs**

1. Add to the release process: `make export` a tarball and attach it to the release in GitHub.

2. Document that GitHub zip files don't work, and why. (AFAICT we can't disable the zip downloads.)

Advantages:

* Version numbers always contain the Git commit & are the same no matter how the software was obtained.
* No extra steps required to run tests.

Disadvantages:

* Releases are more complex.
* Still offer zip files that don't work.
* Working tarballs only available for released versions.
* Harder to use a separate Bats install.

**Option B: Accept trade-offs to make GitHub zip files work**

1. Change the Makefile to build `VERSION.full` as simply a copy of `VERSION` if not in a Git checkout. This would lose some information (Git commit) but make the software buildable.

2. Change the tests to pick up `bats` from `$PATH`, add `test/bats.src/bin` to `$PATH`, and remove `test/bats` symlink. This will let the tests use either a separately installed Bats (e.g., Bats is in Debian already and could be a `Suggests:` or `Recommends:`) or one installed into the Charliecloud source code.

3. Document that Bats is needed to run tests, and how to get it. 

4. Remove `make export` target.

Advantages:

* Zip files from GitHub now work.
* Duplicate install of Bats no longer needed if it's already available.
* Complex `make export` code goes away.

Disadvantages:

* `ch-run --version` etc. can report different things for the same commit depending on how it was obtained.
* Running tests needs extra steps.

Opinions?

olifre:
[...]

Issue #12, opened 2017-06-05
Test-build sanity check: is the user root?


j-ogas:
Bug source:      /charliecloud/test/common.bash

## Reproduction
Assuming the user successfully compiled Charliecloud:

As root, in _/charliecloud/test/_, run `$ make test-build` 

## Expected Behavior
Compiling and building as root is unsupported; `$ make test-build` should fail.  

## Observed Behavior
The test build runs successfully
** **
## Cause
There is no test in common.bash to determine if the user is root

reidpr:
Per comments in PR #85, this breaks common use cases, so we won't do it.

Issue #11, opened 2017-06-02
Test for compiling charliecloud as root


j-ogas:
Bug source:      /charliecloud/Makefile
Edit: *Added markup syntax, reproduction steps, and expected/observed behavior, and cause*

## Reproduction
After cloning the charliecloud repo, as root in _/charliecloud_, run the following:

1. `$ make` or `make install PREFIX=/local/usr`

## Expected Behavior
Building as root is unsupported; the make should fail.  

## Observed Behavior
Charliecloud compiles successfully
** **
## Cause
The _Makefile_ is missing a test to determine if the user is root.

reidpr:
Let's leave it open until the fix lands on master.

j-ogas:
Update: 

The main Makefile's use of `.PHONY all`makes the prevention of compiling charliecloud as root a little more difficult than expected. 

An option discussed was to include adding the root test in bin/Makefile. The `.PHONY all` line in the bin/Makefile would need to be removed.

reidpr:
Per comments in PR #85, this breaks common use cases, so we won't do it.

PR #59, opened 2017-10-28
Make Charliecloud more friendly for packaging.

olifre:
Up to now, Charliecloud is perfect for the main design goal,
which is installation and usage by users.
For system-wide installation or packaging, several issues occur,
which are resolved by this commit.
Default behaviour is unchanged, but packagers
or system maintainers can specify not to install examples,
tests and documentation.
Also, an extra flag allows to disable setting SUID
during build, which usually runs unprivileged,
but only during install, which runs privileged
(in common packaging environments). 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,14 +51,26 @@ export: VERSION.full
 	rm bats.tar
 	ls -lh charliecloud-$$(cat VERSION.full).tar.gz
 
-BIN=$(PREFIX)/bin
-DOC=$(PREFIX)/share/doc/charliecloud
+# PREFIX is the prefix expected at runtime (usually /usr or /usr/local
+#  for system-wide installations).
+#  More at https://www.gnu.org/prep/standards/html_node/Directory-Variables.html
+# DESTDIR is the installation directory using during make install,
+#  which usually coincides for manual installation,
+#  but is chosen to be a temporary directory in packaging
+#  environments. PREFIX needs to be appended.
+#  More at https://www.gnu.org/prep/standards/html_node/DESTDIR.html
+# Reasoning here: Users performing manual install *have* to specify PREFIX,
+# default is to use that also for DESTDIR.
+# If DESTDIR is provided in addition, we use that for installation.
+INSTALL_PREFIX := $(if $(DESTDIR),$(DESTDIR)/$(PREFIX),$(PREFIX))
+BIN=$(INSTALL_PREFIX)/bin
+DOC=$(INSTALL_PREFIX)/share/doc/charliecloud
 TEST=$(DOC)/test
 .PHONY: install
 install: all
 	@test -n "$(PREFIX)" || \
           (echo "No PREFIX specified. Lasciando ogni speranza." && false)
-	@echo Installing in $(PREFIX)
+	@echo Installing in $(INSTALL_PREFIX)
 #       binaries
 	install -d $(BIN)
 	install -pm 755 -t $(BIN) $$(find bin -type f -executable)
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,19 @@ install: all
 #       binaries
 	install -d $(BIN)
 	install -pm 755 -t $(BIN) $$(find bin -type f -executable)
-	if [ -u bin/ch-run ]; then sudo chmod u+s $(BIN)/ch-run; fi
+	# Install ch-run setuid if either SETUID=yes is specified
+	# or the binary in the build directory is setuid.
+	if [ -n "$(SETUID)" ]; then \
+            if [ $$(id -u) -eq 0 ]; then \
+	        chown root $(BIN)/ch-run; \
+	        chmod u+s $(BIN)/ch-run; \
+	    else \
+	        sudo chown root $(BIN)/ch-run; \
+	        sudo chmod u+s $(BIN)/ch-run; \
+	    fi \
+	elif [ -u bin/ch-run ]; then \
+	    sudo chmod u+s $(BIN)/ch-run; \
+	fi
 	install -pm 644 -t $(BIN) bin/base.sh bin/version.h bin/version.sh
 #       misc "documentation"
 	install -d $(DOC)
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,14 @@ INSTALL_PREFIX := $(if $(DESTDIR),$(DESTDIR)/$(PREFIX),$(PREFIX))
 BIN=$(INSTALL_PREFIX)/bin
 DOC=$(INSTALL_PREFIX)/share/doc/charliecloud
 TEST=$(DOC)/test
+# LIBEXEC_DIR is modeled after FHS 3.0 and
+# https://www.gnu.org/prep/standards/html_node/Directory-Variables.html
+# It contains any executable helpers that are not needed in PATH.
+# Default is libexec/charliecloud which will be preprended with
+# the PREFIX.
+LIBEXEC_DIR ?= libexec/charliecloud
+LIBEXEC_INST=$(INSTALL_PREFIX)/$(LIBEXEC_DIR)
+LIBEXEC_RUN=$(PREFIX)/$(LIBEXEC_DIR)
 .PHONY: install
 install: all
 	@test -n "$(PREFIX)" || \
@@ -74,6 +82,10 @@ install: all
 #       binaries
 	install -d $(BIN)
 	install -pm 755 -t $(BIN) $$(find bin -type f -executable)
+	# Modify scripts to relate to new libexec location
+	for scriptfile in $$(find bin -type f -executable -printf "%f\n"); do \
+	    sed -i "s#^LIBEXEC=.*#LIBEXEC=$(LIBEXEC_RUN)#" $(BIN)/$${scriptfile}; \
+	done
 	# Install ch-run setuid if either SETUID=yes is specified
 	# or the binary in the build directory is setuid.
 	if [ -n "$(SETUID)" ]; then \
@@ -87,7 +99,10 @@ install: all
 	elif [ -u bin/ch-run ]; then \
 	    sudo chmod u+s $(BIN)/ch-run; \
 	fi
-	install -pm 644 -t $(BIN) bin/base.sh bin/version.h bin/version.sh
+#       executable helpers
+	install -d $(LIBEXEC_INST)
+	install -pm 644 -t $(LIBEXEC_INST) bin/base.sh bin/version.sh
+	sed -i "s#^LIBEXEC=.*#LIBEXEC=$(LIBEXEC_RUN)#" $(LIBEXEC_INST)/base.sh
 #       misc "documentation"
 	install -d $(DOC)
 	install -pm 644 -t $(DOC) COPYRIGHT LICENSE README.rst
--- a/bin/base.sh
+++ b/bin/base.sh
@@ -1,6 +1,7 @@
 CH_BIN="$(cd "$(dirname "$0")" && pwd)"
 
-. "$CH_BIN/version.sh"
+LIBEXEC="$(cd "$(dirname "$0")" && pwd)"
+. ${LIBEXEC}/version.sh
 
 # Do we need sudo to run docker?
 if ( docker info > /dev/null 2>&1 ); then
--- a/bin/ch-build
+++ b/bin/ch-build
@@ -1,6 +1,7 @@
 #!/bin/sh
 
-. $(dirname "$0")/base.sh
+LIBEXEC="$(cd "$(dirname "$0")" && pwd)"
+. ${LIBEXEC}/base.sh
 
 usage () {
     cat 1>&2 <<EOF
--- a/bin/Makefile
+++ b/bin/Makefile
@@ -16,6 +16,12 @@ ifdef SETUID
 export CFLAGS += -DSETUID
 all: setuid
 setuid: ch-run
-	sudo chown root $<
-	sudo chmod u+s $<
+	# Use sudo only if not already root.
+	if [ $$(id -u) -eq 0 ]; then \
+	    chown root $<; \
+	    chmod u+s $<; \
+	elif ( command -v sudo >/dev/null 2>&1 && sudo -v >/dev/null 2>&1 ); then \
+	    sudo chown root $<; \
+	    sudo chmod u+s $<; \
+	fi
 endif
--- a/bin/ch-build2dir
+++ b/bin/ch-build2dir
@@ -1,6 +1,7 @@
 #!/bin/sh
 
-. $(dirname "$0")/base.sh
+LIBEXEC="$(cd "$(dirname "$0")" && pwd)"
+. ${LIBEXEC}/base.sh
 
 set -e
 
--- a/bin/ch-docker-run
+++ b/bin/ch-docker-run
@@ -2,7 +2,8 @@
 
 # bash is needed for arrays.
 
-. $(dirname "$0")/base.sh
+LIBEXEC="$(cd "$(dirname "$0")" && pwd)"
+. ${LIBEXEC}/base.sh
 
 usage () {
     cat 1>&2 <<EOF
--- a/bin/ch-docker2tar
+++ b/bin/ch-docker2tar
@@ -1,6 +1,7 @@
 #!/bin/sh
 
-. $(dirname "$0")/base.sh
+LIBEXEC="$(cd "$(dirname "$0")" && pwd)"
+. ${LIBEXEC}/base.sh
 
 set -e
 #set -x
--- a/bin/ch-tar2dir
+++ b/bin/ch-tar2dir
@@ -1,6 +1,7 @@
 #!/bin/sh
 
-. $(dirname "$0")/base.sh
+LIBEXEC="$(cd "$(dirname "$0")" && pwd)"
+. ${LIBEXEC}/base.sh
 
 usage () {
     cat 1>&2 <<EOF


olifre:
Just to clarify: Alternatively, package maintainers could remove the files they don't want installed explicitly (thus duplicating some of the install-code in the packaging code, which means additional maintenance effort). So this is mostly a "convenience thing". 
The only real "blocker" for most build systems is the execution of "sudo" in case of "setuid"-build, since the build runs unattended and unprivileged. 
Of course, since Charliecloud will likely only be packaged for distros which are sufficiently modern to not need the "setuid" feature, also this might not be a real blocker. 

So if you are unhappy with these modifications, let me know - it's not critical, but just a bit more inconvenient to not have them (and maybe there are even better ways).

olifre:
Ok, the file renaming (see latest commit) is also a blocker for packaging, since putting `version.h`, `version.sh` and `base.sh` in `/usr/bin/` is not really a good thing to do. 
With the renaming to include the `ch-` prefix, and dropping installation of `version.h` (which seems not really needed at runtime), that should not be an issue.

reidpr:
OK, excellent. I think exactly what to do will require some discussion, but I agree the current install procedure is sub-optimal and I think it's great that someone is trying to improve it. @mej has suggested moving to autoconf, but I'm not yet convinced that's worth the overhead.

Let me outline the concerns as I understand them so we can work to agreement on each.

**1a. Packagers might not want to install examples,**

The reasoning for installing them is so users can see various ways to use Charliecloud.

**1b. ... tests,**

We actually provide pre-built test suite tarballs at our site and encourage users to run the tests, though I'm not sure they do. It's also helpful for sysadmins to be able to log in as a user and run the tests.

**1c. documentation.**

I believe this is fairly limited at the moment, isn't it? I'm hesitant to separate `COPYRIGHT` and `LICENSE` from the product since those are important legal documents. `README.rst` is indeed documentation, but it's short. We don't install the HTML documentation currently, the reasoning that users can just go read it on the web.

For these first three, the reasoning is above. However, I'm not very familiar with packaging as a developer or packager, just a user. So pointers to best practices or whatnot that support the proposals not to include them would be very helpful. I'm not yet convinced that an option to omit these is really needed, but I'm open to argument.

**2. There should be an extra flag to disable setuid builds.**

This I don't quite follow yet — the current behavior is that you get a non-setuid binary unless you specify `make SETUID=yes`, which seems like it should cover your use case, but I must be missing something. Can you clarify?

FWIW we do have issues to error out during build (#11) and test (#12) if privileged.

The setuid builds are only sort of supported. We want the tests to pass, and not make any security messes, but with minimal effort beyond that. That said, if someone is depending on the feature for real users, we can revisit.

**3. `version.h` doesn't need to be installed.**

I agree.

**4. `version.sh` and `base.sh` don't belong in `/.../bin`.**

I agree. I'm happy with a rename and leaving them in `bin`, or if there's a better place to put them I'm open to that too.

The reason they are in `bin` is that it's easy to find them whether running out of a Git working directory or installed, if they are just in the same directory as the scripts.

olifre:
Thanks for opening and agreeing to the discussion! :-)

I think your summary is very nice, so let's go through the points one by one. 



> 1a. Packagers might not want to install examples,

Agreed. I am slightly biased with packaging, of course - I am using Gentoo Linux, and it's also the distro where I am most experienced in packaging (albeit still far from an expert). 
On Gentoo, it's common practice to let the user (or rather, the admin / machine owner) decide via flags what a package should install. In this case, it's helpful to be easily able to disable features that might be unwanted - for example, the "examples" are perfect to learn how to use Charliecloud, but are not needed in an environment where Charliecloud is "just used", for example, a worker node in a cluster (thinking of the approach that e.g. HTCondor fires up the containers, and the users actually never see them). 
This is of course a very different usecase than the original design goal - hence, I left the default to install the examples, but added the possibility to turn that off. 
For what it's worth, for Debian and RedHat, the practice appears to be to "build" and "install" everything, including the examples, to a temporary directory, and then at the end of the install procedure cut the full install tree into separate packages. This might mean, for instance, that the examples end up in a package called "charliecloud-doc" which may contain documentation, examples and other helpful things. 
So for those distros, "installing" the examples by default is likely still the best thing to do. They just group all the files into separate packages at the end. 
So I would be fine with either solution:
1. adding the flag (and leaving it on by default)
2. not adding the flag (then for Gentoo, a simple `rm -rf $(PREFIX)/share/doc/charliecloud/examples` in the packaging code will still allow to add a flag to packaging). Since Gentoo is almost the only distro allowing such configuration, maybe that's the easiest choice (usually, they use autoconf or cmake to enable / disable example installation, though). 


> 1b. ... tests,

Here, there appear to be a lot of wild differences between the distros. 
AFAIK, RedHat does not run the tests of a package, nor does it install them (but I'm no RedHat expert, so I might be wrong). 
Debian has it's own test suite, which might be used to invoke the tests which are part of the package. However, this is only optionally done for "source packages", the binary packages which are actually distributed and installed do not contain any tests. 
On Gentoo, a "test" flag may be specified which at post-compile-time runs the tests of the package. 
However, again, the tests are not installed. 

So the overall approach is, the tests run once during packaging (if at all), but they are never installed, assuming that if they run in the build environment, things must also work out later on the user's machine. 
I'm not sure that's the best approach, but it's common practice. 

> 1c. documentation.

You are right, my bad. `COPYRIGHT, `LICENSE` and `README.rst` should indeed be installed, some distros may want to split `README.rst` in a separate package (e.g. charliecloud-doc). 
However, it might still be helpful / wanted by some distros to ship the full HTML help pages. 
But in that case, it's fine if those distros just do `make -C doc-src` and install `doc` after, so I think that my flag `INSTALL_DOC` can actually be removed. 



> 2. There should be an extra flag to disable setuid builds.

I think here we simply misunderstood each other. 
The problem is that usually, "building for a distro" means that the actual `make` runs in an unattended sandbox as unprivileged user. For that reason, `sudo chown` and `sudo chmod` will fail, and also a `chown` or `chmod` would fail _during build_. 
However, the `make install` step is usually executed as root, so here `chown` and `chmod` would be fine. 
If you look at my changeset, that's the main thing my `INSTALL_FOR_PACKAGING` flag is doing: Shifting the `chmod` and `chown` to the install phase. 
Of course, that means that the "built binaries" can not be sensibly used, since they do not yet have the suid bit set. 

Rethinking about it, I do not think this is so much of an issue anymore. Any distro Charliecloud will be packaged for should be sufficiently recent enough to facility non-setuid mode, unless somebody really wants to still use something like SL 6 (RHEL 6) on actual hardware. So maybe this change is naught. 

On (3) we already agreed, on (4) I am also unsure about a better place to put them. Since `base.sh` and `version.sh` are actually executed, `/usr/bin` is somewhat right - another option would be to install them to `/usr/libexec/charliecloud/`, since they are somewhat "executable helpers", and then they are also hidden from the regular path. Maybe this is the nicer solution in the long run, but would need a bit more adaption of the Makefiles and shell scripts.

reidpr:
What if we modify this PR to do the following:

1. Keep the current behavior of installing everything, but document the question and reasoning above as needed in a new issue, so it's easy to revisit later.

2. Change `sudo` failures at `make SETUID=yes` from errors to warnings. This will retain the current behavior (setuid builds work in source directory if you have `sudo`) but not break such a build in an unprivileged environment.

3. Change the `install` target to take a `SETUID` variable, like `all`, and `chown(1)` & `chmod(2)` if set, instead checking for an existing setuid bit. (This will require modifying the Travis tests and documentation a bit, I think.)

4. Don't install `version.h` any more (as PR does currently).

5. Rename the two `.sh` files as PR does currently.

olifre:
Sounds good - I'll give it a go later this week, and in parallel develop a package for Gentoo to "see" how well it works out. For the beginning, this would be a "personal" package for me, in my (public, but inofficial) Gentoo repository. I would aim at getting that into Gentoo proper at a later stage. 

Depending on our needs, I may also do packaging for Debian and RedHat - but only in a basic way on the free openSUSE Build Service (which means "inoffical, but usable"). It's better if at some point somebody who is more expert for those distros picks that up (but at least that would give a starting point), since the review procedures for packages are delicate and very different for all distros.

mcv21:
I would agree that autoconf seems like way overkill here. From a Debian packaging point of view, you'd want to try and be a bit more FHS-compliant (which I think would put some of the executables into `/usr/lib/charliecloud` or similar)

olifre:
> Change `sudo` failures at `make SETUID=yes` from errors to warnings.

Thinking about the way to actually do that, I am unsure how. 
`sudo` will prompt about a password (unless I specify `-n` which is an issue with the user-triggered builds). 
So if run was an unprivileged user who is not in `sudoers`, as is the case in build environments, the build would just "hang" waiting for input. Also, attempting to run `sudo` when not being in `sudoers`that usually triggers a "user not in sudoers" action (such as a mail). 
Am I missing some better way to do that? 



> From a Debian packaging point of view, you'd want to try and be a bit more FHS-compliant

@mcv21 Good point. That should be the case for the `ch-version.sh` and `ch-base.sh` files. 
I'm a bit unsure from FHS whether `/usr/lib/charliecloud` or `/usr/libexec/charliecloud` is more correct, though.

olifre:
> if ( command -v sudo >/dev/null 2>&1 && sudo -v >/dev/null 2>&1 ); then

This approach indeed works unless the user is in `sudoers` - which is of course not the case for build system's unprivleged users. Nice idea!




> An behavior I believe it important to preserve is that running from the source directory still works

Agreed - after all, that's one of the main design goals, to have charliecloud usable by "users" without installation. I'll give the moving a go and see how I can easily do it without breaking anything (I'm also a bit slower at the moment since there are national holidays around here).

olifre:
Here's my first go at it for today. I did not yet test whether the tests still work, though (let's see how travis does). Maybe, some more changes are needed in this regard. 

The first two commits take care of the discussed points: Make setuid root during build non-fatal and allow to trigger that during install with a `SETUID` variable, and install the helper-scripts to `$(PREFIX)/libexec/charliecloud/`. 

However, several more details were needed: 

1. In packaging, things are first installed into a temporary install directory, and only later moved to the real filesystem. 
For this, autoconf and cmake offer `prefix` and `destdir`. The variable `prefix` specifies where the package should expect to be at runtime, `destdir` is the actual temporary install location. 
I implemented the following logic:
- If `destdir` is not set, only `prefix` will be used, nothing changes.
- If `destdir` is set, it will be used as installation prefix instead. However, `prefix` will still be used to set the new `$(PREFIX)/libexec/charliecloud/` path which is needed to find `base.sh` and `version.sh` later at runtime. 

This matches what autoconf does as far as I know, and I think the same is true for cmake. 

2. `sudo` is now only attempted to use if `id -u` does not return "0". 
I hit this as being an issue when building the Gentoo package. 
`sudo` does some extra things, for example, it may play within `/root/.ssh/` and setup an `ssh-agent` session. Gentoo's sandbox during packaging catches this and prevents writing there, and throws a nice colourful error message that the package tried to modify the filesystem directly. 
I think similar behaviour can be expected for other distros. 

You can find my first well-working go at a Charliecloud package for Gentoo here:
https://github.com/olifre/olifre-portage/blob/master/sys-cluster/charliecloud/charliecloud-9999.ebuild
It does not run the tests yet, but everything else works as expected.

reidpr:
Great!

I might not be able to look at it today, and then I'm out of the office for a bit unfortunately.

Regarding the tests, you might want to rebase onto current `master`, since one of the tests started breaking Travis, and I disabled it earlier today. Side benefit is that the tests are way faster too.

olifre:
> I might not be able to look at it today, and then I'm out of the office for a bit unfortunately.

No problem, take your time!

Yes, I already rebased to master (just in time), but found another small mistake I did - I fixed that just now, travis tests should work now. If not, I'll have another look. 

I'll also add running the tests to the Gentoo package and see if I hit another block (but I don't expect to).

olifre:
> I'll also add running the tests to the Gentoo package and see if I hit another block (but I don't expect to).

I should not have said that... It seems Gentoo's sandbox (which is also active in the test phase) is effectively preventing user namespaces from working - `ch-run` complains it can not open `/proc/self/uid_map` for writing, that's likely the sandboxes doing. 
So I'll leave the test-phase out from the Gentoo package for now and focus on creating basic packages for RedHat- and Debian-based distros.

olifre:
A small update: 
Here is now a working RPM build!
https://build.opensuse.org/package/show/home:ubn:charliecloud/charliecloud-0.2.3-pre
I'm sure a RedHat expert could do better, but it's a starting point (based on current master with this PR applied).
Since the interface might be a bit confusing, the actual CentOS 7 repo belonging to that is here:
https://download.opensuse.org/repositories/home:/ubn:/charliecloud/CentOS_7/
I'm planning to do some basic Debian packaging next.

Makefile
```diff
@@ -51,19 +51,40 @@ export: VERSION.full
 	rm bats.tar
 	ls -lh charliecloud-$$(cat VERSION.full).tar.gz
 
-BIN=$(PREFIX)/bin
-DOC=$(PREFIX)/share/doc/charliecloud
+INSTALL_PREFIX := $(if $(DESTDIR),$(DESTDIR),$(PREFIX))
+BIN=$(INSTALL_PREFIX)/bin
+LIBEXEC=$(INSTALL_PREFIX)/libexec/charliecloud
+LIBEXEC_RUN=$(PREFIX)/libexec/charliecloud
+DOC=$(INSTALL_PREFIX)/share/doc/charliecloud
```
reidpr: Can you add a comment documenting all these variables, for folks with a shaky understanding of them like myself?

This could be as simple as a link to someone else's write-up.

Makefile
```diff
@@ -51,19 +51,40 @@ export: VERSION.full
...
+	@echo Installing in $(INSTALL_PREFIX)
 #       binaries
 	install -d $(BIN)
 	install -pm 755 -t $(BIN) $$(find bin -type f -executable)
-	if [ -u bin/ch-run ]; then sudo chmod u+s $(BIN)/ch-run; fi
-	install -pm 644 -t $(BIN) bin/base.sh bin/version.h bin/version.sh
+	# Modify scripts to relate to new libexec location
+	for scriptfile in $$(find bin -type f -executable -printf "%f\n"); do \
+	    sed -i "s#^libexec_base=.*#libexec_base=$(LIBEXEC_RUN)#" $(BIN)/$${scriptfile}; \
+	done
```
reidpr: It would be nice if we didn't have to patch the scripts at install time. Is this a best practice? Are there alternatives?

Makefile
```diff
@@ -51,19 +51,40 @@ export: VERSION.full
...
+	# SETUID specified now?
```
reidpr: Can you elaborate a little to describe the logic? Perhaps something like: "Install ch-run setuid if either SETUID=yes is specified or the binary in the build directory is setuid."

bin/base.sh
```diff
@@ -1,6 +1,12 @@
-CH_BIN="$(cd "$(dirname "$0")" && pwd)"
+libexec_base=$(dirname "$0")
```
reidpr: Can we just use the `libexec_base` that's computed by the script?

bin/base.sh
```diff
@@ -1,6 +1,12 @@
...
+if [ -e ${libexec_base}/version.sh ]; then
+    . ${libexec_base}/version.sh
```
reidpr: I'm comfortable just sourcing it, without the error checking, since it shouldn't fail unless the install is wrong (i.e., users shouldn't see this and can't fix the error anyway) and I believe the scripts are `set -e`.

bin/ch-build
```diff
@@ -1,6 +1,12 @@
 #!/bin/sh
 
-. $(dirname "$0")/base.sh
+libexec_base=$(dirname "$0")
```
reidpr: * Can we call it `LIBEXEC` for shortness and consistency with upper-case variables elsewhere in the scripts?
* Let's use the longer voodoo currently used for `CH_BIN`, since it's more robust to symlinks and other stuff I've forgotten.

reidpr:
Thanks Oliver!

I put in a first-class nit-picky line-by-line review. Let me know what you think.

Thanks,
Reid

olifre:
> I put in a first-class nit-picky line-by-line review. Let me know what you think.

Thanks for the detailed review! I will go through the points step by step, but may defer some until tomorrow.

Makefile
olifre: I'm not aware of good alternatives. Most packages avoid this issue by not being runnable from the build location, so they directly store the libexec-path inside the scripts - or, they use autotools / cmake, have something like `@LIBEXEC@` in the files and let autotools / cmake do the replacing of that placeholder. 
If you (or anybody else) finds something nicer, I'm all open ears - I'm also not really a fan of using `sed` here ;-).

Makefile
olifre: Yep, I'll give that a go. The naming `PREFIX` and `DESTDIR` are modelled close to what they mean for autotools, I'll see if I can find a nice writeup, and otherwise cook up something compact.

bin/base.sh
olifre: You mean by the script(s) including `base.sh`? Yes, that certainly works. I only wanted to be on the safe side, in case `base.sh` is ever included by anything else. But that's extremely unlikely to happen (since this "something else" in any case needs the `libexec` path to find `base.sh`). I'll drop it, then.

bin/ch-build
olifre: Both good points, I'll adapt!

bin/base.sh
olifre: Indeed. This allows to simplify the code significantly.

Makefile
mcv21: For reference, Debian's interpretation of FHS is to use `lib` not `libexec` (I think FHS 2.3 doesn't have `/usr/libexec`)

olifre: @mcv21 Thanks, indeed! Yes, it's only an (optional) part of FHS 3.0 I believe. 
It's already adopted on (at least) Gentoo and RedHat (deducing from a CentOS 7.4 system). 
So I guess there are two options now:

- Use `libexec` here, and change things in the Debian packaging code to use `lib`. 
- Default to `libexec` and add another variable to `make install` which allows to change that to something else (so Debian packaging code could specify `lib`). 

Do you have a preference?

mcv21: Something that the build process could override on the command line would be easiest, but `dh-install` can move things around if necessary.

olifre: Ok - then I think I'd go with the environment variable. 
The problem with `dh-install` would be that it would not only have to move things around, but then (again) have to patch the script-files to refer to the different libexec-location to find `base.sh`.

mcv21: Good point! Sorry, should have thought of that.

mcv21: ...though thinking about this, debian builds work (conventionally) by setting prefix to debian/tmp and then fishing the build results out of there...

olifre: > ...though thinking about this, debian builds work (conventionally) by setting prefix to debian/tmp and then fishing the build results out of there...

Are you sure they set `prefix`? As I know it from Gentoo and RedHat, they usually use `destdir` for the temporary install location, but specify `prefix` already to the expected future runtime location (i.e. as if things would be installed on the real filesystem). 
This allows to use `prefix` to derive things as we do here ("prepare for run-time location", i.e. put paths in scripts), while `destdir` would then be the build-time sandbox (`debian/tmp`) to which the files are actually copied. 

In any case, I guess I should do my attempt at Debian packaging first before finalizing this PR to be sure - up to now, I never did a Debian package from scratch, but only modified existing ones.

olifre: Sounds good. I'll start on debian packaging tonight, let's see how far I get and whether I note other changes / subleties from this.

Makefile
olifre: I also gave it a thought again and remembered another approach, I'll mention it just for completeness, since it does not offer significant advantages. 
What some programs do, is to have an extra binary (such as `root-config`, `geant4-config`) installed in the path which answers such questions for paths at runtime, i.e. `root-config --exec-prefix`, or also version information, such as `root-config --version`. The way that binary is generated usually involves the same technique, though: autotools or cmake replace macros in some header file, which is then included in the compiled binary. That's commonly used for finding libraries and `LDFLAGS` and things like that (alternatively, `pkgconfig` is used). 
So I don't see any obvious advantage to that - the only gain would be there is one place where paths are hardcoded at install-time, and not each script separately.

Makefile
olifre: I'm almost through with having a working (but not yet finished) debian package - and at least I can confirm they pass `DESTDIR` to `make` in the `install` step, as all other distros do in packaging. 
So at least there are no new surprises on this end.

olifre:
Since I'll not get much farther today, a short update. 
There are now building and working debian packages here (this currently uses a patched snapshot of git master):
https://build.opensuse.org/package/show/home:ubn:charliecloud/charliecloud-0.2.3-pre

The first resulting package is `charliecloud`, which still shows a few Debian QA check warnings (mainly "missing manpages for the binaries"). I think manpages could be done in a separate issue (to enter Debian main, that would be a requirement). 
The second resulting package is `charliecloud-doc` which contains the HTML documentation.
Sadly, there are several Debian QA check warnings and errors for that, which I did not fix yet (they complain that jquery is bundled in by sphinx, and the usage of mathjax from a CDN triggers an error due to a potential "privacy violation"). 

Luckily, all these issues are things that (I believe) I should fix in the Debian packaging (especially since they affect no other distro). So inside Charliecloud, no new requirements arise (apart from the manpages). 

So my next step for tomorrow is to finally attack the PR, include all comments, include the `libexec` configuration (which is already part of the patched source used for the package) and finally retest that the packaging scaffolding still works.

olifre:
Done!
I believe all changes are in, and I rebased to master.
I essentially rewrote all commits of this PR to also have a more consistent commit history. 
Retrying to build the packages for CentOS, Debian, Ubuntu and Gentoo now works immediately without any patching (albeit there are still QA warnings on Debian, but to fix that, I just need to improve the package itself). 

There's one subtle change fixing an issue in the old PR's code:
`DESTDIR` actually specifies the install location's root, i.e. the `PREFIX` still needs to be appended. 
That's fixed now. The new commits also add some detailed comments with links to the GNU coding standards, which, in the `Managing Releases` section, is pretty close to what we do now. And since autotools, cmake and friends are close to the GNU standards, that's good ;-). 
Of course, GNU specifies a lot more functionality and extras, but we don't need the complexity of a configure step here.

olifre:
Mhhm. Travis seems partially unhappy, but it seems that the reason is some issue with the alpine CDN, which I can't reproduce locally:
```
# fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/main/x86_64/APKINDEX.tar.gz
# ERROR: http://dl-cdn.alpinelinux.org/alpine/v3.6/main: IO ERROR
# WARNING: Ignoring APKINDEX.84815163.tar.gz: No such file or directory
```

reidpr:
> some issue with the alpine CDN, which I can't reproduce locally

I've seen this a lot too. It seems to simply be an intermittent failure. To work around it I've been just restarting the failed jobs. I don't have a better solution at the moment.

reidpr:
Merged, thank you! This is great.

Note: if best practice is to maintain packaging code in this upstream repository, I'm happy to do that.

olifre:
Thanks for merging!


> Note: if best practice is to maintain packaging code in this upstream repository, I'm happy to do that.

I'm really unsure about that. 
Some projects do it, most don't - in any case, for most projects that do, it seems the actual code used by distro maintainers is different from that living in the repositories of upstream in the cases I looked at. In some cases it might be inspired by the code from the project, though. 
So at least it's no widespread practice, and I'm not aware of a best upstream practice. 

At least for Gentoo, I'd personally try to get the package in the distro. 
For RedHat / Debian, I don't feel expert enough (yet) to enter the distro's package review process, though (also, I'm not a daily user of those distros myself).

mcv21:
My other hat is a Debian developer, so... :)

Mileage varies on this - if you're going to be happy to take patches to the debian packaging stuff from whoever ends up maintaining this package in Debian, then it's fine to have it in your repo (and, personally, I think neater); if you expect the Debian maintainer to make different packaging decisions then they may end up maintaining it separately.

I think it's probably best to have some Debian packaging in your source for now (to give a Debian person something to start from), and see how it goes in future.

olifre:
> My other hat is a Debian developer, so... :)

Perfect :-). 

Ok, then I would suggest I bring my packaging code at https://build.opensuse.org/package/show/home:ubn:charliecloud/charliecloud-0.2.3-pre a bit more up to QA standards (especially the doc package still shows some errors, related to sphinx embedding CDN links and bundling jquery mainly) and then open a new PR with that. 

We should then probably do the same for RedHat, I guess, for consistency? 

For Gentoo it's (hopefully) not worthwhile, since I'll try to get it into the distro in the very near future. 
But maybe, after it's accepted, we should add it just for consistency.

olifre:
@mcv21 A small update. My Debian packages at:
https://build.opensuse.org/package/show/home:ubn:charliecloud/charliecloud-0.2.3-pre
now are fine with `lintian`, it only complains that there is only one debian changelog entry (well, it's the initial package version...) and that the manpages are missing. 

From this, I have two questions:
- Are manpages are hard requirement? If so, we should think about adding these (or a way to generate them) in a new issue. 
- To get the package into Debian, I guess a Debian user, or even better a Debian developer, is better suited than me. @mcv21 : Would you be interested in taking that up? The main work (writing the packaging code) seems done, so "only" a review and the "bureaucratic" part would remain.

reidpr:
> if you're going to be happy to take patches to the debian packaging stuff from whoever ends up maintaining this package in Debian, then it's fine to have it in your repo (and, personally, I think neater)

My inclination is to treat the packaging code as a "contrib" area and merge any PRs with only cursory review, with style consistent with the distro's preferences. I believe this makes me "happy to take patches" in the above sense.

Brief READMEs explaining what's going on for each packaging directory would be very helpful, in particular to explain who to contact with questions and who I should accept PRs from.

Also personally I think doing it this way would be nice because it would teach me how packaging works.

> Are manpages are hard requirement? If so, we should think about adding these (or a way to generate them, e.g. with help2man) in a new issue.

I'd love to see man pages in Charliecloud.

I would like to be able to include them in the web documentation as well. For example, we currently have [this page](https://hpc.github.io/charliecloud/script-help.html) that's currently generated by running `--help` on all the scripts, but it might be better to put man pages in there instead.

olifre:
Agreed, sounds like a good idea. 
Another positive effect from having these in the tree here (in a subdirectory `contrib/packaging`? or directly `packaging`?) would be that it might attract distro maintainers. 

For the Debian package, I'd still wait for @mcv21 's response, if he agrees to take this over (and get it into Debian proper), I'm happy to hand the code over to him (and he may want to do a small review of my code before submitting a PR here). 

For the `spec`-file (RedHat-packaging) I can go over my code once more, run `rpmlint` on it, and submit a PR if you like (it's rather short in any case). 
The instructions are rather simple, if on an RPM platform and build essentials are installed, it's just a matter of executing `rpmbuild -ba charliecloud.spec` to build the package. 

For the Gentoo ebuild, I see two options:
- wait for the 0.2.3 release of Charliecloud, get it into Gentoo proper (which may lead to some small changes as required by the full-time gentoo devs), and only then add it to the Charliecloud repo.
- add it to Charliecloud already now and adapt it later if needed. 

One general issue right now is that all packaging is hit by #33 , i.e. they can not use the github release packages as they are. So it'd be best to fix this first (although I don't have a clean idea how) or describe it in the appropriate README files. However, to get the package accepted in a distro, it's usually also a requirement that there are upstream tarballs which can be built. 

The only exception to that is the "live ebuild" for Gentoo, which directly checks out and builds from git master (that's a Gentoo specialty). They also check out submodules in that case. 

> I'd love to see man pages in Charliecloud.

I'd also love it. 
Which approach would you take: 
1. Automatic generation with `help2man` ( https://www.gnu.org/software/help2man/ ), which essentially works by calling the binaries with `--help` and `--version` adding some nice formatting,
2. hand-written full-fledged manpages (which may as well be based on `help2man `output to have a starting point, but need more love, e.g. example calls, maybe some FAQ etc., and manual adaptation for any newly added option)?

reidpr:
Why don't we put them in a top-level `packaging` for now.

As for the man pages, let's continue discussion in issue #82.

Issue #71, opened 2017-11-05
ch-run (--version|--help) cannot be called as privileged user


olifre:
This seems to be a design feature, but a bit non-intuitive in the case charliecloud is installed on the host and a privileged user wants to check e.g. the installed version. 

In real life, I hit this when trying to implement charliecloud support in HTCondor. 
HTCondor's daemons usually run privileged, and it needs some way to check whether charliecloud is set up correctly, and advertise the version (so job submitters can e.g. request to only run on nodes with version 0.xy). 
This fails, since calling `ch-run --version` fails when run as root. 

So in short: 
It would be best if arguments which will not lead to actual container execution (`--help`, `--version`) would also work for privileged users.

reidpr:
This seems like a valid usability issue.

While non-setuid mode lets UID 0 run the program, in practice it may still error out because GID 0 is not permitted.

I see there's a PR now, so I'll make technical comments there.

PR #73, opened 2017-11-06
Perform argument evaluation as calling user unless SETUID.

olifre:
This is a suggestion for #71 . I am not sure this is the best approach, or which approach would be better... 

If we are SETUID, check invoking privs and drop privs,
if not, perform argument parsing and handling as the
calling user.
Check invoking user's privileges only after that,
before actually taking action.

### DIFF
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -377,8 +377,9 @@ void privs_verify_invoking()
    TRY (getresuid(&ruid, &euid, &suid));
    TRY (getresgid(&rgid, &egid, &sgid));
 
-   // GIDs should be unprivileged and non-setgid regardless of mode.
-   TRY (egid == 0);
+   // GIDs should be unprivileged unless called by root,
+   // and non-setgid regardless of mode.
+   TRY (!(ruid == 0 && rgid == 0) && egid == 0);
    TRY (egid != rgid || egid != sgid);
 
 #ifdef SETUID
--- a/test/run.bats
+++ b/test/run.bats
@@ -89,6 +89,17 @@ load common
     sudo rm $CH_RUN_TMP
 }
 
+@test 'ch-run (--help|--version) works as privileged user unless setuid mode' {
+    [[ -z $CH_RUN_SETUID ]] || skip 'compiled in setuid mode'
+    [[ -n $CHTEST_HAVE_SUDO ]] || skip 'sudo not available'
+    run sudo $CH_RUN_FILE --version
+    echo "$output"
+    [[ $status -eq 0 ]]
+    run sudo $CH_RUN_FILE --help
+    echo "$output"
+    [[ $status -eq 0 ]]
+}
+
 @test 'ch-run -u and -g refused in setuid mode' {
     [[ -n $CH_RUN_SETUID ]] || skip 'not compiled for setuid'
     run ch-run -u 65534


reidpr:
Thanks Oliver!

The reason the `privs_verify_invoking()` and `privs_drop_temporarily()` are at the very beginning is to protect against bugs in `argp_parse()` and the other library calls (as well as their invocation) in a setuid binary. That is, if this code is run with elevated privileges and there's a bug allowing the control flow to be hijacked, now you have a root shell.

If the user is really root there's no harm in allowing those switches because there's nothing to escalate, but if it's a setuid binary (regardless of compile mode) I believe we don't want to be doing argument parsing while privileged.

So I wonder if the right approach distinguishes between the invoking user really being root vs. it's a setuid binary.

There's probably test changes associated with this too: we currently verify that the binary refuses to run in certain conditions related to this, and we might want to check that `--help` and `--version` are accepted for privileged callers when we expect them to be. Have you discovered `make test-quick` yet, to get test results w/o waiting for Travis?

olifre:
> The reason the `privs_verify_invoking()` and `privs_drop_temporarily()` are at the very beginning is to protect against bugs in `argp_parse()` and the other library calls (as well as their invocation) in a setuid binary. That is, if this code is run with elevated privileges and there's a bug allowing the control flow to be hijacked, now you have a root shell.

Exactly, that's what I thought it was for ;-). 
So this PR leaves them at the very beginning, if the binary is compiled setuid, to ensure privileges are dropped before the argument parsing starts. 

But you are right, the binary might still be setuid root for some other reason, even if it was not compiled with `SETUID=yes`. This case is not catched anymore by my PR, and needs to be fixed, otherwise argument parsing might be done in privileged mode even though the binary is invoked by an unprivileged user. 

And you are also right, I now see the failing tests. I did know test-quick before, but have not yet used it. For it to be full-featured, I first have to install docker (which in my case of using a minimalistic kernel means rebuilding a kernel with the needed features). I'll do that, but I'm not yet sure whether I manage that today (it's on my list since last week ;-) ). 

I'll then come back to this PR with a more elaborate fix, which takes the mentioned points into account. 
Many thanks! :-)

reidpr:
> But you are right, the binary might still be setuid root for some other reason, even if it was not compiled with SETUID=yes. This case is not catched anymore by my PR, and needs to be fixed, otherwise argument parsing might be done in privileged mode even though the binary is invoked by an unprivileged user.

I wonder if the new behavior could be as simple as, leave the `privs_` calls where they are but change them to let root through while still blocking setuid as now?

> And you are also right, I now see the failing tests. I did know test-quick before, but have not yet used it. For it to be full-featured, I first have to install docker (which in my case of using a minimalistic kernel means rebuilding a kernel with the needed features).

There are a lot of skips without Docker but I believe the key ones are still done, since they use the `chtest` image that doesn't require Docker.

olifre:
> I wonder if the new behavior could be as simple as, leave the privs_ calls where they are but change them to let root through while still blocking setuid as now?

That's a good idea! I adapted the code in this PR, but did not have a look at the tests yet.

olifre:
Now also a test is added that in non-setuid mode, `--version` and `--help` work when called as root.

reidpr:
So I tinkered a little bit, trying to simplify the validation logic and tests. Would you look at 144b26e and let me know if this is consistent with your goals in this PR or if I've messed your stuff up somehow?

I also could not get an image to run under `sudo`, and I don't really understand why, especially since I'm getting different errors in the test suite and running what's apparently the same thing manually. Once this PR is merged, if that remains unresolved let's make a separate issue, since I believe that's not a blocker for the present use case.

olifre:
> Would you look at 144b26e and let me know if this is consistent with your goals in this PR or if I've messed your stuff up somehow?

Nice, that's looking good - and I see you also extended the tests a bit to check that being root with a strange gid does not work. 
That's certainly fine with me, and I can confirm the usecase of checking `--version` and `--help` certainly still works. 

I never tested (up to now) actually running an image as root, so I didn't notice up to now - but yes, this looks well kept in a new issue. 

Relating to your comment in `common.bash` ("`sudo` resets `PATH`), I also had a funny experience when not using an absolute path when adding the test yesterday. In my case, it sadly did not fail, since I had `ch-run` installed to `/usr/bin` via the Gentoo package - so it actually called the wrong version, which confused me a bit.

reidpr:
OK, thanks. You're good with me merging once I convince the tests to pass in Travis?

> I had ch-run installed to /usr/bin via the Gentoo package - so it actually called the wrong version, which confused me a bit.

Exactly... I was getting errors pointing to a line in `ch-run.c` that did not exist.

olifre:
Yes, I'm certainly fine with merging.

PR #74, opened 2017-11-06
Bump apk-tools-static to 2.7.4.

olifre:
Version 2.7.3 is not available for download anymore.


reidpr:
Ugh, thank you.

FWIW this is the 2nd or 3rd time this has happened. I thought that apk-tools would retain a consistent version number for the life of Alpine 3.6, but that is clearly false. If you or anyone has a better approach than hard-coding the version number, I'm all ears.

PR #78, opened 2017-11-09
Add -c|--cd to specify container's initial working directory.

### LINKED ISSUE #72: Allow to specify initial working directory in container
Especially when `--no-home` is not passed, the user might expect to end up in his / her home directory. 
In general, it would be a nice shortcut to be able to specify the IWD in the container.

olifre:
Default remains /.

This fixes #72 .

### DIFF
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -68,6 +68,7 @@ const struct argp_option options[] = {
      "mount SRC at guest DST (default /mnt/0, /mnt/1, etc.)"},
    { "write",       'w', 0,     0, "mount image read-write"},
    { "no-home",      -2, 0,     0, "do not bind-mount your home directory"},
+   { "cd",          'c', "DIR", 0, "initial working directory in container"},
 #ifndef SETUID
    { "gid",         'g', "GID", 0, "run as GID within container" },
 #endif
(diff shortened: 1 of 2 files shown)


bin/ch-run.c
```diff
@@ -266,6 +269,10 @@ void enter_udss(char * newroot, bool writable, struct bind * binds,
 #ifdef SETUID
    privs_drop_temporarily();
 #endif
+
+   if (args.initial_working_dir != NULL) {
+      TRY (chdir(args.initial_working_dir));
+   }
```
reidpr: House style is currently no braces here.

bin/ch-run.c
```diff
@@ -314,6 +321,9 @@ static error_t parse_opt(int key, char * arg, struct argp_state * state)
    case -2:
       as->private_home = true;
    break;
```
reidpr: While you're in there, would you indent this `break` to match?

reidpr:
Looks nice, thank you.

Would you add a test and then rebase into a single commit?

Off the top of my head I can think of testing that `--cd` missing ends up at `/` and present ends up where requested.

bin/ch-run.c
olifre: Done as part of the stashed commit.

olifre:
All done, let's see whether travis confirms my local testing. 
I added also a third test: Check that we really get an error if trying to `--cd` to a directory which does not exist in the container.

reidpr:
Merged, thank you!

I did add a couple of commits to make the error friendlier (4f959b5) and to make the tests a bit more idiomatic (f430ecb).

PR #79, opened 2017-11-10
Use findmnt to extract details of / rather than grepping /proc/mounts

mcv21:
On some systems (e.g. RHEL 7.4 in OpenStack) there are two lines for /
in /proc/mounts:

```
[cloud-user@mv3-rh74-charliecloud ~]$ fgrep ' / ' /proc/mounts
rootfs / rootfs rw 0 0
/dev/vda1 / xfs rw,seclabel,relatime,attr2,inode64,noquota 0 0
```

This means that the mount invocation at `run_uidgid.bats` line 95 fails
as there are newlines embedded in the command arguments.

`findmnt`(8) is a command designed for the task of finding out about
mounted filesystems and has been in util-linux for a fair while, so
use that instead.


reidpr:
Thanks!

PR #81, opened 2017-11-11
Do not fail if PATH is not set.

### LINKED ISSUE #80: ch-run fails if PATH is not set
Again something I found when using HTCondor, which by default does not set any basic environment for a job (of course users can override that). 
This means also `PATH` will not be set.

For this reason, the check at:
https://github.com/hpc/charliecloud/blob/a2de30665e1b4f7826075e7cf1c69d5192ed800b/bin/ch-run.c#L478
will fail, since `getenv("PATH")` will indeed return `NULL`. 

It would probably be best if Charliecloud would not fail in that case, but just perform:
```
setenv("PATH", "/bin", 1)
```
The users may still want to perform `source /etc/profile` within the container (depending on their needs), so a missing `PATH` should not prevent users from entering the container.

olifre:
Previously, we failed if PATH was not set,
since a check was performed before applying
the /bin workaround.
Now, just set PATH to /bin is it was unset before.

Fixes #80 and also includes the appropriate test.

### DIFF
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -475,8 +475,11 @@ void run_user_command(int argc, char * argv[], int user_cmd_start)
    argv[argc - user_cmd_start] = NULL;
 
    // Append /bin to $PATH if not already present. See FAQ.
-   TRY (NULL == (old_path = getenv("PATH")));
-   if (strstr(old_path, "/bin") != old_path && !strstr(old_path, ":/bin")) {
+   old_path = getenv("PATH");
+   if (old_path == NULL) {
+      if (args.verbose)
+         fprintf(stderr, "$PATH is not set, leaving it unset\n");
+   } else if (strstr(old_path, "/bin") != old_path && !strstr(old_path, ":/bin")) {
       TRY (0 > asprintf(&new_path, "%s:/bin", old_path));
       TRY (setenv("PATH", new_path, 1));
       if (args.verbose)
--- a/test/run.bats
+++ b/test/run.bats
@@ -187,6 +187,16 @@ load common
     [[ $output = $PATH2:/bin ]]
 }
 
+@test 'ch-run works if $PATH not set, leaves it unset' {
+    BACKUP_PATH=$PATH
+    unset PATH
+    run $CH_RUN_FILE $CHTEST_IMG -- /usr/bin/python3 -c 'import os; print(os.getenv("PATH") is None)'
+    echo "$output"
+    [[ $status -eq 0 ]]
+    [[ $output = "True" ]]
+    PATH=$BACKUP_PATH
+}
+
 @test 'mountns id differs' {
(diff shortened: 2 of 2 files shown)


reidpr:
What do you think about leaving `$PATH` unset in this case?

The use case for adding `/bin` is to avoid people who had a working `$PATH` on the host being surprised by a broken `$PATH` in the container. But if you already have no `$PATH`, I wonder if the principle of least surprise is suggests leaving it unset rather than providing a minimal one.

olifre:
> What do you think about leaving $PATH unset in this case?

Good point. Indeed that would be less surprising, and people with completely unset `PATH` should in any case know what they are up to (and, for example, may plan source `/etc/profile` within the container). 

I'll adapt the PR accordingly.

trandles-lanl:
Environment variable handling is a treacherous area and should be approached with extreme caution.  I haven't thought through all of the various problems one could encounter, but I do want to draw attention to it in general.  Be wary whenever the topic comes up.

It's worth reviewing some of the various ways environment handling has caused problems for sudo throughout the years.  Sudo is a more extreme case than Charliecloud since it involves privilege level changes.  Still, I wouldn't consider Charliecloud to be immune in all cases.  See https://www.exploit-db.com/exploits/27057/ for one example.  The "Related Exploits" section is useful as well.

olifre:
@trandles-lanl  Many thanks, that looks like a good read!

Now I have another issue. As soon as I remove `setenv("PATH",...` from `ch-run` and leave `PATH` fully unset, invoking `sh` within the container seems to set a sanitized value for `PATH`:
```
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
Apart from the fact that this is somewhat unexpected for me (if `PATH` is set to "something", this does not happen), I'm unsure how to formulate the test. Checking for `PATH` to still be unset does not work, since `sh` sets it. 
So I could only test that `sh` produces the expected setting of `PATH`, or just check that `ch-run` does not fail - or are there any better ideas?

reidpr:
I suspect that it's BusyBox supplying that default path, when invoked as `sh`, though I don't know where it's coming from. I get a different value:

```
$ ch-run /img/chtest -- /bin/sh
> (unset PATH && echo $PATH)

> (unset PATH && /bin/sh -c 'echo $PATH')
/sbin:/usr/sbin:/bin:/usr/bin
```

You could try a different interpreter, e.g.:

```
$ (unset PATH && /usr/bin/python3 -c 'import os; print(os.getenv("PATH") is None)')
True
```

olifre:
> I get a different value: [...] 

Indeed, my bad. I used an Ubuntu 16.04 container to play and test manually, and they link `/bin/sh` to `dash`, which again has a different default. 

Good idea using `python3`! Yes, this works - I have adapted the PR accordingly. Thanks!

reidpr:
Thanks!

PR #84, opened 2017-11-14
clean up build process

reidpr:
Addresses issue #33.

I kept `make export` because it turns out we need that for installing with an embedded Bats.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "test/bats"]
-	path = test/bats.src
+	path = test/bats
 	url = https://github.com/sstephenson/bats.git
 	ignore = untracked
--- a/Makefile
+++ b/Makefile
@@ -15,32 +15,43 @@ clean:
 	cd test && $(MAKE) clean
 	cd examples/syscalls && $(MAKE) clean
 
-# If we're in a Git checkout, rebuild VERSION.full every time (since it's hard
-# to tell if it needs to be rebuilt). If not, it had better be there since we
-# can't build it.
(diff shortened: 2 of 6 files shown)


reidpr:
This is ready for review if anyone is available (e.g. @jlowellwofford, @olifre, etc.)

olifre:
Good point - yes, for this `make export` is still needed. 
I'm booked today and might also be busy tomorrow, but will have a look at it latest the day after, if nobody picks it up in the meantime. 
I would then also check that the adapted build procedure still matches with my current packaging code (I'll likely have to add `bats` as dependency for build, but that's perfectly fine and distros will be pleased). :-)

reidpr:
Whoops, I merged this by mistake. I would still like a review, and I'll just put changes in a new commit.

olifre:
No problem! I just went through it. 
I see you also extended the `travis` tests to cover the new `archive`s. 

Testing with my existing packaging and using a tarball made from Github's master now works fine with all RedHat, Debian and Gentoo. I also made some fixes to the packaging code (before, it copied `VERSION` to `VERSION.full` as a dirty hack, that's not needed anymore). 

I didn't find any issue in the newly introduced code. The only thing which I'm not sure about is whether the case should be handled if a user somehow got hold of a Charliecloud checkout including `.git`, but does not actually have `git` installed on the machine used to compile. Then, not-so-helpful error messages will be generated.
So maybe a check in the `VERSION.full` target and in the `export` target checking whether `git` is in path (before calling it) and producing a nicer error message would be helpful.

PR #85, opened 2017-11-15
prevent building and testing as root

reidpr:
This PR address issues #11 and #12.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,14 @@ SHELL=/bin/bash
 # Add some good stuff to CFLAGS.
 export CFLAGS += -std=c11 -Wall
 
+# Only target "install" is allowed as root.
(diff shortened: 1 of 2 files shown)


reidpr:
I wouldn't mind comments on whether this is a good idea: @trandles-lanl, @mej, @olifre, etc.?

olifre:
I'm a bit unsure whether it's a good idea. 
At least, it breaks Debian packaging. Debian does wrap most steps inside `fakeroot`, even the `build` and `clean` phases (RedHat builds as unprivileged user, same for Gentoo). 
And inside `fakeroot`, the UID is 0 even though no additional privileges are granted.

reidpr:
OK, given that, let's abandon this for the time being.

PR #92, opened 2017-11-19
Add Debian and RedHat packaging + tests, add Gentoo ebuild.

olifre:
The testing requires some hacks, since we are running
in a 14.04 container on travis.
Testing the Gentoo ebuild on Ubuntu is not feasible.


olifre:
When adding this, I found the rule in gitignore:
```
charliecloud-*
```
collides with the (necessary) naming scheme for the Gentoo ebuild, so I had to force `git` to add it. 
Is the rule meant to ignore tarballs / zips? If so, it might be best to adapt it to be more explicit.

reidpr:
> rule in gitignore

It's there so that `make export` tarballs (`charliecloud-$GITHASH.tar.gz`) and directories (`charliecloud-$GITHASH`) can hang around the working directory without trouble. I'm open to changing it to be more precise. Any suggestions? If not, I'll think about it.

olifre:
That's what I assumed ;-). 
One idea would be to have:
- `charliecloud-*.tar*` to match the tarballs. 
- `charliecloud-*/` to match the folders

That's probably flexible enough, but maybe you come up with a better idea.

packaging/gentoo/charliecloud-9999.ebuild
```diff
@@ -0,0 +1,75 @@
...
+
+if [[ ${PV} == "9999" ]] ; then
+	inherit git-r3
+	EGIT_REPO_URI="https://github.com/hpc/${PN}.git"
+	S="${WORKDIR}/${P}"
+else
+	SRC_URI="https://github.com/hpc/${PN}/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="amd64 x86 ~amd64-linux ~x86-linux"
+	S="${WORKDIR}/${P}"
+fi
```
nicolasbock: That kind of conditional might break some package tooling in Gentoo. Typically the preferred solution here is to split the ebuild into two and thus avoiding the conditional blocks.

olifre: That's extremely common practice, as I have learnt from Gentoo devs when proxy-maintaining packages. 
Also, I checked just now, and it's done for 5777 / 38657 of total ebuilds in Gentoo portage, and 1264 / 1954 of the live ebuilds. I would deduce from that that common practice is to not have separate ebuilds. 
So whatever tooling breaks, it must break for >65% of the official in-tree packages - could you tell me which tooling is broken by that? Up to now, I have not encountered any such issue, but was even adviced by Gentoo devs when pushing my proxy-maintained package in to have a common ebuild instead of two slightly different ones for live and release.

olifre:
The package has now been accepted into Gentoo proper! 
https://packages.gentoo.org/packages/sys-cluster/charliecloud

There's now the live ebuild and a prerelease version cut from the Charliecloud tree in Gentoo. I'll update that to releases once we have something >= 0.23. 

I think it's only useful to keep the live ebuild (live = build directly from git master) in the charliecloud repo, i.e.e the `9999`-version. 

I have also done a few fixes to the Gentoo ebuild (as can be seen in the commit added to this PR).

packaging/gentoo/charliecloud-9999.ebuild
nicolasbock: Sure, if @monsieurp approved https://github.com/gentoo/gentoo/pull/6349, I am good :smile:

monsieurp: Hang on a sec... I vouched for the opposite, namely: one live ebuild and one "normal" one. See https://github.com/gentoo/gentoo/pull/6349#issuecomment-348210928. Eventually, that's what @olifre did and I merged his contribution. re your first comment @nicolasbock: never heard of breakage stories due to an if conditional right in the middle of an ebuild to test the package version. It is indeed very common practice. Do you have a link to back up your claim? I'm curious now. :)

monsieurp:
@olifre I think you can update this PR and do away with the Gentoo ebuilds. If people want to emerge charliecloud, they'll most likely go through emerge.

monsieurp:
@olifre and whilst at it, you should update the README file of this project to explain how to emerge it on Gentoo.

olifre:
> @olifre I think you can update this PR and do away with the Gentoo ebuilds. If people want to emerge charliecloud, they'll most likely go through emerge.

Agreed. I originall left it in due to the discussion result achieved here: 
https://github.com/hpc/charliecloud/pull/59 
But the points raised there are mostly valid for Debian / RedHat (there may be different packagers for different (sub-)distros), and also only Debian / RedHat can be (more or less) easily tested via Travis. 
So indeed, for Gentoo the points to keep an ebuild in the upstream tree are naught (since all other ebuild-using distros derive from the main portage repo, I believe). As such, an entry to the `README` would be more useful. 
I'll do this, add this as an additional commit on top, and then wait for @reidpr so he can decide whether this is what he'd like to see in the upstream repo here.

packaging/gentoo/charliecloud-9999.ebuild
nicolasbock: My criticism was only related to the conditional block in the beginning of the ebuild and it goes back to a discussion on `gentoo-devel` where I was told that having such conditional "[...] is going to confuse the hell out of ekeyword" (https://archives.gentoo.org/gentoo-dev/message/015bb27eccec1da192bf550f18923693). And indeed, when I run `ekeyword` on this ebuild it changes both `KEYWORDS` assignments (in the `9999` branch _and_ in the `else` branch) which is maybe what @mgorny meant.

olifre: @nicolasbock Thanks for pointing this out!
This issue with `ekeyword` is tracked here:
https://bugs.gentoo.org/614746
The workaround documented there already works for the updated Charliecloud-package in-tree (since I only have one `KEYWORDS` line in there). However, ideally, either `ekeyword` would see a fix or a check to `repoman` would be added and all ebuilds currently in-tree violating that would be updated...

nicolasbock: Nice workaround :smile:. Thanks for pointing out that bug. Hopefully this issue will get fixed soon in portage and the workaround will become obsolete.

reidpr:
> @reidpr so he can decide whether this is what he'd like to see in the upstream repo here.

I haven't had a chance to look at the changes in this PR yet, but here are the principles I propose:

1. Documentation (`.rst` web docs, READMEs, whatever) should be sufficient for people to understand:
  a. which packages are available
  b. how to build them, given prior experience building for that distro or maybe some links (i.e. teaching someone how to build packages for a given distro is out of scope)
  c. who's responsible for a given distro's packages
  d. where to go with questions about a given packages
2. Actual packaging code should follow distro best practices.

Re. the Gentoo files living in another repo, I'm fine with that, given that it sounds like a best practice for Gentoo.

olifre:
Let's start from the back ;-). 


> Re. the Gentoo files living in another repo, I'm fine with that, given that it sounds like a best practice for Gentoo.

I also think so - since that repo they live in for Gentoo is the main package repository (it's "package build recipe" and "package" in one), so it's the place any Gentoo user will look at first. Building a package from that requires no additional infrastructure (if Gentoo is used). 

For other distros, this is indeed different, since there are several flavours (e.g. debian, ubuntu, mint,...) and several versions, and it may be hard for somebody wanting to build a package to find out whether there is already something existing in any of these. Also, building a package requires a somewhat prepared environment, and special utilities and commands. 



> Actual packaging code should follow distro best practices.

I tried that - apparently I succeeded for Gentoo (the package is in!) and for Debian, we can add the final result of #95 (which contains some modernizations and improvements of my packaging code) on top of this PR. 
For RedHat, the package is as good as I can do, but for sure a RedHat dev with more experience can improve that based on my work. 



> Documentation (.rst web docs, READMEs, whatever) should be sufficient for people to understand:
> a. which packages are available
Adressed in the PR. 

> b. how to build them, given prior experience building for that distro or maybe some links (i.e. teaching someone how to build packages for a given distro is out of scope)
Basic instructions are in the PR, I sadly did not find ideal links to include, but I can search again ;-). 

> c. who's responsible for a given distro's packages
On this point, I am unsure - the package formats are self-documenting, i.e. it's also included in the packaging code. Should we add the names to `install.rst` in addition? 

> d. where to go with questions about a given packages
We could link the distro's bug trackers once the packages are officially included. 
For Gentoo, if somebody reports a bug on https://bugs.gentoo.org concerning the charliecloud-package, it will automatically be assigned to me. 
I think the same is true for other distributions as soon as there is a maintainer officially registered with them (which would then likely be @wiene and @lnussbaum for Debian).

reidpr:
OK, thanks. I'll review this PR as soon as I can.

reidpr:
@olifre, I tinkered with this a bit. I changed around the documentation, and changed how the `packaging` directory is organized. It's in my branch `packaging-deb-rpm-ebuild`.

Would you take a look? If you like these changes, you could cherry-pick 18c2dda into your branch and continue any further work you feel is needed.

olifre:
@reidpr This looks very nice on a first quick look, especially the documentation is structured much cleaner! 
I'll have a more detailed look later this week, there's currently too many different tasks piling up at work, so my open source activity is drowned a bit.

reidpr:
Thanks @olifre. Turns out my branch broke Travis, so I added another commit 7208af0.

olifre:
@reidpr I read through everything, nice work - I especially like the nicely restructured documentation. 
Sadly, I don't have a better idea on how to solve the debian package version issue, at least I have not yet seen any clean solution to this so far. 
Maybe @wiene has an idea on this. I'll cherry-pick your two commits on top now, and In case there are further packaging changes by @wiene / @lnussbaum , I could also cherry-pick these later.

reidpr:
OK, thank you!

I am inclined to merge this and leave further packaging changes for a separate PR. What do people think of that?

olifre:
Fine with me - so I am redirecting the question to @wiene and @lnussbaum on what they prefer (it may also depend a bit on the evolution of #95 ).

wiene:
I am fine with @reidpr's proposal.

reidpr:
Merged, thanks all for your hard work!

PR #99, opened 2018-01-24
clarify where to ssh into the VM. and a typo in tutorial

pagrubel:
I added a small comment when using a VM. I don't know if anyone else will try the ssh command in the console like I did, but I think this clarifies it a bit more. Also there was a typo in the tutorial.


reidpr:
Merged, thank you!

PR #101, opened 2018-01-28
Update apk-tools-static version to avoid failure of unit tests

wiene:
See http://dl-cdn.alpinelinux.org/alpine/v3.6/main/x86_64


reidpr:
Thanks @wiene! Merged.

All: I realize that hard-coding this version number (we've updated it ~3 times in the past few months) is sub-optimal, so I am open to suggestions.

PR #102, opened 2018-01-29
Completely switch to python3

wiene:
Remove remaining python2 dependence.

### DIFF
--- a/test/make-auto
+++ b/test/make-auto
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # If environment variable CH_TEST_OMIT is set, then avoid Dockerfiles that
 # seem lengthy. This supports job time limits in Travis CI. See source code
--- a/test/make-perms-test
+++ b/test/make-perms-test
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # This script sets up a test directory for testing filesystem permissions
 # enforcement in UDSS such as virtual machines and containers. It must be run
@@ -155,7 +155,7 @@ class Test_Directory(Test):
       file_ = self.name + '/file'
       with open(file_, 'w') as fp:
          fp.write(FILE_CONTENT)
-      set_perms(file_, my_uid, my_uid, 0660)
+      set_perms(file_, my_uid, my_uid, 0o660)
 
 
 class Test_File(Test):
@@ -179,7 +179,7 @@ try:
 except OSError as x:
    print("can't mkdir %s: %s" % (testdir, str(x)))
    sys.exit(1)
-set_perms(testdir, my_uid, my_gid, 0770)
+set_perms(testdir, my_uid, my_gid, 0o770)
 os.chdir(testdir)
 
 Test_Directory(my_uid, my_gid, 7, 7, 0, 'nopass').write()


reidpr:
Quick speculation based on perhaps faulty memory: The dependency on Python 2.6+ is to support systems such as RHEL6, which have Python 2.6 and no Python 3.

Is it possible to alter this script so it works if `/usr/bin/env python` is either Python 2.6, Python 2.7, or Python 3? This would assume that Debian installs a `python` alias for Python 3 if 2 isn't installed.

wiene:
That is actually a good point which I did not have on my radar screen. My motivation for this patch was twofold:

1. I noticed that there are already various pure python3 scripts included in charliecloud (e. g. test/chtest/bind_priv.py, test/chtest/dev_proc_sys.py, test/chtest/fs_perms.py, test/chtest/fs_perms.py, test/unused/su_wrap.py). I thought it would make the python dependence of charliecloud more consistent to fully switch to python3.

2. The second motivation was [this discussion](https://lists.debian.org/debian-python/2017/12/msg00017.html) on the Debian python mailing list about the python2 EOL.

I am not sure what is the best way to proceed here. One option RHEL/CentOS/SL 6 users would have if python2 support was dropped, is to make use of the [Redhat Software Collections](https://access.redhat.com/documentation/en-us/red_hat_software_collections) or the [CentOS Software Collections](https://www.softwarecollections.org/en/scls/rhscl/rh-python35).

nicolasbock:
@wiene Maybe this guide [[1]](https://docs.python.org/3/howto/pyporting.html) can help here.

reidpr:
Regarding background, I'm fully on board the Python 3 train and would love to drop Python 2 support as soon as practical.

> I noticed that there are already various pure python3 scripts included in charliecloud (e. g. test/chtest/bind_priv.py, test/chtest/dev_proc_sys.py, test/chtest/fs_perms.py, test/chtest/fs_perms.py, test/unused/su_wrap.py)

One source of confusion for these is that they're run inside the `chtest` container, so we have full control of the environment, and thus it's Python 3 only per the principle above (see `test/chtest/Build` line 76).

> The second motivation was this discussion on the Debian python mailing list about the python2 EOL.

I agree.

FWIW, RHEL 6 is going to be supported for an absurdly long time; [this Red Hat page](https://access.redhat.com/support/policy/updates/errata) states 2024 or maybe longer. 

That said, I'm not interested in holding up an important transition so one can run the test suite using the system Python.

> I am not sure what is the best way to proceed here.

Practically speaking, what this script is currently tested on regularly is Python 2.7, because that's what our `python` binary points to and what is in Travis (I think). So I have no idea if it really works on Python 2.6 or 3.4+.

Options include:

1. Do nothing (i.e., reject this PR)
    * Debian package depends on Python 2.7
    * RHEL/CentOS 6 users can continue using system Python (we think)
    * Python 2.6 support is claimed but not tested
    * Python 2.7 is tested
    * Python 3.4+ support is claimed but not tested (and according to [this](http://python-future.org/compatible_idioms.html) should not work due to the octal syntax)
1. Update the octal literal syntax but keep `/usr/bin/env python` (i.e., modify the PR then accept it)
    * Debian package depends on Python 3, and either 2 or 3 is used depending on what `python` binary points to.
    * RHEL/CentOS 6 users can continue using system Python (we think)
    * Python 2.6 support is claimed but not tested
    * Python 2.7 is tested
    * Python 3.4+ support is claimed but not tested
1. Update the octal literal syntax, change to `/usr/bin/env python3` (i.e., accept PR as-is)
    * Docs/comments and code would be inconsistent.
1. Update the octal literal syntax, change to `/usr/bin/env python3`, update stated compatibility in docs and comments (i.e., modify the PR then accept it)
    * Debian package depends on Python 3, and Python 3 is always used
    * RHEL/CentOS 6 users must install Python 3
    * Python 2.6 and 2.7 are not supported (but probably would work if someone changed it back)
    * Python 3.x+ support is claimed and tested (for some value of *x*)

Given that it's pretty easy to install Python 3 on RHEL/CentOS 6, I'm inclined to endorse Option 4, but I wonder if it's worth gathering opinions on the mailing list first.

wiene:
Asking on the mailing list might help to get a feeling how important people still consider python2 support.

nicolasbock:
In my experience it's far from gone (not that it shouldn't die sooner rather than later), and I would advocate to keep supporting it. It's far easier to just install charliecloud than to also having to install Python3.

wiene:
Due to the lack of user namespaces on RHEL/CentOS 6 there is no simple download, compile and have fun with charliecloud for unprivileged users on this platform. It requires admin intervention in any case and as superuser it should be as easy to install python3 from the software collections repo as it is to install charliecloud. A different question is of course whether admins are willing to add another repo. But for restrictive admins it might also be questionable whether they are willing to install charliecloud with setuid.

nicolasbock:
@wiene, this is really not a question of whether it's possible for an admin to install yet another dependency. Of course it's technically possible in most cases. However in order to ease adoption, installation of charliecloud should be as simple as possible and installing dependencies adds complexity.

reidpr:
I solicited opinions on the mailing list: https://groups.io/g/charliecloud/topic/rfc_should_we_continue_to/9588496

Personally I'm still leaning towards option 4, because as @wiene points out, RHEL/CentOS 6 right now only sort of supports Charliecloud anyway, and the other things that one must do (new kernel or accept setuid) seem a heavier lift to me than installing Python 3. But I'm still open to argument.

mej:
For the record, this doesn't just break RHEL/CentOS/SL 6, but also 7.  RHEL7 and rebuilds ship with Python 2.7, and Python 3.4 is only available via EPEL or other 3rd-party package.

reidpr:
I am reverting this merge. But, I think I can make Travis test both Python 2.7 and 3.4, i.e., basically option 2 above. This will make the scripts work whether `python` is 2.7 or 3.

I hope this will let downstream packagers depend on Python 3.

reidpr:
So it's not trivial to make Travis use Python 3 for testing. I can `pyenv global 3.6`, which causes `python` to be version 3.6.3, but it doesn't propagate through `sudo`, which we run some of the scripts under. I believe additional hacking could fix this, but I don't want to pursue further unless it's helpful for people.

@wiene (or anyone): Would it meet your goal of depending on Python 3 if the two scripts kept unadorned `python` in the shebang line but worked whether that was Python 2.7 or 3?

wiene:
Frankly speaking I do not see a nice solution for this issue if the requirement is that it works on RHEL _without_ the [Software Collections](https://access.redhat.com/documentation/en-us/red_hat_software_collections) repository being enabled. The situation might become even more difficult if you want to add support for RHEL 8 which probably will remove Python 2 completely if I interpret [this Redhat announcement](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/7.5_release_notes/chap-red_hat_enterprise_linux-7.5_release_notes-deprecated_functionality) correctly (of course, for a proper judgement one has to wait and see how Redhat actually implements this change).

mej:
The simplest solution is to turn them into `.in` files and manage the incompatible bits by substitution/transform as part of the build process.  Up to Reid if he wants to do things that way or not.

RH Software Collections are horrific.  Yet another solution to a thoroughly solved problem which is entirely incompatible with all the existing solutions (not to mention woefully more complex to use/manage) is not the right way to solve **ANY** problem.

reidpr:
Note that in the case of these Python scripts, it would be sufficient for any Python 2.7 or 3 executable to be available as `python`. Is the RHEL 8 plan to remove this spelling of the command?

I would prefer not to "build" Python scripts.

wiene:
After thinking a bit more about this issue I would like to make a proposal which resembles Reid's option 2:
Keep the patch but leave the original shebang line (i. e. `#!/usr/bin/env python`) in place. In addition modify the last few lines in the file `test/Makefile` in the following way:
```
PYTHON = $(shell which python3 2> /dev/null || which python)

build_auto.bats: $(IMAGES)
        $(PYTHON) ./make-auto build $^ > $@

run_auto.bats: $(IMAGES)
        $(PYTHON) ./make-auto run $^ > $@
```
This way, preference is given to `python3` for Charliecloud without changing the system default settings (which could have unwanted side-effects for other software). If `python3` is not available, `python` is the fallback choice.

Concerning RHEL 8, I have no idea how exactly they want to implement the announced removal of python2. I just checked the situation in the Fedora 28 Beta Workstation live image. It still includes both python2 (available as `python`) and python3 (available as `python3`). Whether `python` will point to python3 in RHEL 8 and whether there will still be `python3` if python2 is removed, remains to be seen.

wiene:
The Python situation for RHEL 8 is nicely described in [this blog article](https://developers.redhat.com/blog/2019/05/07/what-no-python-in-red-hat-enterprise-linux-8/). Are there any plans how to handle this situation in Charliecloud?

reidpr:
I wonder if our RPM packaging changes in PR #385 will cover this? We are patching at install time to specify explicitly `python3` or `python2` depending on what's available in the distro.

j-ogas:
> I wonder if our RPM packaging changes in PR #385 will cover this? We are patching at install time  to specify explicitly `python3` or `python2` depending on what's available in the distro.

The spec file makes use of a function that changes our python script shebangs from  
`/usr/bin/env python` to `/usr/bin/python3`.  

This function was created to deal with Fedora packaging guidelines and doesn't really address this issue. The spec file lists `/usr/bin/python3` as a test package build dependency and doesn't have conditional macros to allow a builder to specify a different python version, i.e., the spec file would have to be edited.

We did have conditional macros that allowed something like: `rpmbuild -ba --with python2` in a previous iteration. IIRC, we removed it at the request of the Fedora package reviewer. FWIIW, I don't think the `--with` conditional solves this problem either, as you'd have to manually build and install with rpmbuild.

reidpr:
Can we just depend on `python3` for RHEL 8? I don't see any need to give choices.

PR #110, opened 2018-02-09
Enable man page generation and add man pages for the ch-* commands

### LINKED ISSUE #82: add man pages
@olifre in PR #59:

> Which approach would you take:
>
> 1. Automatic generation with help2man ( https://www.gnu.org/software/help2man/ ), which essentially works by calling the binaries with --help and --version adding some nice formatting,
> 2. hand-written full-fledged manpages (which may as well be based on help2manoutput to have a starting point, but need more love, e.g. example calls, maybe some FAQ etc., and manual adaptation for any newly added option)?

I think my preference would be for option 2, since I feel the `--help` output should be pretty concise (less than a screen) and hand-written man pages would let us say a bit more about the various nuances and trade-offs.

It would be nice if there were tests to verify that the same options were described in `--help` and the man page, and also frankly the code itself. But I don't think it's worth holding things up over that.

Do man pages need to be build before installing?

wiene:
This is my proposal to fix #82. Comments are welcome.


reidpr:
Also, could you please add yourself to the contributors list in the readme?

wiene:
All comments should be addressed by the last three commits.

What shall we do with the man pages after they are built? Should we copy them to a dedicated man page directory outside the _build directory (similar to how it is done with the html documentation)?

reidpr:
Thanks so much — this is top-notch work!

I spent some time this afternoon tinkering around the edges; see my branch `manpages`. Would you review these changes? If you're happy with them, I can merge directly or further work you'd like to add on top. If not, cherry-pick the commits you do like, if any, and we can continue iterating.

Question: Would it help you to have the man pages in 0.2.3?

wiene:
It looks very nice. I do not have further comments.

I think that including man pages in 0.2.3 would be nice.

reidpr:
Great! I'll merge this afternoon. Thanks again for your hard work on this.

Issue #65, opened 2017-10-31
test shortening is a kludge


reidpr:
If we run the complete test suite on Travis, we run into the wall-clock limit and it takes a long time to get the result. Thus, the Travis tests are a subset. This is accomplished by setting environment variable `$CH_TEST_OMIT`, which if set excludes any tests that contain it as a substring. We set `CH_TEST_OMIT=mpi` for Travis. This is a total kludge.

Further, there's a `make test-quick` target for a short sanity check.

This issue is to plan and implement tests of varying length in a sane way.

Perhaps the notion of test pre-requisites that the `Build` tests have is a step along the way.

PR #131, opened 2018-03-15
deal with test suite duration in a reasonable way

reidpr:
This addresses issue #65.


reidpr:
Thanks for your help, @ollielo!

> I am not so sure how you divide the test into categories

It's a balance between how long the test takes and how important it is. The basic guidelines are at the end of `doc-src/test.rst`.

> Is it correct that the scope of each test actually encoded as a comment in the Dockerfile?

Yes.

> How does the test inspect it?

The test suite picks it up in `test/make-auto`. Individual tests declare their scope with the `scope` shell function, defined in `test/common.bash`.

ollielo:
@reidpr 

Is there any reason this test system have to be implemented in 3 languages?

BTW, there are conflicts, can you resolve them?

reidpr:
> Is there any reason this test system have to be implemented in 3 languages?

Reasoning:

- Bash is there because the test framework (Bats) uses it.
- Python is fast to code in.
- C lets you be more precise about the system calls that are happening.

> BTW, there are conflicts, can you resolve them?

Yes, those are no big deal. Once the PR is approved, I'll rebase and resolve them before merging.

reidpr:
> Are those test actually running on Travis?

[Yes](https://travis-ci.org/hpc/charliecloud). Does the "all checks have passed" notice show up for you?

Travis runs the "standard" scope because the full scope times out with a 10-minute period having not output.

> Can you also put the link/icon to the travis page as well?

Do you mean the "build passing" image on the Charliecloud GitHub? I have a vague recollection that there was some complication, but I don't recall what it was.

I think it's a good idea but orthogonal to this PR.

ollielo:
O.k.

Do I have the write permission to the repo to merge the PR?

reidpr:
Thanks Ollie. I'm not sure on permissions but I prefer to let the BDFL (me) merge branches.

Issue #31, opened 2017-10-04
support nvidia-docker images


bbarker:
Hi,

First, thanks for the awesome work. This really looks great for HPC. One area of HPC that comes up from time to time is visualization.

I was hoping to use Charlie Cloud to run an image that I already have working with X11 and CUDA in docker, but have as yet been unsuccessful in getting OpenGL working. Isolation isn't a priority, so charliecloud seems like a good option. It seems there may be some work needed in order to get [nvidia-docker](https://github.com/NVIDIA/nvidia-docker) images to be compatible, as I get the following error:

```
$ glxgears                                                                                                                                                                                                                                                                                                                                  
glxgears: error while loading shared libraries: libGL.so.1: cannot open shared object file: No such file or directory 
```
I was building this based on `nvidia/cuda:8.0-cudnn7-runtime-ubuntu16.04`

This wasn't an error when run from Docker (the issues in docker are that 1) acceleration doesn't appear to be working based on the reported framerates of glxgears in the console when the GL window is maximized and 2) the OpenGL window is just black).

bbarker:
Not sure it is of interest as it is rather docker specific, but the `docker run` command I use to get X going is in  https://github.com/CornellCAC/docker-nix-intellij/blob/master/idea.sh

I suspect that with charlie cloud most of this would be unnecessary.

reidpr:
Hello Brandon, thanks for the fine bug report! I agree that something like this would be helpful. GPUs are on our list.

I was unaware of the nvidia-docker stuff. Thanks for the pointer. We'll look into it.

This may be the germ of a test case that lives in `examples/`.

If you find any success before we do, please update this bug report.

bbarker:
I was just able to get it working in docker using `docker run --privileged` (will update my `idea.sh` script shortly with the changes). I know that isn't a charlie-cloud solution, and I think the problems there are likely just related to environment configuration. 

For what it is worth, I didn't build my image using the `ch-build` command, but used an existing docker image I had, if that matters.

Will update here when I come back to this particular issue - looking forward to using charlie cloud on some non-OpenGL jobs in HPC soon.

trandles-lanl:
Hi Brandon,

[...]

PR #147, opened 2018-04-18
ch-fromhost to inject shared libraries and other files into image

reidpr:
This PR addresses the GPU support requested in #31 by implemented a generic facility to inject shared libraries, executables, and other files into an image.

Rationale specifically for the approach to nVidia GPU support is in the man page for `ch-fromhost`, i.e., `ch-fromhost_desc.rst`.

Any early feedback would be extremely helpful. If this approach seems good, I'll tidy it up and add a CUDA test.

### DIFF
--- a/bin/ch-fromhost
+++ b/bin/ch-fromhost
@@ -27,7 +27,7 @@ debug () {
 }
 
 ensure_nonempty () {
-    [[ -z $2 ]] && fatal "$1 must not be empty"
+    [[ $2 ]] || fatal "$1 must not be empty"
 }
 
 fatal () {
@@ -41,12 +41,12 @@ info () {
 
 while [ $# -gt 0 ]; do
     OPT="$1"; shift
+    OUT=
     case $OPT in
         -c|--cmd)
             ensure_nonempty --cmd "$1"
             OUT=$($1)
-            [[ $? -ne 0 ]] && fatal "command failed: $1"
-            FOUND_FILES="${FOUND_FILES:+$FOUND_FILES$NEWLINE}$OUT"
+            [[ $? -eq 0 ]] || fatal "command failed: $1"
             shift
             ;;
         -d|--dest)
@@ -57,8 +57,7 @@ while [ $# -gt 0 ]; do
         -f|--file)
             ensure_nonempty --file "$1"
             OUT=$(cat "$1")
-            [[ $? -ne 0 ]] && fatal "cannot read file: $1"
-            FOUND_FILES="${FOUND_FILES:+$FOUND_FILES$NEWLINE}$OUT"
+            [[ $? -eq 0 ]] || fatal "cannot read file: $1"
             shift
             ;;
         -h|--help)
@@ -68,6 +67,8 @@ while [ $# -gt 0 ]; do
             INFER=
             ;;
         --nvidia)
+            OUT=$(nvidia-container-cli list --binaries --libraries)
+            [[ $? -eq 0 ]] || fatal "nvidia-container-cli failed"
             ;;
         -v|--verbose)
             VERBOSE=yes
@@ -82,11 +83,12 @@ while [ $# -gt 0 ]; do
             ;;
         *)
             ensure_nonempty "image path" "$OPT"
-            [[ $IMAGE ]] && fatal "duplicate image path: $OPT"
+            [[ -z $IMAGE ]] || fatal "duplicate image path: $OPT"
             [[ -d $OPT ]] || fatal "image not a directory: $OPT"
             IMAGE="$OPT"
             ;;
     esac
+    FOUND_FILES="${FOUND_FILES:+$FOUND_FILES$NEWLINE}$OUT"
 done
 
 debug "injecting into image: $IMAGE"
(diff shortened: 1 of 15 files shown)


reidpr:
@ryanolson, I'd love to get your opinion on this approach.

bin/ch-fromhost
```diff
@@ -0,0 +1,134 @@
+#!/bin/bash
+
+# Note: The ${x:+y} construction yields $y if $x is set and not null, or $x
```
mej: This should say `${x:+$y}` instead.  The expansion of `${x:+y}` yields exactly `y`, not the value of `$y`, if `$x` has a non-empty value.  Also, if `$x` is empty or unset, **nothing** is substituted, not `$x`.  This is an important distinction for word expansion because "nothing" does not count as a word.

bin/ch-fromhost
```diff
@@ -0,0 +1,134 @@
...
+    cat 1>&2 <<EOF
+Usage: FIXME
+EOF
+    exit ${1:-1}
+}
+
+DEST_DEFAULT=
+IMAGE=
+INFER=yes
+NEWLINE='
```
mej: The more correct way to say this in `bash` is `$'\n'`.

bin/ch-fromhost
```diff
@@ -0,0 +1,134 @@
...
+        case $FILE in
+            */bin*)
+                TYPE_=bin
+                DEST=/usr/bin
+                ;;
+            */lib*)
+                TYPE_=lib
+                DEST=/usr/lib
+                FOUND_LIBS_P=yes
+                ;;
```
mej: What about `/lib64` and/or `/usr/lib64`?

bin/ch-fromhost
```diff
@@ -0,0 +1,134 @@
...
+        esac
+    fi
+    [[ $VERBOSE ]] && echo "  $TYPE_: $FILE -> $DEST" 1>&2
+    [[ $DEST ]] || fatal "no destination for: $FILE"
+    [[ ${DEST:0:1} = '/' ]] || fatal "not an absolute path: $DEST"
+    [[ -d $IMAGE$DEST ]] || fatal "not a directory: $IMAGE$DEST"
+    cp --dereference --preserve=all $FILE "$IMAGE/$DEST"
```
mej: `$FILE` should be quoted here as well, and in the `case` above.

bin/ch-fromhost
reidpr: The script should be targeting POSIX sh, not Bash. `/bin/bash` and all the bashisms are a mistake. (Currently we claim that using Charliecloud requires only sh, but the test suite needs Bash.)

bin/ch-fromhost
reidpr: I'm assuming all libraries can be dropped into `/usr/lib` regardless of where they were found on the host. Is this an error?

bin/ch-fromhost
reidpr: Thanks for the `$FILE` catch.

Re. `case`, [not needed](http://wiki.bash-hackers.org/syntax/ccmd/case): "The word `<WORD>` is expanded using tilde, parameter and variable expansion; arithmetic, command and process substitution; and quote removal. No word splitting, brace, or pathname expansion is done, which means you can leave expansions unquoted without problems".

bin/ch-fromhost
mej: On multilib systems, I believe placing 64-bit libraries into 32-bit `/usr/lib` could cause problems, if for no other reason than `/usr/lib` isn't searched by default by the x86_64 `ld.so` AIUI.  But I've never tried, so perhaps I'm mistaken there.  But certainly one could not maintain the multilib split correctly once 64-bit libraries are placed in `/usr/lib`.

bin/ch-fromhost
mej: By the same token, quote removal is performed, which means it will work fine to quote it there.  And there is something to be said for consistency.  :-)

But you are correct, from what I can tell reading http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html -- it's not required in that particular instance.

bin/ch-fromhost
mej: POSIX `sh` does not appear to support `$'...'` constructs, so I guess it should stay as-is.  But I don't believe `[[ ... ]]` is POSIX either, so those would need to be changed.  I think `${DEST:0:1}` is also `bash`-specific.

bin/ch-fromhost
```diff
@@ -0,0 +1,134 @@
...
+            */lib*)
+                TYPE_=lib
+                DEST=/usr/lib
+                FOUND_LIBS_P=yes
+                ;;
+        esac
+    fi
+    [[ $VERBOSE ]] && echo "  $TYPE_: $FILE -> $DEST" 1>&2
+    [[ $DEST ]] || fatal "no destination for: $FILE"
+    [[ ${DEST:0:1} = '/' ]] || fatal "not an absolute path: $DEST"
```
mej: This is `bash`-specific, AFAIK.  Use this instead:  `[ -z "${DEST%%/*}" ] || fatal ...`

reidpr:
I plan to tidy this up and merge it tomorrow unless anyone objects.

bin/ch-fromhost
reidpr: Thanks. I'm going to leave it for now, but I recorded the potential problem in issue #153.

reidpr:
Assuming the tests pass, I plan to merge this on Monday. Last chance for any feedback!

PR #149, opened 2018-04-27
ch-docker2tar: add progressbar using pv

### LINKED ISSUE #148: docker2tar: use pv for status reporting
docker2tar can take a while, using `pv` (http://www.ivarch.com/programs/pv.shtml) might help users to stay patient!

junghans:
Fix #148

### DIFF
--- a/bin/base.sh
+++ b/bin/base.sh
@@ -16,3 +16,15 @@ if ( command -v pigz >/dev/null 2>&1 ); then
 else
     export GZIP_CMD=gzip
 fi
+
+# pipe view is only useful when using an interactive terminal
+if ( [ -t 1 ] && command -v pv >/dev/null 2>&1 ); then
+    PV() {
+        pv "$@"
+    }
+else
+   PV() {
+       # ignore all agruments, so that PV() is just a pass-through and does nothing
+       cat
+   }
+fi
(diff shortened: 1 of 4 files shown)


junghans:
We could install `pv` in Travis, too!

bin/base.sh
```diff
@@ -16,3 +16,13 @@ if ( command -v pigz >/dev/null 2>&1 ); then
 else
     export GZIP_CMD=gzip
 fi
+
+if ( command -v pv >/dev/null 2>&1 ); then
```
reidpr: What if we also add to this condition, only do `pv` if we are on a terminal?

bin/base.sh
```diff
@@ -16,3 +16,13 @@ if ( command -v pigz >/dev/null 2>&1 ); then
...
+    PV() {
+        pv "$@"
+    }
+else
+   PV() {
+       cat
```
reidpr: Can you add a comment her saying that the arguments are being ignored and why?

bin/base.sh
junghans: Done

bin/base.sh
junghans: Done

junghans:
> Thanks Christof. This helps!

`s/f/ph/`

> In addition to the in-line comments, do you have time to also:
> 
> Review the other scripts and see if there are more opportunities for this trick (e.g. ch-tar2dir)?

I did `ch-tar2dir` as well.

> Add yourself to the contributors list in the readme?

Done

Issue #118, opened 2018-02-19
debian link can cause conflicts with debian directory used by packaging tools


wiene:
Since commit 18c2dda Charliecloud contains a link `debian` pointing to `packaging/debian`. This link creates a conflict with the `debian` directory of the [official Debian packaging repository for Charliecloud](https://salsa.debian.org/hpc-team/charliecloud). I suppose @olifre's idea to put the packaging files in a separate directory was to avoid conflicts like this. Otherwise he would have faced similar problems with his [openSUSE build service builds](https://build.opensuse.org/project/show/home:ubn:charliecloud), I guess.

Therefore my suggestion would be to remove this link if there is no particular reason to keep it.

olifre:
> I suppose @olifre's idea to put the packaging files in a separate directory was to avoid conflicts like this. Otherwise he would have faced similar problems with his openSUSE build service builds, I guess.

Exactly, this was the initial reason. There appear to be all kinds of different techniques:
- https://github.com/ceph/ceph uses `debian`
- https://github.com/htcondor/htcondor hides it in `build/packaging`

I guess the best would be to go what's easiest for Debian packaging.

reidpr:
I put it there; I'm sorry for the hassle.

The reasoning was to give a one-liner that builds a `.deb` from the upstream source code. For example, see lines 33–39 of `test/travis.sh`, which in turn invokes `packaging/debian/travis.sh`. With the symlink there, you can just `debuild`.

This is only one of the failures to follow best practices in the Debian packaging. I'd love to get it in better shape. Would it work to put the official Debian packaging in `packaging/debian` and leave the link in place?

As for an immediate fix to this bug, I can remove the link from the repository and create/remove it in `packaging/debian/travis.sh`. Is there a better approach we should do now?

wiene:
I found a way to tell `gbp` to ignore the debian link. This fixes the problem for the official Debian packaging repository but I suppose that the current structure still requires using patched tarballs to get successful builds on the Open Build Service (which might be a tool to address #100)

reidpr:
OK.

What's the appropriate thing for me to do now to address this bug?

wiene:
All deb packaging tools which I know (like `git-buildpackage` and the Open Build Service) keep the upstream source and - at some point - add a `debian` directory to the root of the upstream directory structure before starting the actual package build process. If the upstream source already contains a `debian` directory/file/link, you get a conflict during the merge of the upstream source and the `debian` directory.

[...]

PR #159, opened 2018-05-12
Add deb target to Makefile and remove debian link

wiene:
This is an implementation of the proposal made in #118.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,3 +168,9 @@ install: all
 	    ln -sf ../libexec/bats $(TEST)/bats/bin/bats && \
 	    ln -sf bats/bin/bats $(TEST)/bats; \
 	fi
+
+.PHONY: deb
+deb:
+	ln -s packaging/debian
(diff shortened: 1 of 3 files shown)


reidpr:
Thanks @wiene!

Can you comment briefly on putting the target in a top-level Makefile, rather than a (new IIRC) `packaging/Makefile`? I had been thinking the latter would be better, but @mej has also proposed packaging targets in the top-level Makefile as well, so I'd like to understand this reasoning.

wiene:
I think depending on your point of view you might prefer different locations for the deb target.

Since package building is a process involving several subcomponents, one might favour the top-level Makefile as place for the deb target. On the other hand the deb target might be irrelevant for many users. Therefore one could also argue that one should "hide" it in the packaging directory. I do not have a clear preference.

BTW: An updated version of `packaging/debian/README` should be added to this PR.

reidpr:
Merged, thank you!

Issue #127, opened 2018-03-07
add more robust MPI test


reidpr:
The little program in `mpihello` doesn't sufficiently exercise MPI. For example, it works fine in cases where LAMMPS (`examples/mpi/lammps_mpi`) does not.

PR #162, opened 2018-05-15
better MPI tests

reidpr:
This addresses issue #127.

I have not yet tested on any clusters, but I'll do so before merging.

@hjelmn, @hppritcha: Any opinions on this approach?


hppritcha:
:+1: this looks like a good way to add IMB smoke tests.

PR #167, opened 2018-05-25
fix description of ch-run to include mounted home directory

pagrubel:
I wasn't looking to use the --no-home option so didn't see in the description that my home directory is mounted as /home/$USER. I hope this helps the next person this happens to.


reidpr:
I extended this PR to document all the standard bind mounts, not just `$HOME`.

Merged; thank you!

Issue #128, opened 2018-03-08
OpenMPI single-copy via cross-memory attach (CMA) fails with EPERM


reidpr:
OpenMPI single-copy message transfers via CMA do not work because the processes need permission to `ptrace(2)` one another, which they don't have.

Details are in the FAQ.

This issue is to track whether we can do something about this in Charliecloud.

See also issue #126.

PR #174, opened 2018-06-05
implement --join

reidpr:
This PR addresses #128 by adding `--join`, which makes a group of peer `ch-run` invocations started in parallel use the same namespaces, instead of different namespaces for each invocation.

### DIFF
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -65,8 +65,8 @@ struct {
    sem_t * sem;
    char * shm_name;
    struct {
-      pid_t winner_pid;
-      int proc_left_ct;
+      pid_t winner_pid;  // access anytime after initialization (write-once)
+      int proc_left_ct;  // access only while serial
    } * shared;
 } join;
 
@@ -380,9 +380,6 @@ void join_end()
    join.shared->proc_left_ct--;
    INFO("join: %d peers left excluding myself", join.shared->proc_left_ct);
 
-   // Parallelize.
-   Z_ (sem_post(join.sem));
-
    if (join.shared->proc_left_ct <= 0) {
       INFO("join: cleaning up IPC resources");
       Te (join.shared->proc_left_ct == 0, "expected 0 peers left but found %d",
@@ -390,6 +387,9 @@ void join_end()
       Zf (sem_unlink(join.sem_name), "can't unlink sem: %s", join.sem_name);
       Zf (shm_unlink(join.shm_name), "can't unlink shm: %s", join.shm_name);
    }
+
+   // Parallelize.
+   Z_ (sem_post(join.sem));
 
    Z_ (munmap(join.shared, sizeof(*join.shared)));
    Z_ (sem_close(join.sem));
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -311,8 +311,8 @@ void join_begin()
 {
    int fd;
 
-   T_ (1 <= asprintf(&join.sem_name, "/ch-run.s.%s", args.join_tag));
-   T_ (1 <= asprintf(&join.shm_name, "/ch-run.m.%s", args.join_tag));
+   T_ (1 <= asprintf(&join.sem_name, "/ch-run_%s", args.join_tag));
+   T_ (1 <= asprintf(&join.shm_name, "/ch-run_%s", args.join_tag));
 
    // Serialize.
    join.sem = sem_open(join.sem_name, O_CREAT, 0600, 1);
@@ -390,8 +390,8 @@ void join_end()
       INFO("join: cleaning up IPC resources");
       Te (join.shared->proc_left_ct == 0, "expected 0 peers left but found %d",
           join.shared->proc_left_ct);
-      Z_ (sem_unlink(join.sem_name));
-      Z_ (shm_unlink(join.shm_name));
+      Zf (sem_unlink(join.sem_name), "can't unlink sem: %s", join.sem_name);
+      Zf (shm_unlink(join.shm_name), "can't unlink shm: %s", join.shm_name);
    }
 
    Z_ (munmap(join.shared, sizeof(*join.shared)));
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -44,6 +44,14 @@ const char * DEFAULT_BINDS[] = { "/dev",
 /* Log the current UIDs. */
 #define LOG_IDS log_ids(__func__, __LINE__)
 
+/* Environment variables used for --join parameters. */
+char * JOIN_CT_ENV[] =  { "OMPI_COMM_WORLD_LOCAL_SIZE",
+                          "SLURM_STEP_TASKS_PER_NODE",
+                          "SLURM_CPUS_ON_NODE",
+                          NULL };
+char * JOIN_TAG_ENV[] = { "SLURM_STEP_ID",
+                          NULL };
+
 
 /** Command line options **/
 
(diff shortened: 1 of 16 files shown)


reidpr:
@mej, @pagrubel, @mcpherson, I'd love your opinions on this.

mcpherson:
I'm trying to figure out if this is a workaround or a feature (or both). I did look at your new documentation, but it's still a bit unclear. As I understand it, a major motivator for this is that MPI automatically switches to a communication interface (on-node shared memory?) when message sizes get above a certain size. So, this is **required** to work around that issue? Are there parameters to `mpirun` that restrict that automatic switch (turn it off), removing the need for this workaround (but perhaps with performance degradation)? I'm assuming not, or we would have been using them while this PR was in the works. However, if there are such parameters, maybe they should be clearly documented, giving the user a choice of using `join` or not. Also, is there a future path of MPI development that you know will make this workaround unnecessary (or temporary)? If so, you may want to put a note about that in the documentation as well.

As a feature you mentioned attaching a debugger. Are there any other examples you can think of? @trandles-lanl mentioned today that monitoring might be another use for this feature. Perhaps the more close-to-the-metal HPC performance monitors could be enhanced by this feature?

mcpherson:
Working backwards, and looking at the issues, they are answering some of my questions.

pagrubel:
I won't get to this till next week

bin/ch-run.c
```diff
@@ -387,7 +627,23 @@ void run_user_command(int argc, char * argv[], int user_cmd_start)
 
    Zf (prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0), "can't set no_new_privs");
    execvp(argv[0], argv);  // only returns if error
-   Tf (0, "can't execve(2) user command");
+   Tf (0, "can't execve(2): %s", argv[0]);
+}
+
+/* Wait for semaphore sem for up to timeout seconds. If timeout or an error,
+   exit unsuccessfully. */
+void sem_timedwait_relative(sem_t * sem, int timeout)
```
hppritcha: do you want this function to have global visibility?

bin/ch-run.c
```diff
@@ -113,26 +153,40 @@ int main(int argc, char * argv[])
...
-   enter_udss(args.newroot, args.writable, args.binds,
-              args.private_tmp, args.private_home);
+   INFO("verbosity: %d", verbose);
+   INFO("newroot: %s", args.newroot);
+   INFO("container uid: %u", args.container_uid);
+   INFO("container gid: %u", args.container_gid);
+   INFO("join: %d %d %s", args.join, args.join_ct, args.join_tag);
+   INFO("private /tmp: %d", args.private_tmp);
+
+   if (args.join)
```
hppritcha: this series of steps looks correct as is, but are signal handlers in place here protect against signals interfering with the procedure/result in uncleaned up resources?

Also,  are we confident that if things go south and the job is killed, that if charlie cloud is running under SLURM, that these semaphore and shm files will be cleaned up as part of job cleanup?

doc-src/ch-run_desc.rst
```diff
@@ -47,11 +57,76 @@ unpacked image directory located at :code:`NEWROOT`.
...
+* :code:`--join-tag` sets the tag that names the peer group. The default is
+  environment variable :code:`SLURM_STEP_ID`, if defined; otherwise, the PID
+  of :code:`ch-run`'s parent. Tags can be re-used for peer groups that start
+  at different times, i.e., once all peer :code:`ch-run` have replaced
+  themselves with the user command, the tag can be re-used.
+
+Caveats:
+
+* One cannot currently add peers after the fact, for example, if one decides
+  to start a debugger after the fact. (This is only required for code with
```
hppritcha: ha ha

bin/ch-run.c
hppritcha: never mind, I'm thinking libraries but this is an executable

test/travis.yml
```diff
@@ -45,6 +45,13 @@ install:
   - sudo apt-get install pigz
   - if [ -n "$INSTALL_PV" ]; then sudo apt-get install pv; fi
   - sudo pip install sphinx sphinx-rtd-theme
+# Ubuntu Trusty symlinks /dev/shm -> /run/shm; the latter is unavailable after
+# entering the namespaces and we get ENOENT. Work around this bug. See:
+# https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=851427
+# https://wiki.debian.org/ReleaseGoals/RunDirectory
+  - sudo rm /dev/shm
+  - sudo mkdir /dev/shm
+  - sudo mount --bind /run/shm /dev/shm
```
hjelmn: In case they change this behavior you might want to protect it with ```if test -h /dev/shm```.

bin/ch-run.c
```diff
@@ -181,9 +183,8 @@ int main(int argc, char * argv[])
       setup_namespaces(args.container_uid, args.container_gid);
       enter_udss(args.newroot, args.writable, args.binds,
                  args.private_tmp, args.private_home);
-   } else {
-      //join_namespaces();
-   }
+   } else
+      join_namespaces();
    if (args.join)
```
hjelmn: Is a ```}``` missing here? You deleted one and I do not see one added anywhere. The indentation on L188 makes it look like deleting the ```}``` was unintentional.

bin/ch-run.c
reidpr: > this series of steps looks correct as is, but are signal handlers in place here protect against signals interfering with the procedure/result in uncleaned up resources?

No, there aren't any signal handlers. Which signals would pose a risk?

> Also, are we confident that if things go south and the job is killed, that if charlie cloud is running under SLURM, that these semaphore and shm files will be cleaned up as part of job cleanup?

On our systems, yes we are confident.

If they're not cleaned up, though, and there's a collision later: if it's a different user, their job will fail with permission denied (because other users can't access the files), and if it's the same user, something messy will happen and the `--join` coordination will fail somehow.

test/travis.yml
reidpr: That's a good point. In this particular case, we're targeting a specific environment in Travis (Ubuntu 14.04 IIRC). Newer versions of Ubuntu don't have this bug. So the Travis tests will break then and we'll remove the above workaround.

bin/ch-run.c
reidpr: The GitHub diff is hiding the matching also-removed brace (line 184 in the original).

reidpr:
@mcpherson, in case your questions weren't all answered:

> I'm trying to figure out if this is a workaround or a feature (or both).

It's a workaround. The kernel disallows `ptrace(2)`, and hence CMA, between sibling namespaces. I personally think it's unnecessary, and maybe this will change in the future.

> As I understand it, a major motivator for this is that MPI automatically switches to a communication interface (on-node shared memory?) when message sizes get above a certain size. So, this is required to work around that issue?

I'm not sure exactly what the triggers are, but the optimization is to read and write directly from other ranks' memory, using the `process_vm_readv(2)` and `process_vm_writev(2)` system calls. This is a one-copy thing rather than two-copy as in standard shared memory.

> Are there parameters to mpirun that restrict that automatic switch (turn it off), removing the need for this workaround (but perhaps with performance degradation)? [...] However, if there are such parameters, maybe they should be clearly documented, giving the user a choice of using join or not.

I'll improve the documentation.

> Also, is there a future path of MPI development that you know will make this workaround unnecessary (or temporary)?

Because it's a performance thing, I don't think it's going away from an MPI perspective. However maybe the kernel will change one of these days, and then we'll see it in Red Hat a decade later. ;)

> Are there any other examples you can think of? [...] Perhaps the more close-to-the-metal HPC performance monitors could be enhanced by this feature?

Maybe. Anything that uses `ptrace(2)` or things that traverse the same permissions path would be affected.

Let me know what other questions you have.

reidpr:
I found a race condition in `join_end()`. I will try to fix it on Monday.

reidpr:
This should be marked merged, not closed.

Issue #170, opened 2018-06-01
refactor actual container stuff into charliecloud.c


reidpr:
Currently, `ch-run.c` contains both CLI stuff and "real" container code. This should be split, with the container stuff going in `charliecloud.c` and the CLI stuff remaining in `ch-run.c`.

PR #188, opened 2018-06-22
refactor container stuff into charliecloud.c

reidpr:
Addresses #170 and #187.

### DIFF
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -16,8 +16,6 @@
 
 /** Constants and macros **/
 
-/* Maximum number of user binds. */
-
 /* Environment variables used by --join parameters. */
 char * JOIN_CT_ENV[] =  { "OMPI_COMM_WORLD_LOCAL_SIZE",
                           "SLURM_STEP_TASKS_PER_NODE",
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -6,69 +6,25 @@
 #define _GNU_SOURCE
 #include <argp.h>
 #include <assert.h>
-#include <errno.h>
-#include <fcntl.h>
-#include <libgen.h>
-#include <sched.h>
-#include <semaphore.h>
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
-#include <sys/mman.h>
-#include <sys/mount.h>
-#include <sys/prctl.h>
-#include <sys/stat.h>
-#include <sys/syscall.h>
-#include <time.h>
 #include <unistd.h>
 
 #include "charliecloud.h"
 
+
 /** Constants and macros **/
 
-/* Host filesystems to bind. */
-#define USER_BINDS_MAX 10
-const char * DEFAULT_BINDS[] = { "/dev",
-                                 "/etc/passwd",
-                                 "/etc/group",
-                                 "/etc/hosts",
-                                 "/etc/resolv.conf",
-                                 "/proc",
-                                 "/sys",
-                                 NULL };
+/* Maximum number of user binds. */
 
-/* Number of supplemental GIDs we can deal with. */
-#define SUPP_GIDS_MAX 128
-
-/* Maximum length of paths we're willing to deal with. (Note that
-   system-defined PATH_MAX isn't reliable.) */
-#define PATH_CHARS 4096
-
-/* Log the current UIDs. */
-#define LOG_IDS log_ids(__func__, __LINE__)
-
-/* Environment variables used for --join parameters. */
+/* Environment variables used by --join parameters. */
 char * JOIN_CT_ENV[] =  { "OMPI_COMM_WORLD_LOCAL_SIZE",
                           "SLURM_STEP_TASKS_PER_NODE",
                           "SLURM_CPUS_ON_NODE",
                           NULL };
 char * JOIN_TAG_ENV[] = { "SLURM_STEP_ID",
                           NULL };
-
-/* Timeout in seconds for waiting for join semaphore. */
-#define JOIN_TIMEOUT 30
-
-/* Variables for coordinating --join. */
-struct {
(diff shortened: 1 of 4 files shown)


bin/charliecloud.c
```diff
@@ -2,21 +2,321 @@
...
+#define PATH_CHARS 4096
+
+/* Number of supplemental GIDs we can deal with. */
+#define SUPP_GIDS_MAX 128
+
+
+/** Constants **/
+
+/* Names of verbosity levels. */
+const char * VERBOSE_LEVELS[] = { "error", "warning", "info", "debug" };
```
j-ogas: What is the reasoning behind changing lowercase verbose_levels to uppercase?

j-ogas: Cancel that.

1. Consistency. 2. Its a marco/constant.

bin/ch-run.c
```diff
@@ -6,70 +6,26 @@
...
-#define SUPP_GIDS_MAX 128
 
-/* Maximum length of paths we're willing to deal with. (Note that
-   system-defined PATH_MAX isn't reliable.) */
-#define PATH_CHARS 4096
+/** Constants and macros **/
 
-/* Log the current UIDs. */
-#define LOG_IDS log_ids(__func__, __LINE__)
+/* Maximum number of user binds. */
```
j-ogas: Can we remove this comment? USER_BINDS_MAX[] is not present in charliecloud.c and was removed from ch-run.c

bin/ch-run.c
```diff
@@ -103,191 +59,81 @@ const struct argp_option options[] = {
...
+   int arg_next;
+   int c_argc;
+   char ** c_argv;
+
    privs_verify_invoking();
-   memset(args.binds, 0, sizeof(args.binds));
-   args.container_gid = getegid();
-   args.container_uid = geteuid();
+
+   T_ (args.c.binds = calloc(1, sizeof(struct bind)));
```
j-ogas: nitpick: I remember seeing dialog about renaming TRYX a while back. I also remember being surprised to learn that the Tf_, Te_, Z_  nomenclature is some kind of standard.

Although I agree that TRYX() needed some renaming, I feel like this new syntax is unintuitive and introduces ambiguity. Maybe after opening charliecloud.h a bunch more times it will stick...?

bin/ch-run.c
```diff
@@ -6,70 +6,26 @@
...
+/* Maximum number of user binds. */
 
-/* Environment variables used for --join parameters. */
+/* Environment variables used by --join parameters. */
 char * JOIN_CT_ENV[] =  { "OMPI_COMM_WORLD_LOCAL_SIZE",
                           "SLURM_STEP_TASKS_PER_NODE",
                           "SLURM_CPUS_ON_NODE",
                           NULL };
 char * JOIN_TAG_ENV[] = { "SLURM_STEP_ID",
                           NULL };
```
j-ogas: I feel like these join environment variables should be moved to charliecloud.c. Keeping them bundled with the join macro definitions is much more intuitive and easier to read.

bin/ch-run.c
```diff
@@ -303,59 +149,16 @@ bool get_first_env(char ** array, char ** name, char ** value)
    return false;
 }
```
j-ogas: Looks like these lines need to be moved to charliecloud.c? Aren't they part of `enter_udss()`?

bin/ch-run.c
```diff
@@ -497,53 +221,53 @@ int parse_int(char * s, bool extra_ok, char * error_tag)
 /* Parse one command line option. Called by argp_parse(). */
 static error_t parse_opt(int key, char * arg, struct argp_state * state)
 {
-   struct args * as = state->input;
+   struct args * args = state->input;
```
j-ogas: I like args > as

bin/ch-run.c
```diff
@@ -103,191 +59,81 @@ const struct argp_option options[] = {
...
-void join_namespaces();
-char * join_tag();
-void log_ids(const char * func, int line);
-void run_user_command(int argc, char * argv[], int user_cmd_start);
-static error_t parse_opt(int key, char * arg, struct argp_state * state);
+int join_ct(int cli_ct);
+char * join_tag(char * cli_tag);
 int parse_int(char * s, bool extra_ok, char * error_tag);
+static error_t parse_opt(int key, char * arg, struct argp_state * state);
 void privs_verify_invoking();
```
j-ogas: Personally, I would move ch-run's prototypes to charliecloud.h.

bin/ch-run.c
reidpr: Fair enough. The reasoning is "true" versus "zero". Then you get *f* for `printf(2)`-style, *e* for without `errno`, underscore to keep them the same width.... Good feedback that they're still a bit confusing.

bin/ch-run.c
reidpr: They're in `ch-run.c` because I saw them as part of the user interface rather than containerization itself. Convincing?

bin/ch-run.c
reidpr: No, it's to gather environment variables (see above).

bin/ch-run.c
reidpr: They're in `ch-run.c` because that's the only place they're used. Prototypes in `charliecloud.h` should be applicable to anyone who uses `charliecloud.c`, which may eventually be more than `ch-run.c`.

bin/ch-run.c
reidpr: Done.

bin/ch-run.c
reidpr: Future users may want a different set of variables, if any. E.g., if we put something into Slurm we won't want to be looking at OpenMPI environment variables in that context.

PR #192, opened 2018-06-28
Add current git branch information to VERSION.full tag used by ch-run…

rstyd:
This addresses #175 . Adds git branch information to the version tag iff the current branch is not master.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,16 +21,18 @@ clean:
 # * If VERSION is an unadorned release (e.g. 0.2.3 not 0.2.3~pre), or there's
 #   no Git information available, VERSION.full is simply a copy of VERSION.
 #
-# * Otherwise, we add the Git commit and a note if the working directory
-#   contains uncommitted changes, e.g. "0.2.3~pre+ae24a4e.dirty".
+# * Otherwise, we add the Git branch if the current branch is not master, the Git commit, and a note if the working directory
+#   contains uncommitted changes, e.g. "0.2.3~pre+experimental:ae24a4e.dirty".
 #
+
 ifeq ($(shell test -d .git && fgrep -q \~ VERSION && echo true),true)
 .PHONY: VERSION.full  # depends on git metadata, not a simple file
 VERSION.full: VERSION
 	(git --version > /dev/null 2>&1) || \
           (echo "This is a Git working directory but no git found." && false)
-	printf '%s+%s%s\n' \
+	printf '%s+%s%s%s\n' \
 	       $$(cat $<) \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -21,16 +21,18 @@ clean:
 # * If VERSION is an unadorned release (e.g. 0.2.3 not 0.2.3~pre), or there's
 #   no Git information available, VERSION.full is simply a copy of VERSION.
 #
-# * Otherwise, we add the Git commit and a note if the working directory
-#   contains uncommitted changes, e.g. "0.2.3~pre+ae24a4e.dirty".
+# * Otherwise, we add the Git branch if the current branch is not master, the Git commit, and a note if the working directory
```
reidpr: Wrap to 80 characters please.

Makefile
```diff
@@ -21,16 +21,18 @@ clean:
...
+#   contains uncommitted changes, e.g. "0.2.3~pre+experimental:ae24a4e.dirty".
```
reidpr: Use a dot instead of a colon and all your wildest dreams will come true.

Makefile
```diff
@@ -21,16 +21,18 @@ clean:
...
 #
+
```
reidpr: Probably don't need a new blank line here.

Makefile
```diff
@@ -21,16 +21,18 @@ clean:
...
 ifeq ($(shell test -d .git && fgrep -q \~ VERSION && echo true),true)
 .PHONY: VERSION.full  # depends on git metadata, not a simple file
 VERSION.full: VERSION
 	(git --version > /dev/null 2>&1) || \
           (echo "This is a Git working directory but no git found." && false)
-	printf '%s+%s%s\n' \
+	printf '%s+%s%s%s\n' \
 	       $$(cat $<) \
+	       $$(git branch | grep \* | cut -d ' ' -f2 | sed 's/.*/&:/g' | sed 's/master://g') \
```
reidpr: 1. Based on the Travis results, I think a situation you may not have covered is the detached `HEAD` case (no branch at all). It’s fine with me to put in whatever there; e.g., `HEAD` is fine but you need to lose the parentheses. You can create this situation by checking out a specific commit hash that’s not on a branch.

2. Use the “plumbing” Git commands rather than ”porcelain” for this type of thing, because they’re designed to be machine-readable. I’ve identified a specific command to use that will give you just the branch and nothing else. I’d like you to practice your Google-fu to find it, and if you run into trouble, I can assist.

3. If you’re searching for a string rather than a regex, use `fgrep`. (And I recommend `egrep` instead of `grep` because the regex syntax is better.)

reidpr:
Merged, thank you!

Issue #212, opened 2018-07-19
id doesn't map properly in Charliecloud with LDAP


sameershende:
Hi,
   I do not have my id in /etc/passwd and /etc/group because we are using LDAP. When Charliecloud starts up, it can't map my ids properly and complains when I source a startup script. 
/usr/bin/id: cannot find name for group ID 1023
/usr/bin/id: cannot find name for user ID 13395

[I have no name!@host/]$ id
uid=13395 gid=1023 groups=1023,65534(nfsnobody)

Thanks,
- Sameer

reidpr:
Thanks Sameer. Can you have your admins provide more details on your setup? Specifically, technical details on how authentication and user lookup works. I'm not familiar with this area, so I need some things to Google to get me started on reproducing the problem.

reidpr:
@sameershende : Does the Charliecloud test suite pass in your environment?

sameershende:
Hi Reid,
	I apologize for the delay. The machine where I do most of the testing had an issue with Charliecloud because of the older kernel:
wget http://tau.uoregon.edu/ecp-cc.tgz; tar xf ecp-cc.tgz

[sameer@talapas-ln2 gpfs]$ ch-run ./ecp-cc -- /bin/bash
ch-run[194601]: can't init user+mount namespaces: Invalid argument (charliecloud.c:406 22)


	When I run it on another machine where it works. it works well:
[root@pegasus imagebuild]# tar xf ecp-cc.tgz
[root@pegasus imagebuild]# exit
[sameer@pegasus imagebuild]$ export PATH=$HOME/charliecloud/bin:$PATH
[sameer@pegasus imagebuild]$ ch-run ./ecp-cc -- /bin/bash
[sameer@pegasus /]$ whoami
sameer
[sameer@pegasus /]$ cd ~/
[sameer@pegasus mpi]$ mkdir t
[sameer@pegasus mpi]$ rmdir t

	I wish it were easier to set this option (namespace.unpriv_enable=1) instead sending it to the kernel command line. 
	Thanks,
	- Sameer
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html-single/getting_started_with_containers/index#user_namespaces_options

PR #220, opened 2018-08-08
bind-mount custom /etc/passwd and /etc/group

reidpr:
This addresses #212.


trandles-lanl:
This works, mostly.  When building the temporary passwd and group files you're not including UID/GID 65534 (nfsnobody) so there are still unresolved mappings.  Fix that and I think you're good to go.

Issue #145, opened 2018-04-17
style: lower-case variables in shell scripts


reidpr:
Currently the shell scripts use upper-case variable names. This is poor style because it can clash with variables used by the shell (and I have already run into this issue). Thus, change the variable names to lower case.

E.g., see: https://unix.stackexchange.com/questions/42847

mej:
Just as a warning, I've seen exactly the opposite -- lowercase variable name use conflicting with shell variables.  `csh` and derivatives use lowercase internal variables.

I'll just say a few things on this:
- Variables you export to the environment should always be uppercase.
- Uppercase variable names make the code easier to read because shell keywords and command names are almost always lowercase.  In the absence of colored syntax highlighting, lowercase variable names intermixed with lowercase everything else makes the code extremely difficult to sight-read.
- The best practice for preventing variable name conflicts, particularly in shell scripts/tools, is always the same:  Namespace them.  Pick a prefix, like `CC_` or `CHC_`, and stick with it.  That's one of the big mistakes I made early on with NHC that I'm now stuck with -- not namespacing my variables.
- Also, properly declare and scope variables via `declare`/`typeset`, `local`, and `readonly`, and only export the ones you must.

reidpr:
> Just as a warning, I've seen exactly the opposite -- lowercase variable name use conflicting with shell variables. csh and derivatives use lowercase internal variables.

I spent a little bit of time on Google just now trying to see if there's a consensus on the matter. The above appears to be a minority view; e.g.:

* https://unix.stackexchange.com/questions/42847
* https://stackoverflow.com/questions/673055
* https://google.github.io/styleguide/shell.xml#Naming_Conventions
* http://wiki.bash-hackers.org/scripting/style#variable_names

all support the lowercase convention. I couldn't find anything supporting the upper-case convention.

What have I missed?

> * Variables you export to the environment should always be uppercase.

Agreed.

>  * Uppercase variable names make the code easier to read because shell keywords and command names are almost always lowercase. In the absence of colored syntax highlighting,

I'm a little skeptical of this argument because it seems to me that the situations where one needs to read shell code in the absence of syntax coloring should be somewhat unusual (e.g., don't we have `vim` or `emacs` pretty much everywhere.

Also, wouldn't this apply to every scripting language, e.g. Perl and Python too?

[...]

PR #221, opened 2018-08-09
Lowercase script variables

j-ogas:
This addresses #145.

### DIFF
--- a/bin/ch-tar2dir
+++ b/bin/ch-tar2dir
@@ -22,8 +22,8 @@ fi
 if [ $# -lt 2 ]; then
     usage
 fi
-tarball="$1"
-newroot="${2}/$(basename "${tarball%.tar.gz}")"
+tarball=$1
+newroot=${2}/$(basename "${tarball%.tar.gz}")
 
 sentinel=WEIRD_AL_YANKOVIC
 
@@ -51,16 +51,16 @@ fi
 mkdir "$newroot"
 echo 'This directory is a Charliecloud container image.' > "${newroot}/${sentinel}"
 # Use a pipe because PV ignores arguments if it's cat rather than PV.
-size=$(stat -c%s "${tarball}")
-  pv_ -s "$size" < "${tarball}" \
+size=$(stat -c%s "$tarball")
+  pv_ -s "$size" < "$tarball" \
 | gzip_ -dc \
-| tar x$verbose -C "${newroot}" -f - \
+| tar x$verbose -C "$newroot" -f - \
       --anchored --exclude='dev/*' --exclude='./dev/*'
 # Make all directories writeable so we can delete image later (hello, Red Hat).
-find "${newroot}" -type d -a ! -perm /200 -exec chmod u+w {} +
+find "$newroot" -type d -a ! -perm /200 -exec chmod u+w {} +
 
 # Ensure directories that ch-run needs exist.
 mkdir -p "${newroot}/dev"
-for i in $(seq 0 9); do mkdir -p "${newroot}/mnt/$i"; done
+for i in $(seq 0 9); do mkdir -p "${newroot}/mnt/${i}"; done
 
 echo "${newroot} unpacked ok"
--- a/bin/ch-tar2dir
+++ b/bin/ch-tar2dir
@@ -16,51 +16,51 @@ EOF
 parse_basic_args "$@"
 
 if [ "$1" = --verbose ]; then
-    VERBOSE=v
+    verbose=yes
     shift
 fi
 if [ $# -lt 2 ]; then
     usage
 fi
-TARBALL="$1"
-NEWROOT="$2/$(basename "${TARBALL%.tar.gz}")"
+tarball="$1"
+newroot="${2}/$(basename "${tarball%.tar.gz}")"
 
-SENTINEL=WEIRD_AL_YANKOVIC
+sentinel=WEIRD_AL_YANKOVIC
 
 # Is the tarball a regular file (or symlink) and readable?
-if [ ! -f "$TARBALL" ] || [ ! -r "$TARBALL" ]; then
-    echo "can't read $TARBALL" 1>&2
+if [ ! -f "$tarball" ] || [ ! -r "$tarball" ]; then
+    echo "can't read ${tarball}" 1>&2
     exit 1
 fi
 
-if [ ! -d "$NEWROOT" ]; then
-    echo "creating new image $NEWROOT"
+if [ ! -d "$newroot" ]; then
+    echo "creating new image ${newroot}"
 else
-    if    [ -f "$NEWROOT/$SENTINEL" ] \
-       && [ -d "$NEWROOT/bin" ] \
-       && [ -d "$NEWROOT/lib" ] \
-       && [ -d "$NEWROOT/usr" ]; then
-        echo "replacing existing image $NEWROOT" 1>&2
-        rm -Rf --one-file-system "$NEWROOT"
+    if    [ -f "${newroot}/${sentinel}" ] \
+       && [ -d "${newroot}/bin" ] \
+       && [ -d "${newroot}/lib" ] \
+       && [ -d "${newroot}/usr" ]; then
+        echo "replacing existing image ${newroot}" 1>&2
+        rm -Rf --one-file-system "${newroot}"
     else
-        echo "$NEWROOT exists but does not appear to be an image" 1>&2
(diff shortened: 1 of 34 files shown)


test/run/ch-run_misc.bats
```diff
@@ -2,20 +2,20 @@ load ../common
 
 @test 'relative path to image' {  # issue #6
     scope quick
-    DIRNAME=$(dirname "$CHTEST_IMG")
-    BASEDIR=$(basename "$CHTEST_IMG")
-    cd "$DIRNAME" && ch-run "$BASEDIR" -- true
+    dirname=$(dirname "$chtest_img")
```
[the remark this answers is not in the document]
j-ogas: Good catch!

test/run/ch-run_uidgid.bats
```diff
@@ -2,51 +2,51 @@ load ../common
 
 setup () {
     scope standard
-    if [[ -n $GUEST_USER ]]; then
```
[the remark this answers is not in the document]
j-ogas: Yes, it is exported in test Makefile.

bin/ch-tar2dir
```diff
@@ -9,58 +9,58 @@ Unpack a Docker export tarball into a directory.
...
-  $ $(basename "$0") TARBALL DIR
+  $ $(basename "$0") tarball DIR
 EOF
 )
 
 parse_basic_args "$@"
 
 if [ "$1" = --verbose ]; then
-    VERBOSE=v
+    verbose=v
```
reidpr: Change this to `verbose=yes` to match style elsewhere. (Note: boolean variables in shells scripts are tricky. We use the convention empty/unset=false, "yes"=true.)

examples/mpi/lammps/test.bats
```diff
@@ -38,25 +38,24 @@ setup () {
     scope full
     prerequisites_ok lammps
     multiprocess_ok
-    IMG=$IMGDIR/lammps
```
reidpr: \o/

examples/mpi/mpihello/test.bats
```diff
@@ -11,43 +11,43 @@ count_ranks () {
...
     # shellcheck disable=SC2086
-    run $MPIRUN_CORE ch-run --join "$IMG" -- /hello/hello
+    run $mpirun_core ch-run --join "$ch_img" -- /hello/hello
     echo "$output"
     [[ $status -eq 0 ]]
     rank_ct=$(count_ranks "$output")
-    echo "found $rank_ct ranks, expected $CHTEST_CORES_TOTAL"
-    [[ $rank_ct -eq "$CHTEST_CORES_TOTAL" ]]
+    echo "found $rank_ct ranks, expected $chtest_cores_total"
+    [[ $rank_ct -eq "$chtest_cores_total" ]]
```
reidpr: Isn't this `ch_cores_total` now? In fact looks like this whole file needs to be updated.

examples/serial/obspy/test.bats
```diff
@@ -3,16 +3,15 @@ load ../../../test/common
 setup () {
     scope skip  # issue #64
     prerequisites_ok obspy
-    IMG=$IMGDIR/obspy
 }
 
-@test "$EXAMPLE_TAG/runtests" {
+@test "$example_tag/runtests" {
```
reidpr: Should be `ch_tag`.

examples/serial/spack/test.bats
```diff
@@ -4,25 +4,25 @@ setup() {
     scope full
     [[ -z $CHTEST_CRAY ]] || skip 'issue #193 and Spack issue #8618'
     prerequisites_ok spack
-    SPACK_IMG="$IMGDIR/spack"
+    ch_img="$CH_TEST_IMGDIR/spack"
```
reidpr: Can use the one from `common.bash` instead?

test/chtest/Build
```diff
@@ -16,44 +16,44 @@
```
reidpr: Should probably add this file to the ShellCheck test.

test/run/ch-fromhost.bats
```diff
@@ -26,194 +26,194 @@ fromhost_ls () {
 @test 'ch-fromhost (Debian)' {
     scope standard
     prerequisites_ok debian9
-    IMG=$IMGDIR/debian9
+    img=$ch_imgdir/debian9
```
reidpr: Should use the one from `common.bash`?

test/run/ch-run_misc.bats
```diff
@@ -2,20 +2,20 @@ load ../common
 
 @test 'relative path to image' {  # issue #6
     scope quick
-    DIRNAME=$(dirname "$CHTEST_IMG")
-    BASEDIR=$(basename "$CHTEST_IMG")
-    cd "$DIRNAME" && ch-run "$BASEDIR" -- true
+    dir=$(dirname "$ch_imgtest")
```
reidpr: Perhaps `dirname_` and `basename_`? `/bin/dir` is also a command that you don't want to shadow. Or, you might simply eliminate the variables.

test/run/ch-run_misc.bats
```diff
@@ -119,11 +119,11 @@ EOF
 # shellcheck disable=SC2016
 @test '$PATH: unset' {
     scope standard
-    BACKUP_PATH=$PATH
+    backup_path=$PATH
```
reidpr: Call this `old_path` for consistency with `old_ifs` elsewhere.

test/run/ch-run_misc.bats
```diff
@@ -155,169 +155,169 @@ EOF
...
     [[ $status -eq 1 ]]
-    r="can't bind .+/bind2 to $CHTEST_IMG/goops: No such file or directory"
+    r="can't bind .+/bind2 to $ch_imgtest/goops: No such file or directory"
     [[ $output =~ $r ]]
 }
 
 @test 'broken image errors' {
     scope standard
-    IMG="$BATS_TMPDIR/broken-image"
+    img="$BATS_TMPDIR/broken-image"
```
reidpr: If we use `img` for the one in `common.bash`, this should be something different to not shadow it — maybe?

test/common.bash
```diff
@@ -133,75 +133,75 @@ chmod 700 "$btnew"
...
 
+# Separate directories for tarballs and images
+ch_imgdir=$CH_TEST_IMGDIR
+ch_tardir=$CH_TEST_TARDIR
+
 # Some test variables
-EXAMPLE_TAG=$(basename "$BATS_TEST_DIRNAME")
-EXAMPLE_IMG=$IMGDIR/$EXAMPLE_TAG
-CHTEST_TARBALL=$TARDIR/chtest.tar.gz
-CHTEST_IMG=$IMGDIR/chtest
```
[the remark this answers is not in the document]
j-ogas: >Because this one is used so pervasively, I think we should consider calling it img without the prefix, as a special case. I'd like to hear your opinion, though. If we agree on that, you could add a comment to that effect.

I am reluctant to use `img` as the special case variable for `ch_chtest_img` because `img` appears to be used as a local variable for other scripts, e.g., `test/chtest/Build`. Although that particular script doesn't load `common.bash` and thus there are no conflicts, I would prefer to keep the `ch_chtest_img` variable unique. Perhaps we should use a special prefix? Something like `cht_img`, as you had suggested in the past?

test/common.bash
```diff
@@ -133,75 +133,75 @@ chmod 700 "$btnew"
...
+    ch_cray=yes
 else
-    CHTEST_CRAY=
+    ch_cray=
 fi
 
+# Separate directories for tarballs and images
+ch_imgdir=$CH_TEST_IMGDIR
+ch_tardir=$CH_TEST_TARDIR
+
```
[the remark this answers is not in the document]
j-ogas: Readability. Previously, these two variable declarations were between a couple lines of bats temporary directory voodoo and ch_mpi variable assignments. 

It seemed to me that these these insulating variables should be declared directly above where they are tapped ('test variable declaration' section). This change was largely intuitive, it is entirely possible that I missed a detail that would compel me to keep them where they were.

test/run/ch-fromhost.bats
j-ogas: ch_img uses `ch_tag=$(basename "$BATS_TEST_DIRNAME")`, which I think for ch-fromhost.bats will be `run` (correct me if I am wrong). I think the local img variable is appropriate here.

PR #223, opened 2018-08-15
Fixed missing provision command

rstyd:
Went through the code and noticed this issue today.

### DIFF
--- a/doc-src/vm.rst
+++ b/doc-src/vm.rst
@@ -239,7 +239,7 @@ The most important difference with this build procedure is that a second user
 ::
 
    $ CH_VERSION=v0.9.1 vagrant up
-   $ vagrant --provision-with=ova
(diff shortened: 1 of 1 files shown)


rstyd:
Whoops. Yeah no problem.

mej:
It doesn't really make sense to me that `vagrant up` be run twice in a row, so I think either @rstyd *is* correct that the command should be `vagrant provision --provision-with=ova`, or the `--provision-with=ova` directive belongs on the preceding `vagrant up` command.

Is there something I'm missing?

reidpr:
Aha; the problem here is that I was ignorant of the `vagrant provision` command and gave Rusty bad guidance. Michael, thanks for the correction, and Rusty, I apologize. I'll just make a quick commit on master to fix this.

Issue #227, opened 2018-08-23
hard-code gcc for test/sotest binaries


reidpr:
Building the binaries in `test/sotest` with certain compilers causes the tests to fail with missing shared libraries. For example, with Intel compiler, the Intel shared libraries are not present inside the container:

```
$ ldd libsotest.so.1
	linux-vdso.so.1 =>  (0x00007fff825fb000)
	libc.so.6 => /usr/lib64/libc.so.6 (0x00002b7396fb1000)
	libimf.so => not found
	libsvml.so => not found
	libirng.so => not found
	libm.so.6 => /usr/lib64/libm.so.6 (0x00002b7397376000)
	libgcc_s.so.1 => /usr/lib64/libgcc_s.so.1 (0x00002b7397678000)
	libintlc.so.5 => not found
	libdl.so.2 => /usr/lib64/libdl.so.2 (0x00002b739788f000)
	/lib64/ld-linux-x86-64.so.2 (0x000055ef5768a000)

```

Because these binaries will be deployed in a predictable environment, we'll hard-code `gcc` in `test/Makefile` instead of using the normally more appropriate `$(CC)`.

PR #229, opened 2018-08-24
test/Makefile: hardcode gcc

reidpr:
This addresses issue #227.


reidpr:
@j-ogas, will you (a) try this in the places you're having trouble, and (b) check in with your team to make sure this isn't doing something dumb compiler wise?

j-ogas:
No feedback from the team on this PR yet but the branch does run `ch-fromhost` without tickling the sotest bug on both cray and non-cray systems. FWIIW I think it is good to merge.

Issue #239, opened 2018-09-13
Unavailable PSM2 version resulting in Openmpi test image build failing


heasterday:
The current libpsm2 version used in Dockerfile.openmpi (10.3.58-1) is no longer available from http://http.us.debian.org/debian/pool/main/libp/libpsm2/ this results in all the OpenMPI docker image builds breaking. Recommend, if possible, incrementing from version 10.3.58-1 to 10.3.58-2 within Dockerfile.openmpi on line 142 which on my system resulted in successful image builds.

reidpr:
Good find! Can you please submit a PR?

PR #240, opened 2018-09-13
Incremement version of PSM2 in Dockerfile.openmpi

heasterday:
Incrementing version of PSM2 to resolved openmpi image build failures. This will resolve issue #239.


reidpr:
I'm pretty sure the test failures are not related to this PR but rather to a new version of Sphinx. Therefore, I'll merge.

PR #243, opened 2018-09-19
update apk-tools-static version

j-ogas:
old link is broken.

Issue #242, opened 2018-09-19
apk-tools-static-2.7.5-r0.apk — 404 error


j-ogas:
`chtest/Build`'s apk tools package version is out of date.

PR #245, opened 2018-09-20
Auto find the apk-tools-static version using wget

PaulUTK:
Looks like this is a consistent issue when apk-tools gets rev'd. I'm proposing a solution that will always find the latest version and populate the variable with it instead of hard coding it.

### DIFF
--- a/test/chtest/Build
+++ b/test/chtest/Build
@@ -21,7 +21,7 @@ TARBALL=$2
 TARBALL_UNCOMPRESSED=${TARBALL%.*}
 WORKDIR=$3
 MIRROR=http://dl-cdn.alpinelinux.org/alpine/v3.6
-APK_TOOLS=apk-tools-static-2.7.5-r0.apk
+APK_TOOLS=$(wget --quiet -O - http://dl-cdn.alpinelinux.org/alpine/v3.6/main/x86_64/ | grep apk-tools-static | sed -e 's/<[^>]*>//g' | awk '{print $1}')
 IMG="$WORKDIR/img"
 
 cd $WORKDIR
--- a/test/chtest/Build
+++ b/test/chtest/Build
@@ -21,7 +21,7 @@ TARBALL=$2
 TARBALL_UNCOMPRESSED=${TARBALL%.*}
 WORKDIR=$3
 MIRROR=http://dl-cdn.alpinelinux.org/alpine/v3.6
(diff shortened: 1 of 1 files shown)


trandles-lanl:
Is there a reason for checking v3.6 of alpine instead of latest-stable?

PaulUTK:
Agreed, we may also want to consider doing an arch check as well in the future for these tests. I notice a lot of hardcoded x86_64 paths and the ARMs are coming...

trandles-lanl:
FWIW, I think the check implemented in the commit looks fine to me.  It works when I tested it.  @mej has sharper shell scripting skills than I do so I await his evaluation of the syntax.  The only minor quibble I have is that wget is less likely to be available everywhere.  For instance, on RHEL-like systems (CentOS & Fedora in particular), curl is installed by default instead of wget.

RE: architecture check...  add "uname -p" where appropriate?

PaulUTK:
True, want me to find a curl equivalent and also add the latest-stable as the path?

PaulUTK:
curl -s http://dl-cdn.alpinelinux.org/alpine/latest-stable/main/x86_64/ | grep apk-tools-static | sed -e 's/<[^>]*>//g' | awk '{print $1}'
apk-tools-static-2.10.1-r0.apk

trandles-lanl:
Sure, finding a curl equivalent sounds good.  I'm not sure if `uname -p` or `uname -i`, both of which are noted as "non-portable," is the correct way to get architecture.

trandles-lanl:
Using `arch` is good with me.  It's easier to read and understand than uname with a flag as well.

I had overlooked `uname -m` which doesn't have the non-portable caveat.

PaulUTK:
Hmm it looks like the latest-stable breaks the tests.

ok 91 ch-tar2dir: unpack image
ok 92 ch-tar2dir: /dev cleaning
ok 93 ch-tar2dir: errors

reidpr:
Let's keep this PR to a single topic. I'll create different issues for the Alpine version and arch business.

As for `wget`, that's the documented dependency for the test suite, so please stick with it. There is a comment somewhere explaining why, that probably should be made easier to find. I'm open to argument to change this to `curl`, but propose that in a separate issue.

reidpr:
This is also an alternative to PR #243 to solve issue #242.

reidpr:
Merged, thank you!

Issue #219, opened 2018-08-08
refactor: wrapper for mounting stuff under the new root


reidpr:
We have a ton of sequences similar to:

```
T_ (1 <= asprintf(&path, "%s%s", c->newroot, DEFAULT_BINDS[i]));
Zf (mount(DEFAULT_BINDS[i], path, NULL, MS_REC|MS_BIND|MS_RDONLY, NULL),
    "can't bind %s to %s", (char *) DEFAULT_BINDS[i], path);
```

and

```
T_ (1 <= asprintf(&path, "%s/home", c->newroot));
Zf (mount(NULL, path, "tmpfs", 0, "size=4m"),
    "can't mount tmpfs at %s", path);
```

Factor these out into a function.

Issue #26, opened 2017-07-18
Move default bind mounts to args.binds


j-ogas:
Perhaps down the line, ch-run.c lines 195 - 199 should be moved to args.binds

reidpr:
I think the right approach may now be to make `charliecloud.c:DEFAULT_BINDS` into an array of `struct bind`, just like `args.c.binds`, and factor out the loop into a function?

PR #260, opened 2018-10-02
refactor mount(2) related code

reidpr:
This addresses #26 and #219.

It depends on the `cray-mpich` branch. Expect some rebasing and force-pushing.


reidpr:
Yes, we do care, but IMO that portability issue is out of scope for this PR.

Issue #259, opened 2018-10-02
CH_BIN and LIBEXEC in scripts


reidpr:
The variables `CH_BIN` and `LIBEXEC` in shell scripts are upper case. To follow up #145, it seems to me that they should also be converted to lower case?

j-ogas:
These two particular variables were skipped due to ShellCheck [SC2154](https://github.com/koalaman/shellcheck/wiki/SC2154).  

`bin/ch-build` and `bin/ch-fromhost`both use a non-local variable (`ch_bin`) declared in `bin/bash.sh`. Shellcheck SC2154 flags this as an error because it cannot detect that the variable has been assigned. 

Possible solutions: 1) We can disable this directive with `-e SC2154` or 2) check if the variable is assigned at runtime (or fail) with `${ch_bin:?}`. FWIIW, I prefer the latter.

PR #261, opened 2018-10-04
Convert CH_BIN and LIBEXEC to lower case variables

j-ogas:
Per #259.

### DIFF
--- a/bin/ch-build2dir
+++ b/bin/ch-build2dir
@@ -1,7 +1,7 @@
 #!/bin/sh
 
-LIBEXEC="$(cd "$(dirname "$0")" && pwd)"
-. "${LIBEXEC}/base.sh"
+libexec="$(cd "$(dirname "$0")" && pwd)"
+. "${libexec}/base.sh"
 
 # shellcheck disable=SC2034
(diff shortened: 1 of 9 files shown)


bin/ch-build2dir
```diff
@@ -35,7 +35,7 @@ tag=$(basename "$PWD")
 
 set -x
 
-"$CH_BIN"/ch-build -t "$tag" "$context" "$@"
-"$CH_BIN"/ch-docker2tar "$tag" "$dest"
-"$CH_BIN"/ch-tar2dir "${dest}/${tag}.tar.gz" "$dest"
+"${ch_bin:?}"/ch-build -t "$tag" "$context" "$@"
```
reidpr: 1. So in principle the `${foo:?}` seems appealing, but we don't use it anywhere else, and in the cases in this PR you're going to get an immediate failure anyway. I think we should just stick with `${foo}` and disable SC2154 in the BATS test.

2. Please change to `"${ch_bin}/ch-build"` (i.e., entire path inside quotation marks) throughout the PR.

j-ogas:
101ab23 addresses the requested changes.

Issue #263, opened 2018-10-05
Add docker pull convenience commands


j-ogas:
It would be nice to simplify our dockerhub interactions with scripts that automate the workflow. For example, if want to use an image from dockerhub we would do something like the following: 

```
$docker pull $img
$ch-docker2tar $img /var/tmp/ch-tarballs/
$ch-tar2dir /var/tmp/ch-tarballs/$img.tar.gz /var/tmp/ch-images/
```

PR #201, opened 2018-07-09
Adds ch-pull2dir and ch-pull2tar convenience commands

ivotron:
Shortcuts for `docker pull` followed by `ch-docker2tar` (`ch-pull2tar`) and `ch-tar2dir` (`ch-pull2dir`)

PR #264, opened 2018-10-09
add docker pull convenience commands

j-ogas:
Extends #201
Addresses #263


doc-src/ch-pull2dir_desc.rst
```diff
@@ -0,0 +1,59 @@
...
+::
+
+  $ ch-pull2dir IMAGE[:TAG] DIR
+
+Description
+===========
+
+Pulls a Docker image named :code:`IMAGE[:TAG]` from Docker Hub, flattens it
+into a Charliecloud tarball and extracts the tarball into a subdirectory of 
+:code:`DIR`.
```
reidpr: Style: Write in the imperative mood, giving instructions to the computer rather than describing what is done. For example, here you'd want:

```
Pull a Docker image named :code:`IMAGE[:TAG]` from Docker Hub, flatten it
into a Charliecloud tarball, and extract ...
```

(Note also addition of Oxford comma.)

doc-src/ch-pull2dir_desc.rst
```diff
@@ -0,0 +1,59 @@
...
+
+Inside :code:`DIR`, the subdirectory will be created whose name corresponds to
+the name of the tarball with the :code:`.tar.gz` suffix removed. If such a
+directory exists already and appears to be a Charliecloud container image, it
+is removed and replaced. If the existing directory doesn't appear to be a
+container image, the script aborts with an error.
+
+Sudo privileges are required to run the :code:`docker pull` command.
```
reidpr: There's a fair amount of copy and paste description here. I would refer people to the underlying command documentation instead.

For example, the man page for `ch-build2dir` is a bit stale, but it's still a decent model.

test/build.bats
```diff
@@ -137,6 +137,31 @@ load common
     rm -Rf --one-file-system "$tar" "$img"
 }
 
+@test 'ch-pull2tar' {
+    scope standard
+    need_docker
+    tag='alpine:latest'
```
reidpr: Use the same version of Alpine we're using elsewhere.

test/build.bats
```diff
@@ -137,6 +137,31 @@ load common
...
+    tar="${ch_tardir}/${tag}.tar.gz"
+    ch-pull2tar "$tag" "$ch_tardir"
+    [[ $status -eq 0 ]]
+    [[ -e $tar ]] 
+    rm "${ch_tardir}/${tag}.tar.gz"
```
reidpr: Explain in comment why we're deleting the image instead of keeping it around to test.

Or maybe we should keep these around to test? They could be renamed to avoid clashing with the other tarball from the same source. Does the existing framework pick up tarballs added to `$ch_tardir` like this?

test/build.bats
```diff
@@ -137,6 +137,31 @@ load common
     rm -Rf --one-file-system "$tar" "$img"
 }
 
+@test 'ch-pull2tar' {
+    scope standard
+    need_docker
+    tag='alpine:latest'
+    tar="${ch_tardir}/${tag}.tar.gz"
```
reidpr: Might use `$CH_TEST_TMPDIR` rather than `$ch_tardir`.

test/build.bats
```diff
@@ -137,6 +137,31 @@ load common
...
+@test 'ch-pull2dir' {
+    scope standard
+    # This test unpacks in $ch_tardir and is removed upon completion.
+    need_docker
+    tag='alpine:latest'
+    img="${ch_tardir}/${tag}"
+    ch-pull2dir "$tag" "$ch_tardir"
+    [[ status -eq 0 ]]
+    [[ -d $img ]]
+    rm -Rf --one-file-system "$img"
```
reidpr: So I'm a bit queasy that this just deletes the directory image instead of testing it, but `ch-build2dir` has the same problem. Maybe a new issue to test both.

j-ogas: I agree. We should probably do a `ch-run $img -- /bin/true` at the very least. That said, I think it should probably be a new issue to address both tests.

test/build.bats
j-ogas: I do think it is a bit awkward to unpack to our tarball directory instead of a temp directory. However, since we do this for both `ch-pull2dir` and `ch-build2dir`, we should probably make a new issue to address both if we decide that is the approach we want to take.

test/build.bats
j-ogas: Commit 4f3b126 adds a comment similar to that of `ch-build2dir`. I am uncertain that it is the correct approach.

I am unsure if the image directories from `ch-pull2dir` and `ch-build2dir` should persist after these specific tests. I think they should be tested with `ch-run` at the very least before being removed. What do you think?

Yes, the existing framework does pick up tarballs with `tarball=$ch_tardir/%(tag)s.tar.gz`.

test/build.bats
reidpr: Yes, let's solve both separately.

reidpr:
Is this ready for re-review and merge?

j-ogas:
The PR has been rebased and squashed. Provided that 1) the travis tests pass and 2) we will handle the addition of image tests for both `ch-build2dir` and `ch-pull*` in a separate PR, it is ready for review.

j-ogas:
> 1) the travis tests pass

Tests failed with 404 errors. Restarting them now. If error persists I'll poke around.

reidpr:
I think Travis may be having some trouble — I'm getting similar errors on the master too.

j-ogas:
The restart worked apparently. Ready for review.

Issue #198, opened 2018-07-05
new script ch-tar2dir to tar up an existing image directory


reidpr:
This issue proposes a new script `ch-dir2tar` that tars up an existing image directory (created by whatever means) so that it works well with `ch-tar2dir`, e.g. by ensuring there's no top-level directory other than dot.

It was suggested in issue #150 and may also help with issue #154.

reidpr:
So I should have been more clear: I'm skeptical we should do this. The alternative is that the other scripts work better with images packed up with plain `tar`, which seems simpler to me. The purpose of the issue was to come to a decision.

I'll need to be convinced it's a good idea before accepting any related PRs.

j-ogas:
FWIW, I had encountered #150 while trying to create a tarball of an altered image after a`ch-fromhost` injection. Although the instructions provided in the FAQ  are effective, the solution is unintuitive and seemingly clunky. Seems problematic that both @trandles and I experienced this same error using `tar` in a natural manner. 

If that is not convincing enough, consider that the issue from #150 is one that is not easily detected; the tarball is created without warnings or error. It isn't until we unpack the tarball and try to execute `ch-run` that a problem becomes apparent. It is reasonable to imagine a situation where a user uses`tar` in a similar way and shares a bogus image. 

If nothing else, I think we should consider providing this mechanism as an optional argument to `ch-fromhost`. It would be nice to inject files into an image directory and automatically pack the altered image into a verified tarball.

reidpr:
My current preference would be to implement #199, which would make this script unnecessary, because you could just use standard `tar` in the usual way.

The last paragraph above is a reasonable feature proposal, but put it in a new issue.

rstyd:
I've run into this problem while trying to tar img unpack directories. Imo, we should include tar2dir until #199 is implemented. The solution is simple but incredibly unintuitive as a longtime tar user.

reidpr:
Closing as `WONTFIX`, since PR #198 makes it easy to just use plain `tar` in the traditional way.

Issue #150, opened 2018-04-27
manual tar then ch-tar2dir difficulties


trandles-lanl:
I wanted to do some work in a Charliecloud container and then tar up the results to use later.  I got the following:

```
[trandles@cn812 ~]$ ch-tar2dir /scratch/beedev/arm64v8.gcc.tar.gz /var/tmp
creating new image /var/tmp/arm64v8.gcc
/var/tmp/arm64v8.gcc unpacked ok
[trandles@cn812 ~]$ ch-run -w /var/tmp/arm64v8.gcc/ -- /bin/bash
trandles@cn812:/$ ls
WEIRD_AL_YANKOVIC  boot  etc   lib    mnt  proc  run   srv  tmp  var
bin                dev   home  media  opt  root  sbin  sys  usr
trandles@cn812:/$ git clone https://github.com/hpc/charliecloud.git
Cloning into 'charliecloud'...
remote: Counting objects: 3078, done.
remote: Compressing objects: 100% (7/7), done.
remote: Total 3078 (delta 0), reused 1 (delta 0), pack-reused 3071
Receiving objects: 100% (3078/3078), 3.02 MiB | 2.06 MiB/s, done.
Resolving deltas: 100% (2011/2011), done.
trandles@cn812:/$ ls
WEIRD_AL_YANKOVIC  boot          dev  home  media  opt   root  sbin  sys  usr
bin                charliecloud  etc  lib   mnt    proc  run   srv   tmp  var
trandles@cn812:/$ exit
exit
[trandles@cn812 ~]$ cd /var/tmp
[trandles@cn812 tmp]$ tar -zcf /scratch/beedev/arm64v8.gcc.charliecloud.tar.gz arm64v8.gcc/
[trandles@cn812 tmp]$ cd
[trandles@cn812 ~]$ rm -rf /var/tmp/arm64v8.gcc/
[trandles@cn812 ~]$ ch-tar2dir /scratch/beedev/arm64v8.gcc.charliecloud.tar.gz /var/tmp
creating new image /var/tmp/arm64v8.gcc.charliecloud
/var/tmp/arm64v8.gcc.charliecloud unpacked ok
[trandles@cn812 ~]$ ch-run -w /var/tmp/arm64v8.gcc.charliecloud/ -- /bin/bash
ch-run: can't bind /etc/passwd to /ram/var/tmp/arm64v8.gcc.charliecloud/etc/passwd: No such file or directory (ch-run.c:203 2)
[...]

PR #266, opened 2018-10-10
add new command ch-dir2tar

j-ogas:
Addresses #198. Related to #150 and #200.

Tests still need to be added, however, it is not immediately clear to me where they belong.


j-ogas:
Closing due to further discussion on #198.

PR #270, opened 2018-10-19
document locations of .bats files

reidpr:


rstyd:
Exactly what I was looking for!

Issue #267, opened 2018-10-12
'ch-fromhost --nivida with GPU' test failure


j-ogas:
```bash
 ✗ ch-fromhost --nvidia with GPU
   (in test file run/ch-fromhost.bats, line 337)
     `[[ $status -eq 127 ]]' failed
```

The test above expects `ch-run $imgdir/nvidia -- /matrixMulCUBLAS` to fail. I suspect this is because the function `fromhost_clean`, which is called prior to `ch-run`, initially removed files injected with `ch-fromhost` that matrixMulCUBLAS once depended on. At a quick glance, I am uncertain whether the new two-stage nvidia dockerfile build made this portion of the test obsolete, or the `fromhost_clean` function needs revising (aside from the removal of the obsolete line  7: `/usr/local/cuda-9.1/targets/x86_64-linux/lib/libsotest.so.1{.0,}`).

j-ogas:
The `fromhost_clean` function does not remove the nvidia shared objects injected with `ch-fromhost --nvidia`. We can account for these objects by adding `/usr/local/lib/libcuda.so*` and `/usr/local/lib/libnvidia-*` to this list of targets in fromhost_clean. If this is a reasonable approach I'll go ahead and submit a PR.

reidpr:
That sounds good to me.

PR #281, opened 2018-11-06
add /usr/sbin to is_bin function

j-ogas:
This PR is to address the case where `nvidia-container-cli list --binaries --libraries` finds binaries in `/usr/sbin`. `ch-fromhost` will now include these binaries inside the container `/usr/bin` directory.


j-ogas:
I'll add that this PR doesn't fail the test suite with #267.

Issue #287, opened 2018-11-21
test/make-auto needs cleanup


reidpr:
`test/make-auto` is pretty ugly. It needs tidying and refactoring.

reidpr:
Moot given #584. Closing.

Issue #235, opened 2018-08-31
change all Dockerfiles to use context of .


reidpr:
Proposed at the end of question issue #225's discussion is changing the context for the hello example to just `examples/serial/hello` rather than the root of the source code. This would change the tutorial example from:

```
$ cd /usr/local/src/charliecloud/examples/serial/hello
$ ch-build -t hello /usr/local/src/charliecloud
```

to:

```
$ cd /usr/local/src/charliecloud/examples/serial/hello
$ ch-build -t hello .
```

The argument is that this example would closer match Docker doing the same thing.

This issue is to decide whether to do it, and if so make the change.

j-ogas:
FWIIW, I vote to move the context to `examples/serial/hello`.

reidpr:
I am still not convinced, but moving to 0.9.4 milestone so we discuss and come to a decision.

reidpr:
I'm going to tag this `deferred` unless more argument for doing it arises soon.

reidpr:
All our Dockerfiles; checks indicate `.` would work for the context directory:

* [ ] `examples/serial/hello/Dockerfile`
* [ ] `examples/serial/spack/Dockerfile`
* [ ] `examples/serial/obspy/Dockerfile`
* [ ] `examples/mpi/lammps/Dockerfile`
* [ ] `examples/mpi/mpihello/Dockerfile`
* [ ] `examples/mpi/mpihello-mpich/Dockerfile`
* [ ] `examples/mpi/mpibench/Dockerfile`
* [ ] `examples/mpi/mpibench-mpich/Dockerfile`
* [ ] `examples/mpi/paraview/Dockerfile`
* [ ] `examples/other/spark/Dockerfile`
* [ ] `test/Dockerfile.alpine36`
* [ ] `test/Dockerfile.centos7`
* [ ] `test/Dockerfile.openmpi`
* [ ] `test/Dockerfile.debian9`
* [ ] `test/Dockerfile.mpich`
* [ ] `test/Dockerfile.nvidia`
* [ ] `test/Dockerfile.alpineedge`
* [ ] `test/Dockerfile.centos6`
* [ ] `test/Dockerfile.ubuntu1604`
* [ ] `test/Dockerfile.python3`

PR #288, opened 2018-11-21
refactor tests & examples to use context directory .

reidpr:
Addresses #235.

This also makes it simpler to have two flavors of an example; see `mpihello` and `mpibench`, which now have a single directory with two Dockerfiles.

Follow-up work in #287.


reidpr:
This will need testing on a Cray.

PR #301, opened 2018-12-05
update ch-fromhost.bats for --nvidia with GPU test failure

j-ogas:


reidpr:
Prior PR #290 to fix #267 was incomplete. This PR completes the fix.

Issue #199, opened 2018-07-05
smarten ch-tar2dir to disregard top-level directories


reidpr:
`ch-tar2dir` could be enhanced to disregard top-level directories, e.g. with `--strip-components`. This would make things less strict about how to tar up an existing image.

Suggested in issue #150. May also solve #158.

Disadvantage: Can we tell whether there is a top-level directory without reading the entire tarball? If not, this may be a performance issue.

FAQ would need to be updated.

reidpr:
Also, this should take advantage of issue #262 to barf on bad images.

Issue #158, opened 2018-05-11
deal with both presence and absence of leading ./ in tarball files


reidpr:
`docker export` tarballs do not have leading `./` in filenames:

```
$ tar tvf debian9.tar.gz | head -5
-rwxr-xr-x 0/0               0 2018-05-11 09:21 .dockerenv
drwxr-xr-x 0/0               0 2018-03-11 18:00 bin/
-rwxr-xr-x 0/0         1099016 2017-05-15 13:45 bin/bash
-rwxr-xr-x 0/0           35688 2017-02-22 05:23 bin/cat
-rwxr-xr-x 0/0           64424 2017-02-22 05:23 bin/chgrp
```

Normal people tar up the current directory with `tar cf .`, e.g.:

```
$ tar tvf debian9.tar.gz | head -5
-rwxr-xr-x 0/0               0 2018-05-11 09:21 .dockerenv
drwxr-xr-x 0/0               0 2018-03-11 18:00 ./bin/
-rwxr-xr-x 0/0         1099016 2017-05-15 13:45 ./bin/bash
-rwxr-xr-x 0/0           35688 2017-02-22 05:23 ./bin/cat
-rwxr-xr-x 0/0           64424 2017-02-22 05:23 ./bin/chgrp
```

We need to either deal with both ways (e.g., right now `ch-tar2dir` excludes only leading `dev/*` but not `./dev/*`) or require one or the other. If the latter, tarballs might not work across versions.

PR #318, opened 2018-12-14
ch-tar2dir: handle single top-level directories

reidpr:
Addresses #199.


reidpr:
Closes #199 and #158. Also closes #198 as `WONTFIX`.

Issue #334, opened 2019-01-17
add skopeo + umoci demo


reidpr:
As a step toward adding skopeo and/or umoci support to the scripts directly (#325), create and test the workflow in `test/Build.skopeo-umoci`.

Issue #332, opened 2019-01-17
ch-tar2dir: silent failure when tarball needs raising and contains no files starting with double dot


reidpr:
If `ch-tar2dir` is presented with a tarball with a single top-level directory that contains no files starting with a double dot, it returns failure but prints no error message.

This is because `[ -e "$f" ]` on line 109 returns failure, which is then propagated by `set -e` at the top of the file.

We missed it because `test/chtest/Build` does create such a file, in order to test that it's properly raised. Currently, the no hidden files case is tested by `test/Build.skopeo-umoci`.

PR #335, opened 2019-01-17
demo skopeo + umoci workflow, supporting workflow

reidpr:
This PR is a little quirky because it both adds a new feature (#334) and fixes a bug (#332), the latter being necessary to test the former.

Thus, at merge time I'll rebase and merge instead of squashing in order to preserve the two mostly unrelated commits.

@samcmill, I'd love your review on this as well, if you have time.

### DIFF
--- a/test/Build.skopeo-umoci
+++ b/test/Build.skopeo-umoci
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+# Build an Alpine Linux image using skopeo and umoci. This is a precursor to
+# proper support within the tools (issue #325).
+#
+# Note that this approach is a little contrived, in that we unpack the image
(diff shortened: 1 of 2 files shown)


test/Build.skopeo-umoci
```diff
@@ -0,0 +1,37 @@
...
+
+srcdir=$1
+tarball_gz=${2}.tar.gz
+workdir=$3
+
+cd "$workdir"
+
+if ( ! command -v skopeo >/dev/null 2>&1 ); then
+    echo 'skopeo not found' 1>&2
+    exit 65
```
j-ogas: Minor nitpick: would a return value of 69 (service unavailable) be more appropriate?

PR #336, opened 2019-01-17
update install permissions (closes #328)

### LINKED ISSUE #328: make install PREFIX gives unintended executable permissions to some test suite files
After a `make install PREFIX`, some files are given executable permission bits. 

```bash
$ ls -lh /usr/local/share/doc/charliecloud/test/sotest/
total 28K
drwxr-xr-x. 2 root   20 Jan  8 17:50 bin/
-rwxr-xr-x. 1 root   46 Oct 31 18:34 files_inferrable.txt*
drwxr-xr-x. 2 root   30 Jan  8 17:50 lib/
lrwxrwxrwx. 1 root   18 Jan  8 17:50 libsotest.so -> ./libsotest.so.1.0*
lrwxrwxrwx. 1 root   18 Jan  8 17:50 libsotest.so.1 -> ./libsotest.so.1.0*
-rwxr-xr-x. 1 root 7.7K Dec 20 12:15 libsotest.so.1.0*
-rwxr-xr-x. 1 root 8.4K Dec 20 12:15 sotest*
-rwxr-xr-x. 1 root  179 Oct 31 18:34 sotest.c*
```

Note the permissions for `files_inferrable.txt` and `sotest.c`.

j-ogas:
Addresses #328.

Note: not sure what the deal with the white space is (I use spaces, not tabs); everything is aligned perfectly on my local Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,7 +134,8 @@ install: all
 	done
 	chmod 755 $(DOC)/examples/serial/hello/hello.sh \
 	          $(DOC)/examples/syscalls/pivot_root \
-	          $(DOC)/examples/syscalls/userns
+	          $(DOC)/examples/syscalls/userns \
+			  $(DOC)/examples/*/*/*.sh
 	find $(DOC)/examples -name Build -exec chmod 755 {} \;
 #       tests
 	install -d $(TEST) $(TEST)/run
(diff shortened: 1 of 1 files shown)


reidpr:
So one obnoxious quirk here is that Makefiles *must* prefix each recipe line with at least one tab. I suspect that's the source of the alignment issues; the existing code will have one tab followed by as many spaces as needed. For example line 138 in the new `Makefile` is indented with three tabs and then two spaces but 137 is (correctly) one tab and then 10 spaces.

You could use `hexdump` to see exactly what is going on.

I think if you set your editor tabs to 8 spaces it should display the same as in GitHub. Also some editors will highlight Makefiles to make this issue more obvious.

j-ogas:
Thanks for the tip! I believe it is now aligned properly; adding ready-to-merge tag.

Issue #295, opened 2018-11-29
add an option to ch-run to suppress forwarding of SLURM env variables to processes in the container


hppritcha:
Hunter and I are going around in circles on some edge cases with Open MPI and doing 
- singleton launch within a container (that means running a MPI executable without using mpirun, etc.)
- mpirun launch within a container which was in turn started within a SLURM allocation

These problems would vanish if there were a ```ch-run```  command line option to suppress forwarding of SLURM environment variables to the processes (including mpirun) within the container.

hppritcha:
assigning to Jordan

j-ogas:
Related to #224. The proposed solution to 224 would provide a script, `docker-env-only`, in the container root that would clean the container environment variables and add those declared in the Dockerfile via the `ENV` directive. SLURM variables would be unset in that case.

@hppritcha Would that suffice? Or is there a use case in which you may want _only_ the SLURM environment variables unset?

reidpr:
Maybe #224 should be "add Docker environment variables to existing environment" and this issue should be "remove variables from environment".

heasterday:
Something to bear in mind for the future, we likely will also want to drop ALPS env variables on the Crays to prevent the same behavior that breaks guest launches when Slurm is detected. In a Cray environment first Slurm env variables are detected generating in guest launches the error "The application appears to have been direct launched using "srun" but upon dropping them then a guest launch results in the following error "The application appears to have been direct launched using "aprun". It is only after dropping both sets is the launch successful.

Issue #226, opened 2018-08-22
upgrade examples to OpenMPI 3.1


reidpr:

hppritcha:
see for example the hello world example in
https://github.com/hpc/charliecloud/tree/master/examples/mpi/mpihello

hppritcha:
@heasterday and @hppritcha worked on this today and ran into severe turbulence owing to changes in the way Open MPI starting with the 3.0.x series checks whether or not to initialize its PMIx framework.  Starting with this series (see commit open-mpi/ompi@011403c04a7368f7266e6b6da27e41f28b49f502) there's some "helpful" checking within the binary whether or not the user is trying to launch the process(es) using mpirun (aka orte) or some other launcher (srun or alps).  This "helpful" checking requires the --with-slurm=no NOT be set in order for it to be able to detect that the job is being used with SLURM and so can try and use SLURMs PMI2 or PMI1.  This is irrespective of whether or not --with-pmi was requested.  

Now there are several ways around this 
- open a PR against Open MPI to undo some of the effects of the above mentioned commit.  However I very much doubt the author of that commit will accept such changes as he'd say why are you asking for SLURM pmi support but then disabling use of SLURM itself.
- maintain a patch (small and only to configury and not source code itself) which would be applied as part of building Open MPI for the container
- use an addition mca parameter on the ```mpirun``` command line when running mpirun within the container (the use case that hit turbulence without using the --with-slurm=no configure option).

Hunter is testing a patched Open MPI with the configury change and also the additional mca parameter option.

reidpr:
Thanks for the detective work. Couple ideas I had:

1. Remove Slurm stuff at a finer grain via `configure --enable-mca-no-build`.
2. Document Slurm environment variable to unset if you are doing guest launch.

hppritcha:
hmm... okay for 1. then we'd remove ```--with-slurm=no``` and add in ```--enable-mca-no-build=plm-slurm``` to disable using slurmd based launch of the orted(s) when doing guest based launch.  If this works,  there should be no need to do 2 which is gross.

reidpr:
Closed by PR #272, commit 3c371c8. Forgot to close in commit message.

reidpr:
Re-opening; all or most of the MPI tests fail on OPA clusters, and perhaps other multi-node environments. Errors include:

```
$ srun -n2 ch-run /var/tmp/mpihello -- /hello/hello
[...]

PR #337, opened 2019-01-22
Openmpi 3.1.3 build with UCX don't use HFI patch

heasterday:
This PR should resolve issues with host launches and guest launches within the test suite while skipping serial launches of mpi applications pending issue 295. Guest launches require the command line option "--host localhost" to prevent Openmpi attempting to startup on remote hosts within a multinode allocation. 
Dockerfile.openmpi likely needs some clean up. The current way I'm lugging around the patch file and applying it may need to be changed as well.

### DIFF
--- a/examples/mpi/mpibench/test.bats
+++ b/examples/mpi/mpibench/test.bats
@@ -38,6 +38,7 @@ check_process_ct () {
 
 # one from "Single Transfer Benchmarks"
 @test "${ch_tag}/pingpong (guest launch)" {
+    unslurmify_environment_maybe
     [[ $ch_cray && $ch_mpi = mpich ]] && skip "issue #255"
     # shellcheck disable=SC2086
     run ch-run "$ch_img" -- mpirun $ch_mpirun_np "$imb_mpi1" $imb_args PingPong
@@ -61,6 +62,7 @@ check_process_ct () {
 
 # one from "Parallel Transfer Benchmarks"
 @test "${ch_tag}/sendrecv (guest launch)" {
+    unslurmify_environment_maybe
     [[ $ch_cray && $ch_mpi = mpich ]] && skip "issue #255"
     # shellcheck disable=SC2086
     run ch-run "$ch_img" -- mpirun $ch_mpirun_np "$imb_mpi1" $imb_args Sendrecv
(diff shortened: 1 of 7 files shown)


j-ogas:
Relevant to #226 and #295.

j-ogas:
Preliminary testing was successful. Note that @heasterday instructed me to `export OMPI_MCA_btl="^vader"`. He explained to me why we needed to do this but I can't seem to remember why. In any case, I thought it would be a good idea to document this step.

heasterday:
Within our test suite at least one of the images has a version of gcc that gets bitten by the following issue:
https://github.com/open-mpi/ompi/issues/6014
Where a certain atomic operation gets treated as a no-op, this results in the vader btl hanging. For the sake of testing Howard's UCX patch I used the parameter Jordan showed to remove vader from use. This was simply a workaround to test UCX's behavior without that issue potentially being in the mix.

Until Openmpi comes out with a new release, building images from openmpi's nightly tarball for 3.1.3 should resolve this.

reidpr:
Hunter, re. "nightly builds not being stable", who is concerned about stability? The OpenMPI team?

Using a nightly build pending 3.1.4 is appealing to me if the OpenMPI people are comfortable with it. If OpenMPI 3.1.4 comes out on April 1, that will mean either Charliecloud 0.9.9 (April 1) or 0.9.10 (May 6).

I don't want to disable `vader` or require users to set an environment variable.

reidpr:
We can also consider installing a newer GCC in the relevant images. I looked briefly at those OpenMPI bugs and it seemed that was not too hard.

reidpr:
OK, let's proceed with a PR that includes the nightly tarball. Is there a way to get a specific nightly tarball, to retain reproducible builds on our end?

heasterday:
We could grab the January 11th 3.1.x tarball and test with that as I believe it has the vader fix, the hugepages fix is still pending on that branch though but that is a separate issue.

reidpr:
@heasterday: Can you look over my changes? I have tests passing on both a cluster and my workstation. If you're happy, I'll merge.

heasterday:
@reidpr Looks good to me

PR #338, opened 2019-01-22
add Fedora/EPEL compliant spec file

j-ogas:
The packages produced by this spec file (`charliecloud-${version}` and `charliecloud-doc-${version}`) are compliant with the Fedora/EPEL guidelines and the Filesystem Hierarchy Standard (FHS). 

Note that the `charliecloud-doc` package generates a couple warnings and one (false) error with`rpmlint`: 

```
charliecloud-doc.x86_64: E: no-ldconfig-symlink /usr/lib64/charliecloud-0.9.6/test/sotest/lib/libsotest.so.1.0
charliecloud-doc.x86_64: W: dangerous-command-in-%post ln
charliecloud-doc.x86_64: W: dangerous-command-in-%postun rm
1 packages and 0 specfiles checked; 1 errors, 2 warnings.
```

It is fairly common for test files to produce errors. In our case, rpmlint detects `libsotest.so.1.0` and complains that we did not provide a link via `/sbin/ldconfig`. This is understandable, however, charliecloud is a container runtime; we inject the `libsotest.so` into a guest environment via `ch-fromhost` and execute `/sbin/ldconfig` inside the guest.


packaging/el7/charliecloud.spec
```diff
@@ -0,0 +1,91 @@
...
+%prep
+%setup -q
+
+%build
+%{__make} %{?mflags}
+
+%install
+%{__make} %{?mflags_install} install PREFIX=%{_prefix} DESTDIR=%{buildroot}
+
+%postun
```
mej: This will blow up the very first time you do an upgrade on a charliecloud RPM.  The `%postun` scriptlets don't only run at uninstall time; they also run at **upgrade** time.  You need to check the value of `$1` to see which is which.  See https://fedoraproject.org/wiki/Packaging:Scriptlets#Syntax for a breakdown.

packaging/el7/charliecloud.spec
```diff
@@ -0,0 +1,91 @@
...
+rm -rf %{_libexecdir}/%{name}
+
+%postun doc
+rm -rf %{_libexecdir}/%{name}/examples
+rm -rf %{_libexecdir}/%{name}/test
+
+%files
+# Documentation
```
mej: As a general rule, documentation files should be handled via `%doc` so that `--excludedocs` functions correctly.

packaging/el7/charliecloud.spec
reidpr: > upgrade on a charliecloud RPM

Any way we can test an upgrade with just one RPM (i.e., no lower-versioned package already on the system)?

mej:
Those scriptlet warnings may be a result of the lack of `$1` checking in your scriptlets, which is, indeed, extremely dangerous.

packaging/el7/charliecloud.spec
mej: No; you have to install a lower-versioned RPM in order to have it upgrade to a higher-versioned one.  The `%postun` scriptlet can't get run if there's no older RPM being removed that has a `%postun` scriptlet. :-)

j-ogas:
@mej Thank you for the feedback! I will update the spec file accordingly.

Issue #324, opened 2018-12-18
ch-fromhost: better error message if no path returned by ldconfig


reidpr:
(Reported by @samcmill in PR #321.)

Actual behavior:

```
$ cd $CH_TEST_IMGDIR
$ cp -a debian9 foo
$ rm foo/sbin/ldconfig
$ ch-fromhost --lib-path ./foo

$ echo $?
0
```

Expected behavior:

```
$ ch-fromhost --lib-path ./foo
ch-fromhost: empty path from ldconfig
$ echo $?
1
```

Note that in this case, `ldconfig` failed (because it didn't exist), but we lose that because the rest of the pipeline at `ch-fromhost:236` proceeds.

We might want to distinguish between `ldconfig` failing and returning zero paths.

Currently we pipe stderr of `ldconfig` to `/dev/null` because of excessive chatter, but this also hides error messages.

PR #342, opened 2019-01-25
ch-fromhost: proper error if no path from ldconfig

reidpr:
Addresses #324.

This involves some pretty bonkers shell code, so @mej I'd love your opinion on whether we've gone too far.

@samcmill, any opinions on this solution for your bug? Thanks again for reporting it.

### DIFF
--- a/bin/ch-fromhost
+++ b/bin/ch-fromhost
@@ -233,8 +233,15 @@ if [ $lib_found ]; then
     # searches, so that we can override (or overwrite) any of the same library
     # that may already be in the image.
     debug "asking ldconfig for shared library destination"
-    lib_dest=$(  "${ch_bin}/ch-run" "$image" -- /sbin/ldconfig -Nv 2> /dev/null \
-               | grep -E '^/' | cut -d: -f1 | head -1)
+    # "ldconfig -Nv" gives some pointless warnings on stderr even if
+    # successful; we don't want to show those to users. However, we don't want
+    # to simply pipe stderr to /dev/null because this hides real errors. Thus,
+    # use the following abomination to pipe stdout and stderr to *separate
+    # grep commands*. See: https://stackoverflow.com/a/31151808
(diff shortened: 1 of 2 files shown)


bin/ch-fromhost
```diff
@@ -233,8 +233,15 @@ if [ $lib_found ]; then
     # searches, so that we can override (or overwrite) any of the same library
     # that may already be in the image.
     debug "asking ldconfig for shared library destination"
-    lib_dest=$(  "${ch_bin}/ch-run" "$image" -- /sbin/ldconfig -Nv 2> /dev/null \
-               | grep -E '^/' | cut -d: -f1 | head -1)
+    # "ldconfig -Nv" gives some pointless warnings on stderr even if
+    # successful; we don't want to show those to users. However, we don't want
+    # to simply pipe stderr to /dev/null because this hides real errors. Thus,
+    # use the following abomination to pipe stdout and stderr to *separate
```
j-ogas: (haha!)

test/run/ch-fromhost.bats
```diff
@@ -261,6 +261,19 @@ fromhost_ls () {
     [[ $status -eq 1 ]]
     [[ $output = *'duplicate image'* ]]
     fromhost_clean_p "$img"
+
+    # ldconfig gives no shared library path (#324)
+    #
+    # (I don't think this is the best way to get ldconfig to fail, but I
+    # couldn't come up with anything better. E.g., bad ld.so.conf or broken
+    # .so's seem to produce only warnings.)
```
samcmill: Another mode would be to break ch-run (that's the case I ran into).

With this PR, I get a useful error message that would give me a good start on figuring out what was wrong.

```
$ ch-fromhost --nvidia ex2/rootfs
ch-run[73]: error: Permission denied (charliecloud.c:508 13)
ch-fromhost: empty path from ldconfig
```

samcmill:
Looks good to me.

PR #351, opened 2019-02-06
test suite: better error if test/bin symlink does not contain ch-run

j-ogas:
The current error message implies that it will only be seen if a user attempts to execute the test suite before building charliecloud. However, this error can also be seen when charliecloud is built but the symlink `charliecloud/test/bin` is broken or points to the wrong directory. In that case, the error message is not helpful.

### DIFF
--- a/test/common.bash
+++ b/test/common.bash
@@ -132,7 +132,7 @@ ch_runfile=$(command -v ch-run)
 # shellcheck disable=SC2034
 ch_libexec=$(ch-build --libexec-path)
 if [[ ! -x ${ch_bin}/ch-run ]]; then
-    printf 'ch-run not found in ${ch_bin} (build with "make"?).\n\n' >&2
+    printf "ch-run not found in %s (has charliecloud been built with \"make\"?).\n\n" "${ch_bin}" >&2
     exit 1
 fi
 
--- a/test/common.bash
+++ b/test/common.bash
@@ -132,7 +132,7 @@ ch_runfile=$(command -v ch-run)
 # shellcheck disable=SC2034
 ch_libexec=$(ch-build --libexec-path)
 if [[ ! -x ${ch_bin}/ch-run ]]; then
-    printf 'Must build with "make" before running tests.\n\n' >&2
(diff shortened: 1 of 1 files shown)


j-ogas:
Ready to be merged, deferred, or deleted (haha).

test/common.bash
```diff
@@ -132,7 +132,7 @@ ch_runfile=$(command -v ch-run)
 # shellcheck disable=SC2034
 ch_libexec=$(ch-build --libexec-path)
 if [[ ! -x ${ch_bin}/ch-run ]]; then
-    printf 'Must build with "make" before running tests.\n\n' >&2
+    printf "ch-run not found in %s (has charliecloud been built with \"make\"?).\n\n" "${ch_bin}" >&2
```
reidpr: Let's do:

```
printf 'ch-run: not in %s; did you build with "make"?\n\n' "$ch_bin" >&2
```

j-ogas:
> but not with a broken symlink:
> 
> ```
> $ rm test/bin
> $ ln -s /doesnotexist test/bin
> $ (cd test && CH_TEST_SCOPE=standard make test)
> bats build.bats build_auto.bats build_post.bats
> /home/reidpr/charliecloud/test/common.bash: line 127: cd: /home/reidpr/charliecloud/test/bin: No such file or directory
> /home/reidpr/charliecloud/test/bats/libexec/bats-exec-suite: line 20: let: count+=: syntax error: operand expected (error token is "+=")
> Makefile:30: recipe for target 'test-build' failed
> make: *** [test-build] Error 1
> ```

I believe this would be addressed with the `readlink(1)` wrapper from #353. It would error as follows (in the current state):

```
ch_bin=/doesnotexist: path does not exist
```

>     1. How does the symlink pointing to an incorrect directory happen? 

Hard to imagine a situation other than when a developer is messing with the master Makefile. In that case `ch_bin=/doesnotexist: path does not exist` (or a similar flavor) appears  to be sufficient.

reidpr:
Closing per request from @j-ogas.

PR #352, opened 2019-02-07
modify test and example install dest; add export-bats

j-ogas:
Supports #338.

We differentiate between `export` and `export-bats`, which produce charliecloud tarballs with and without bats respectively. 

We also change the target installation path for our tests and examples to be `${prefix}/libexecdir/`.

Note that this will break `packaging/debian` and `packaging/redhat`. The former will need to be updated, the latter will be replaced with #338.


j-ogas:
@reidpr Ready to merge. Note that a new issue, 'update debian package' will need to be created as this PR disables it.

reidpr:
Thanks Jordan. What's the reasoning to do this in a separate PR rather than part of #338?

j-ogas:
@reidpr In case we wanted to push #338 to 0.9.8. In which case we'd have the necessary assets (tarball with documentation and no bats) already in place.

reidpr:
OK. I'd prefer to roll this into #338. I'm open to argument, but otherwise go ahead and close this PR.

PR #372, opened 2019-03-01
Reboot after setting up passwordless logins

### LINKED ISSUE #369: docs: mention that Vagrant VM must be restarted for passwordless access to work

rstyd:
Resolves #369 by removing the need for the user to restart the system.

### DIFF
--- a/packaging/vagrant/Vagrantfile
+++ b/packaging/vagrant/Vagrantfile
@@ -235,6 +235,8 @@ EOF2
     passwd -l root
     passwd -l vagrant
 
+    # Reboot the machine so the paswordless login takes affect
+    sudo reboot
   EOF
 
   # Test Charliecloud (optional).


packaging/vagrant/Vagrantfile
```diff
@@ -235,6 +235,8 @@ EOF2
     passwd -l root
     passwd -l vagrant
 
+    # Reboot the machine so the paswordless login takes affect
+    sudo reboot
```
j-ogas: Would `c.vm.provision :reload` work here?

rstyd: I was initially using c.vm.provision :reload, but that can't be called inside of the provision since everything in that context has to be bash. Additionally, since the "ova" provisioner is only called conditionally, it wouldn't run after it either. I spent a long time looking for a way to conditionally perform :reload based upon the name of a provision, but there doesn't seem to be one. Therefore, this seems like a simple solution, and it has the intended effect.

reidpr:
The instructions already have a shutdown and restart in §5.3.3, which is right after `--provision-with=ova`. This is so we can make the distributed export from a shut-down state.

Rebooting in the provisioner thus seems unneeded. Can you clarify? Am I missing something?

rstyd:
Closing since this should be handled by altering the documentation instead.

PR #377, opened 2019-03-06
ch-tar2dir: better error message

### LINKED ISSUE #340: ch-tar2dir: clarify error message if DIR doesn't exist
@samcmill reports in PR #335 review that `ch-tar2dir` gives a confusing error message if `DIR` doesn't exist. One can tickle this usability bug if one misinterprets `DIR` to be the image directory, not its parent:

```
$ ch-tar2dir alpine.tar.gz ch-img
creating new image ch-img/alpine
mkdir: cannot create directory 'ch-img/alpine': No such file or directory
$ mkdir ch-img
$ ch-tar2dir alpine.tar.gz ch-img
creating new image ch-img/alpine
ch-img/alpine unpacked ok
$ ls ch-img/alpine/
WEIRD_AL_YANKOVIC  dev  home  media  proc  run   srv  tmp  var
bin                etc  lib   mnt    root  sbin  sys  usr
```

reidpr:
Closes #340.

### DIFF
--- a/bin/ch-tar2dir
+++ b/bin/ch-tar2dir
@@ -27,6 +27,16 @@ if [ $# -lt 2 ]; then
     usage
 fi
 
+# Complain clearly if destination not valid (issue #340).
+if [ ! -e "${2}" ]; then
+    echo "can't unpack: ${2} does not exist" 1>&2
+    exit 1
+fi
+if [ ! -d "${2}" ]; then
+    echo "can't unpack: ${2} is not a directory" 1>&2
+    exit 1
+fi
+
 # Figure out the real tarball name. If the provided $1 already has a tar
 # extension, just test that name; if not, also append the plausible extensions
 # and try those too.
--- a/test/run/ch-tar2dir.bats
+++ b/test/run/ch-tar2dir.bats
@@ -34,6 +34,18 @@ load ../common
 @test 'ch-tar2dir: errors' {
     scope quick
 
+    # destination doesn't exist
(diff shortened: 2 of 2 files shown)


reidpr:
@samcmill, you reported the usability issue; do you think this is a reasonable fix?

reidpr:
Thanks @samcmill.

I find it counter-intuitive at times too. The reasoning for having it this way is as follows, though whether it's *good* reasoning is open to judgement. It's for consistency; all the scripts have an argument of the form "put the output artifact in this directory", with the name itself computed for less typing. It might be a little more intuitive for `ch-docker2tar`:

```
$ ch-docker2tar alpine36 /var/tmp                   # current behavior
```

vs.

```
$ ch-docker2tar alpine36 /var/tmp/alpine36.tar.gz  # alternate: more expressive but longer to type
```

We did actually have it the way you suggest (argument is image path, not its parent) for a while, but people tended to make the opposite error, assuming that it was the parent like the other scripts.

samcmill:
My perspective is that `ch-run` is the "core" command.  In that light, consistency would primarily mean being consistent with `ch-run` which takes the image path, not the parent path, as it's argument.

(This is just provide some more color on my perspective and why I personally find it counter intuitive.  I'm not suggesting you re-open this topic or make any changes.  The new error message will immediately correct me if I misinterpret this in the future.)

reidpr:
Thanks! I hadn't thought of that way. That's a valuable perspective.

reidpr:
I agree on `fatal`. Rusty suggested it too I think.

Issue #248, opened 2018-09-21
test suite: make it work on ARM (single node)


reidpr:
@PaulUTK correctly points out in #245 that we have “a lot of hardcoded x86_64 paths and the ARMs are coming”.

PR #381, opened 2019-03-12
remove un-advertised dependency & propose clarified policy

### LINKED ISSUE #376: clarify and tighten up the prerequisites
There are a couple of problems here. First, looking over the documentation:

* Run time says POSIX shell only — but `Makefile` sets `SHELL=/bin/bash`.
* Build time states a dependency on "Bash 4.1+", but I can't figure out what that would be.
* We ought to state briefly what each dependency is for.

Second, some things are really difficult in POSIX sh but easier in Bash and much easier in Python. For example:

* I'm trying to improve `ch-build2dir` UI (#370), and sh makes it difficult or impossible (not sure which) to strip the last argument and pass the rest to a subcommand; this is a convenience script, so not having it wouldn't mean you could do fewer things. 
* On the other hand, an OCI layer for `ch-run` (#364) seems so much easier to implement in Python than sh/Bash or C that even though it really is additional functionality, maybe we want to require Python for that. (RHEL7 has Python 2.7 but not 3.)

This issue is to work out a policy and write it into the documentation.

reidpr:
Closes #376.


doc-src/install.rst
```diff
@@ -56,32 +56,35 @@ example:
...
+* Shell scripts with various manual bootstrap and :code:`ch-run`
+* Docker
+* `Buildah <https://github.com/containers/buildah>`_
+* `skopeo <https://github.com/containers/skopeo>`_ and
+  `umoci <https://github.com/openSUSE/umoci>`_
 
-Note that without Docker on the build system, some of the test suite will be
-skipped.
+Most of the test images are built with Docker, so the test suite is less
+thorough if it's not installed.
```
j-ogas: Is architecture worth mentioning, at least while we have some hard coding (#248)?

reidpr: Yes, that's a good point.

reidpr:
TODO:
- `make` requires Python because it builds in `test/`
- document `sudo` use in test suite for fixture building

reidpr:
Consider: How much do we really need POSIX sh compatibility? What are the cost/benefit of requiring Bash?

reidpr:
Also addresses #389.

Makefile
```diff
@@ -1,12 +1,15 @@
-SHELL=/bin/bash
+SHELL=/bin/sh
```
j-ogas: I'm not sure we should cling to POSIX sh instead of adopting bash.

doc-src/dev.rst
```diff
@@ -431,8 +435,8 @@ Item 2 is tested; i.e., if you break the RPM build, the test suite will fail.
 This section describes how to build the RPMs and the pain we've hopefully
 abstracted away.
 
-Prerequisites
--------------
+Dependencies
+------------
 
   * Python 2.7
```
j-ogas: It would be nice if we had a homogeneous python version requirement. Could we list python3 as a standard requirement with the note "should work with python2"?

doc-src/dev.rst
```diff
@@ -555,9 +559,51 @@ Writing English
...
+All dependencies must be stated and justified in the documentation.
+
+We want Charliecloud to run on as many systems as practical, so we work hard
+to keep dependencies minimal. However, because Charliecloud depends on new-ish
+kernel features, we do depend on standards of similar vintage.
+
+Core functionality should be available even on small systems with basic Linux
+distributions, so dependencies for run-time and build-time are only the bare
+essentials. Exceptions, to be used judiciously:
+
```
j-ogas: We position ourselves to appeal to HPC environments, should our dependencies reflect the lowest common denominator (e.g., posix shell)? Have we looked at our competitor's dependencies?

Makefile
reidpr: I agree, but I wanted to defer this decision; hence #396.

doc-src/dev.rst
reidpr: Not really but I assumed they're way bigger.

doc-src/dev.rst
reidpr: I agree.

PR #385, opened 2019-03-13
RPM build improvements

### LINKED ISSUE #384: RPMs accepted into Fedora
Review request: https://bugzilla.redhat.com/show_bug.cgi?id=1690046

j-ogas:
Addresses #384.


reidpr:
Looks good. I added one more commit.

On your approval, do you want me to merge, or should we wait for feedback from Fedora?

j-ogas:
Another thought: I think Build should—for release version—download the tarball via the `source0` URL instead of doing a git checkout and `make export`. This makes it such that the source tarball in the generated SRPM will have a hash that matches the source0 URL tarball, which is required by Fedora.

reidpr:
Downloading `Source0` is fine with me. I wonder if `build` should take an option to say whether to download or `make export`.

j-ogas:
I have pushed two commits. The first updates the source URL and the second adds `python 2.7` as a temporary build requirement for charliecloud (ch-run). 

I propose that we open a new issue to modify the build procedure to skip building the test suite if python is not detected (0.9.9). Once 0.9.9 is released we can modify the spec file and move the `python 2.7` dependency to the `charliecloud-doc` subpackage and add a release note for 0.9.10.

reidpr:
This is the Bugzilla bug, with some feedback: https://bugzilla.redhat.com/show_bug.cgi?id=1690046

j-ogas:
TODO: add README to `charliecloud/test`.

j-ogas:
Second round of feedback:

- Conditionals for SUSE are not allowed,
- Remove `Buildrequires: Make` (redundant),
- Version constraints are unnecessary since they will always be satisfied  
- LICENSE shouldn't be duplicated in %doc (redundant),
- Documentation should be in `%_pkgdocdir` not `%{_datadir}/doc`
- Conditional for `README.EL7` is incorrect,
- Need to find sponsor.

j-ogas:
Fedora requires that spec files either install documentation to `%{prefix}/doc/%{name}-%{version}` (`%_pkgdocdir`), or use `%doc`, which automatically copies the documentation files into the correct directory, e.g., `%doc README.rst` will copy the readme file to `${prefix}/charliecloud-0.9.8/README.rst`.

c8d94ad addresses the issue by using `%doc` in conjunction with `%exclude ${prefix}/doc/charliecloud` to ensure that the documentation files are not duplicated. We are essentially moving the readme installed via `make install PREFIX` at package build time.

We should consider modifying our top level Makefile to reflect the `${prefix}/doc/charliecloud-${version}` requirement as part of 0.9.9. This would allow use to use the `%_pkgdocdir` macro exclusively (no need to exclude files or worry about duplicates).

j-ogas:
Removing `ready to merge`. We have `/usr/bin/python3` listed as a charliecloud build dependency, it should only be a test-package dependency. I need to rebuild the package and submit to Fedora.

j-ogas:
> Removing `ready to merge`. We have `/usr/bin/python3` listed as a charliecloud build dependency, it should only be a test-package dependency. I need to rebuild the package and submit to Fedora.

Right, I had forgotten we have python listed as a build dependency because our top level Makefile will also make the contents of charliecloud/test, which fails on systems without python installed.

reidpr:
> Right, I had forgotten we have python listed as a build dependency because our top level Makefile will also make the contents of charliecloud/test, which fails on systems without python installed.

We did change it so it no longer fails to build if Python isn't installed — it just doesn't build/run the tests — but a build dependency is still needed IMO so we can build the test package.

PR #388, opened 2019-03-18
update tests to use alpine 3.9

### LINKED ISSUE #386: upgrade Alpine in examples to 3.9
Looks like we have both 3.6 and 3.8 floating around the test suite.

j-ogas:
Addresses #386.


reidpr:
Thanks Jordan. I'd like to bring this in after we merge the ARM changes (issue #248), which should not take too long.

reidpr:
Looks like the test failures are real now.

Issue #180, opened 2018-06-12
make OpenMPI examples work on Crays


reidpr:

reidpr:
The current status is: The existing examples work, but only using the `tcp` BTL, which gives prohibitively bad performance.

The problem is that we need to build OpenMPI with Cray's uGNI libraries, and I currently don't know how to get those anywhere other than from a Cray system or from Cray people. Because this example is a reference implementation, we can only use things from the public internet.

I am checking in with a few people about availability of these libraries on the public internet.

Another option is to add the uGNI modules at runtime (e.g. by augmenting `ch-fromhost`), though this would not be the preferred approach.

reidpr:
Today we discussed how to deal with guest launch on the Crays. It seemed that the choice was go down a rabbit hole of MPI mess or just don't crayify for guest launch, i.e., put the guest launch tests first, then crayify, then host launch. We selected the latter. We also need to ensure that on the *next* run of the test suite the guest launch doesn't get a crayified image, which is a little tricky because `make test-run` only unpacks the images on the first node in a Slurm allocation.

reidpr:
I think the way to do this may be:

1. Rename `crayify_mpi_maybe` to `crayify_mpi_or_skip` and make it skip instead of doing nothing.
2. Reorder the relevant `test.bats` files to do guest launch first, then host launch.
3. Before host launch, add a new test that calls `crayify_mpi_or_skip`.
4. After the host launch sequence, add a new test that re-unpacks the relevant image on all nodes.

PR #402, opened 2019-03-29
Expand ch-fromhost --cray-mpi functionality for OpenMPI images

heasterday:
Addresses the host launch case of #180 

This PR expands the functionality of ch-fromhost to also "crayify" images containing OpenMPI allowing the usage of Cray's high speed interconnect.


bin/ch-fromhost
```diff
@@ -306,21 +304,55 @@ if [ $cray_mpich ]; then
...
+    queue_files "$host_libmpi"
+    queue_files "$(  ldd "$host_libmpi" \
+                   | grep -E "/usr|/opt" \
+                   | sed -E 's/^.+ => (.+) \(0x.+\)$/\1/')" 
+
+    # This is appears to be the only .so required from /lib64
+    queue_files /lib64/liblustreapi.so
+
+    # FIXME: This is done to remove libmpi from the image before injecting
+    # from the host to workaround paraview linking to the original libmpi
```
reidpr: this also solves the UCX problem

reidpr:
@heasterday, if you're happy with my changes, this is good to merge I think, with one caveat:

`make test-test` tended to hang, at random times, perhaps preferring to hang in the serial tests we were skipping anyway? I did get it to pass once. I attributed these hangs to "it's a Cray" but maybe you have a better idea.

heasterday:
@reidpr changes look good. As for the hangs ran a couple full tests and I'm not seeing any hangs currently so hopefully whatever you were seeing was transient. I also tested whether or not this work addresses issue 411 and it does so I removed the workarounds, feel free to revert if you would like more testing in this area.

Issue #401, opened 2019-03-26
read-only mount fails on local disk


rodwalker:
Hi,
I see the error
ch-run[38718]: can't re-mount image read-only (is it on NFS?): Operation not permitted (charliecloud.c:230 1)
but the image is on a local disk partition
/dev/sda7 on /tmp type ext3 (rw,nosuid,relatime,data=ordered)

I can live with the read-write option for now, but would be interested to know what feature of /tmp makes it look like NFS.

Cheers,
Rod.

j-ogas:
Hello @rodwalker, that is a generic error message that occurs when the mount syscall,  
`mount(NULL, c->newroot, NULL, MS_REMOUNT|MS_BIND|MS_RDONLY, NULL)`, fails. This typically occurs when attempting to use ch-run on a NFS without the read-write option. Thus, we added the NFS hint. 

Related: #9, #7.

In this case, the remount as read-only on your disk partitioned `/tmp` is failing. I am in the process of building a vm with a manually partioned (non-tmpfs) tmp in an effort to reproduce.

j-ogas:
I am unable to reproduce the error in a simple centos 7.6 virtualbox environment. 

```
$ df -Th | grep ^/dev/sda2
dev/sda2               ext3      1.9G  354M  1.4G  21% /tmp
$ ./bin/ch-run /tmp/centos7/ -- /bin/true
$ echo $?
0
```

There could be an issue with your /tmp partition.

rodwalker:
Hi,
The machine is SLES12 SP3, so it would not surprise me if there is something odd. /var/tmp is ok!

rod/charliecloud/bin/ch-run  /tmp/centos7_atlas.0  pwd
ch-run[29392]: can't re-mount image read-only (is it on NFS?): Operation not permitted (charliecloud.c:230 1)

rod/charliecloud/bin/ch-run  /var/tmp/centos7_atlas.0  pwd
/
The respective partitions are
/dev/sda7 on /tmp type ext3 (rw,nosuid,relatime,data=ordered)
/dev/sda5 on /var type ext3 (rw,relatime,data=ordered)

so only difference nosuid. Unless the size is relevant for some reason
/dev/sda5       4,8G  885M  3,7G  19% /var
/dev/sda7       227G   69G  147G  32% /tmp

Cheers,
Rod.

j-ogas:
Aha! 

That does seem to be the issue. I added `nosuid` to my /tmp partition and remounted it:

```
[...]

PR #407, opened 2019-04-10
re-mount read-only using existing flags

reidpr:
Use `statvfs(3)` to pass existing mount flags (except the ones we want to change) to `mount(2)` when we remount the new root filesystem read-only. This addresses #9 and #401.

Passes Travis and full-scope on my box. These tests do not, however, exercise the specific cases identified in #401. I'd like to try it on that before merging.

### DIFF
--- a/bin/charliecloud.c
+++ b/bin/charliecloud.c
@@ -17,6 +17,7 @@
 #include <sys/mount.h>
 #include <sys/prctl.h>
 #include <sys/stat.h>
+#include <sys/statvfs.h>
 #include <sys/syscall.h>
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


j-ogas:
Ran successfully on front-end and computes in both cray and non-cray production environments.

reidpr:
Does it work with an image on our NFS filesystems now too?

reidpr:
@rodwalker, could you give this PR a try in your environment and see if it fixes the bug you reported in #401?

j-ogas:
> Does it work with an image on our NFS filesystems now too?

Yes!

Issue #412, opened 2019-04-12
HTML docs: include in "make install" and "make export"


reidpr:
Among other things, this supports RPM packaging efforts.

Extends #302.

PR #417, opened 2019-04-17
include html files in `make install` and `make export`

j-ogas:
Addresses #412

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ bin/version.sh: VERSION.full
 # the man pages but don't know how to build them.
 #
 # They are phony because I haven't figured out their real dependencies.
-.PHONY: main.tar
+.PHONY: main.tar doc/index.html
 main.tar: VERSION.full man/charliecloud.1
 	git diff-index --quiet HEAD || [ -n "$$CH_MAKE_EXPORT_UNCLEAN_OK" ]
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -135,9 +135,20 @@ install: all
 	    install -d $(INSTALL_PREFIX)/share/man/man1; \
 	    install -pm 644 -t $(INSTALL_PREFIX)/share/man/man1 man/*.1; \
 	fi
-#       misc "documentation"
+#       license and readme
 	install -d $(DOC)
 	install -pm 644 -t $(DOC) LICENSE README.rst
+#	html files if they were built
+	if (ls -A doc | grep -q .); then \
```
reidpr: [Don't parse `ls`](http://mywiki.wooledge.org/ParsingLs). You want to see if there are any HTML files in there, right? I think something [based on `find(1)`](https://stackoverflow.com/a/15663760) would be better.

reidpr: Actually, we know the filenames, so we don't need `find(1)`. I think something like `if [ -f doc/index.html ]` would do the trick.

reidpr:
Added a commit to exclude `doc/.git` and `doc/.jekyll` files. (The former shows up when it's a checkout of the `gh-pages` branch for publishing to GitHub Pages.) Does this break anything?

Issue #346, opened 2019-01-28
upgrade OpenMPI examples to 3.1.4


reidpr:
OpenMPI 3.1.4 is due out on April 1.

This release includes a fix for open-mpi/ompi#6014, so remove our workaround and comments.

reidpr:
While we're at it, looks like libpsm2 is behind several versions, and let's check UCX too.

reidpr:
This was fixed in PR #418, not 318 as typo'ed in the commit log.

PR #418, opened 2019-04-18
Update OpenMPI examples to 3.1.4

heasterday:
This addresses #346. This PR increments the version of PSM2, ibverbs, and UCX within the image. This PR also removes workarounds for UCX issue #750 and Openmpi issue #6014. This PR also introduces a patch to OpenMPI's UCX PML plugin such that it will not activate on systems with Cray or Intel HSNs.


heasterday:
@reidpr Good catch on that variable. The UCX pml has a higher priority than the ob1 PML that provides access to the psm2 stuff, UCX doesn't compete for priority against PSM2 directly.  I adjusted the wording slightly to avoid confusion. Things look good.

Issue #403, opened 2019-04-03
Allow user to skip tests that require sudo


pagrubel:
Check for a flag such as CH_HAVE_SUDO before performing tests that need sudo

j-ogas:
We have a similar skipping mechanism for docker, e.g., `CH_TEST_SKIP_DOCKER=yes`. Perhaps we could include `CH_TEST_SKIP_SUDO=yes` to common.bash.

reidpr:
Right now we just test for access to the `sudo` command in `common.bash`. I think the right approach is just extend that test to only set the variable if it's unset and document it. It's already named for external access.

PR #420, opened 2019-04-22
tests: add CH_TEST_DONT_SUDO, move no-Docker tests

reidpr:
Addresses #403. This also tightens up Travis testing:

- do the no-Docker test just once instead of at the end of everything
- test without `sudo` privileges after build
- test with `sudo` turned off after build

We don't yet test without `sudo` at build time.

### DIFF
--- a/doc-src/test.rst
+++ b/doc-src/test.rst
@@ -20,6 +20,10 @@ at the top level of the source code or installed at
 
   $ cd test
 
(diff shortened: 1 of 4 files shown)


pagrubel:
I approve works fine for me.  I approve. I agree with Jordan on the name, but not really important.

Issue #419, opened 2019-04-18
make OpenMPI examples work on Crays (guest launch)


reidpr:
Split from #180. See that issue for a plan and some nodes on how to maybe do it.

Issue #380, opened 2019-03-12
fix "guest starts ranks" tests on Crays - OpenMPI


reidpr:
Split from #255 because the MPICH and OpenMPI problems are different.

Issue #255, opened 2018-09-27
fix "guest starts ranks" tests on Crays - MPICH


reidpr:
We're getting a bunch of independent rank-0 processes again.

heasterday:
I'm currently only seeing this behavior (independent rank-0 processes) after ch-fromhost --cray-mpi is called, guest launches with the original MPICH 3.2.1 libmpi (not cray-ified) don't appear to exhibit this behavior. Interestingly there are also process managment issues for OpenMPI 3.1.3 after overriding its libmpi on the crays. 

Towards that end I would like to propose we only use ch-fromhost --cray-mpi when host launch via srun is required, not prior to guest launches. In the case of guest launches this would preserve the MPI install in its original state and also not tie it to the libmpi of the host. For the case of guest launches I don't see any pressing need for anything from the MPI installs on the host since it is single node anyway. Guest launches would still require the unsetting of certain Slurm environment variables in the case of MPICH and OpenMPI.

I would like to get some thoughts from the team on this.

reidpr:
Discussed at today's meeting.

* We are going to split this into two issues because the MPICH and OpenMPI problems are different. #380 is for OpenMPI.

* Why does MPICH not work? Does it help to use Cray's `mpirun` instead of the open source one?

* The proposed fix (don't `--cray-mpi` for guest launch) would introduce a subtle ordering issue into the test suite. This may be another argument to get rid of `ch-fromhost`.

PR #426, opened 2019-04-22
tests: reorder mpi tests to do guest launches first

heasterday:
This PR address #419 , #380, and #255

### DIFF
--- a/test/common.bash
+++ b/test/common.bash
@@ -107,7 +107,7 @@ tarball_ok () {
     test -s "$1"
 }
 
-revert_mpi () {
+revert_crayify () {
     if [[ $ch_cray ]]; then
         $ch_mpirun_node ch-tar2dir "${ch_tardir}/${ch_tag}.tar.gz" "$ch_imgdir"
     else
--- a/test/common.bash
+++ b/test/common.bash
@@ -107,6 +107,14 @@ tarball_ok () {
     test -s "$1"
 }
 
+revert_mpi () {
+    if [[ $ch_cray ]]; then
+        $ch_mpirun_node ch-tar2dir "${ch_tardir}/${ch_tag}.tar.gz" "$ch_imgdir"
+    else
+        skip 'image mpi not modified'
+    fi
+}
+
 # Predictable sorting and collation
 export LC_ALL=C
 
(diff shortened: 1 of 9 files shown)


test/common.bash
```diff
@@ -77,6 +79,14 @@ prerequisites_ok () {
...
 }
 
+revert_crayify () {
+    if [[ $ch_cray ]]; then
+        $ch_mpirun_node ch-tar2dir "${ch_tardir}/${ch_tag}.tar.gz" "$ch_imgdir"
+    else
+        skip 'image mpi not modified'
+    fi
+}
+
```
j-ogas: If we are going to use an unpacking image function, I feel like it should be generic, e.g., `unpack_image ()` or perhaps `revert_image ()`.

That said, didn't we want to re-order the guest and host launches to avoid using such a function? `run_auto.bats` will unpack the image each time the test suite is executed.

FWIIW, I have suggested adding a function like this in this past. One of the benefits is that we can target portions of the test suite, e.g., `make test-test` without having to manually unpack the images or run `make test-run` prior. However, it is reasonable to argue that a quality of life change, such as this, doesn't really justify deviating from the standard test suite behavior.

reidpr:
> If we are going to use an unpacking image function, I feel like it should be generic, e.g., `unpack_image ()` or perhaps `revert_image ()`.

I agree; in my (pending) inevitable `tidy` commit I've done that.
 
> That said, didn't we want to re-order the guest and host launches to avoid using such a function? `run_auto.bats` will unpack the image each time the test suite is executed.

`run_auto.bats` is flawed in that it only unpacks on the first node. So, it doesn't reset things for a multi-node test.

> FWIIW, I have suggested adding a function like this in this past. One of the benefits is that we can target portions of the test suite, e.g., `make test-test` without having to manually unpack the images or run `make test-run` prior. However, it is reasonable to argue that a quality of life change, such as this, doesn't really justify deviating from the standard test suite behavior.

I do think you're on to something. I don't think this PR is the place for it, but I'd like a world where `make test-run` unpacks on all nodes in the allocation and `make test-test` leaves things as they started unless something really went wrong.

reidpr:
@heasterday, this is ready to go, pending your approval and successful tests on our Crays. Well done!

heasterday:
@reidpr passed the tests in full.

Issue #364, opened 2019-02-22
OCI compatibility layer to make Buildah work with ch-run instead of runc


reidpr:
Write and document an OCI compatibility layer for `ch-run` sufficient to make `buildah bud` and `buildah push` in `test/Build.buildah` work with it.

Tests including Build.buildah should pass in both Travis (Ubuntu) and the CentOS 7 VM image.

One useful resource from @rstyd: https://github.com/opencontainers/runtime-spec/blob/master/runtime.md

PR #431, opened 2019-04-26
OCI shim for ch-run

reidpr:
Addresses #364 and #471.

This PR will eventually introduce an OCI compatibility shim for `ch-run`, called `ch-run-oci`. The goal is to substitute Buildah and this shim for Docker, paving the way for a completely unprivileged workflow (this PR implements only Buildah with setuid helpers).

There is a [second OCI branch](https://github.com/hpc/charliecloud/tree/oci_runtime) as well, and you are encouraged to look at it. The point is that this branch will move somewhat slowly and carefully, with the goal of being merged before too long, while the other one moves quickly and is more experimental.

Partial to-do list:

* [x] update man pages throughout
* [x] update script help
* [x] rename doc files as needed
* [x] check over man pages
* [x] update dependency list
* [x] we do in fact need `kill`
  * but it doesn't actually interact with the user command
* [x] document required path of `--pid-file`
* [x] document general sloppiness and lack of error handling
* [x] validate OCI version
* [x] `--squash` breaks Buildah caching
* [x] remove `newuidmap`/`newgidmap` emulation
* [x] move Buildah stuff from `Build.buildah` into `ch-build`
* [x] pass tests
* [x] pass Travis


j-ogas:
Informative and well written. I have no suggestions to offer at this time.

reidpr:
So, the assumption of a privileged UID/GID map (installed with `newuidmap` and `newgidmap`) runs deep. When we replace them with simple unprivileged versions that create the same one-entry maps Charliecloud uses, we get:

```
STEP 1: FROM alpine:3.9
error creating build container: error creating container: error creating an ID-mapped copy of layer "a464c54f93a9e88fc1d33df1e0e39cca427d60145a360962e8f19a1dbf900da9": error creating copy of template layer "a464c54f93a9e88fc1d33df1e0e39cca427d60145a360962e8f19a1dbf900da9" with ID "325741a5babe70aae8321cbae05f6210e7607fb46c7e04ecbdbf90a761624a6b": there might not be enough IDs available in the namespace (requested 0:65534 for /scratch/buildah/graphroot/vfs/dir/325741a5babe70aae8321cbae05f6210e7607fb46c7e04ecbdbf90a761624a6b/etc/shadow): lchown /scratch/buildah/graphroot/vfs/dir/325741a5babe70aae8321cbae05f6210e7607fb46c7e04ecbdbf90a761624a6b/etc/shadow: invalid argument
```

reidpr:
I ran into a Buildah bug. With version 1.8 packaged in the Project Atomic PPC, we were getting failures building any Dockerfile when the storage was configured with files rather than `STORAGE_DRIVER=vfs` and command line options (`mkdir /run/runc: permission denied`). Buildah 1.9.0 built from Git source seems OK so far.

doc-src/ch-build_desc.rst
```diff
@@ -18,10 +18,19 @@ vagaries of making the supported builders work smoothly with Charliecloud and
...
+  * :code:`ch-grow`: our internal builder
+
+Supported builders, privileged:
+
+  * :code:`buildah-runc`: Buildah in "rootless" mode with setuid
+    helpers, using the default :code:`runc` for :code:`RUN` instructions
+
+  * :code:`buildah-setuid`: Buildah in "rootless" mode with setuid helpers,
+    using :code:`ch-run` (via :code:`ch-run-oci`) for :code:`RUN` instructions
+
```
rstyd: A little confused by this section. So buildah-setuid uses the oci shim, but still uses new[u,g]idmap?

reidpr: Yes, that's correct.

rstyd: Ah! So fully unprivileged buildah isn't supported with this patch since we still have logistical issues to work out?

Issue #430, opened 2019-04-26
test 'broken image errors' fails on Ubuntu 19.04


Brofessional:
The broken image test does not work on Ubuntu 19. On centos7 the test completes normally.

`✗ broken image errors
   (in test file run/ch-run_misc.bats, line 576)
     '[[ $output = *"can't execve(2): true: No such file or directory"* ]]' failed
   ch-run[3864]: can't execve(2): true: Not a directory (charliecloud.c:437 20)
`

The output for `execve(2)` in Ubuntu 18 appears slightly different than the output expected.

Tested using the squashfs branch of charliecloud on both machines.

reidpr:
Good find! Shouldn't be too hard to fix.

PR #433, opened 2019-04-29
change broken images test

Brofessional:
addresses #430 

centos7 and ubuntu19 produce different output during the broken image test.
the test can now match output for centos7 and ubuntu19

### DIFF
--- a/test/run/ch-run_misc.bats
+++ b/test/run/ch-run_misc.bats
@@ -570,10 +570,10 @@ EOF
     for f in $files $files_optional; do touch "${img}/${f}"; done
 
     # This should start up the container OK but fail to find the user command.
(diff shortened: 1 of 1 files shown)


test/run/ch-run_misc.bats
```diff
@@ -570,10 +570,10 @@ EOF
     for f in $files $files_optional; do touch "${img}/${f}"; done
 
     # This should start up the container OK but fail to find the user command.
-    run ch-run "$img" -- true
+    run ch-run "$img" -- /bin/true
```
reidpr: You'll need to change all the other `true` run in a container to `/bin/true` as well (e.g., line 583). Run on the host, no need (e.g., line 582).

PR #437, opened 2019-05-02
OpenMPI examples: downgrade libpsm2, libibverbs, UCX

heasterday:
The newer version is less compatible among systems, notably the ARM systems I have been testing on, than the previous version without offering any apparent benefits, this PR reverts it back prior to:
https://github.com/hpc/charliecloud/commit/00f630dea5e82170b0a9b67a565fc8f3d3a55432

The issue is that the current version we use is too new and there is a mismatch between it and the host drivers resulting in failures when using an ib device.


heasterday:
Removed the ready to merge label for now as I want to investigate this issue more

heasterday:
I'm seeing interesting issues with UCX in my testing so I'm reverting the UCX version back to 1.3.1 while I investigate

reidpr:
Partially reverts PR #418.

Issue #445, opened 2019-05-15
add ch-test helper script


j-ogas:
During an offline meeting we discussed adding a new script, `ch-test`, to facilitate test suite usage.

Issue #329, opened 2019-01-10
test suite: missing error message if Charliecloud not built


rstyd:
Running the test suite without charliecloud being built results in this error message. 

```
/usr/local/libexec/bats-exec-suite: line 21: let: count+=: syntax error: operand expected (error token is "+=")
make: *** [test-build] Error 1
```

This is confusing and should be replaced by something explaining how the Charliecloud executables cannot be found.

reidpr:
This is what I get:

```
$ cd ~/charliecloud
$ make clean
[...]
$ cd test
$ make test
./make-auto build chtest/Build Build.missing Dockerfile.alpine36 Dockerfile.alpineedge Dockerfile.centos6 Dockerfile.centos7 Dockerfile.debian9 Dockerfile.mpich Dockerfile.nvidia Dockerfile.openmpi Dockerfile.python3 Dockerfile.ubuntu1604 Docker_Pull.alpine36_dp ../examples/mpi/lammps/Dockerfile ../examples/mpi/paraview/Dockerfile ../examples/other/spark/Dockerfile ../examples/serial/hello/Dockerfile ../examples/serial/obspy/Dockerfile ../examples/serial/spack/Dockerfile ../examples/mpi/mpibench/Dockerfile.mpich ../examples/mpi/mpibench/Dockerfile.openmpi ../examples/mpi/mpihello/Dockerfile.mpich ../examples/mpi/mpihello/Dockerfile.openmpi > build_auto.bats
bats build.bats build_auto.bats build_post.bats
Must build with "make" before running tests.

/home/reidpr/charliecloud/test/bats/libexec/bats-exec-suite: line 20: let: count+=: syntax error: operand expected (error token is "+=")
Makefile:30: recipe for target 'test-build' failed
make: *** [test-build] Error 1
```

Are you not getting the "Must build" line?

The confusing error from `bats-exec-suite` comes from BATS. I don't know if or how it can be fixed. That's why we have a blank line after our error, to try to make it a little more obvious.

reidpr:
OK. Thanks for the report. Let's leave it alone for now, unless someone else encounters it, because you know how to work around it and so far you're the only one who's seen it.

j-ogas:
FWIIW, in my experience, that cryptic error is almost always due to malformed `CH_TEST_*` variables, e.g., `CH_TEST_TARDIR=./var/tmp/tar`, or the path has nested directories that do not exist #347.

reidpr:
[...]

Issue #278, opened 2018-11-01
tests fail if $BATS_TMPDIR is not /tmp


reidpr:
We use `$BATS_TMPDIR` inside the test containers. If it's not `/tmp`, then it's not mapped to the same path both inside and outside, and the test suite fails.

PR #447, opened 2019-05-15
new feature: ch-test

j-ogas:
Addresses #445.


doc-src/ch-test.rst
```diff
@@ -0,0 +1,9 @@
+:orphan:
+
+ch-test man page
+++++++++++++++++++++++
+
+.. include:: ./ch-test_desc.rst
+.. include:: ./bugs.rst
+.. include:: ./see_also.rst
+.. include:: ./docker_tips.rst
```
reidpr: Probably don't need the Docker tips for this one.

j-ogas:
Aside from a minor travis fix in `run/ch-test.bats` to address the the installation case testing, this is ready to be reviewed.

How this will be integrated with travis is not clear. I don't imagine doubling the test time to run the tests via `test/Makefile` and again via ch-test is an option. Do we then alter travis to run the test suite exclusively through `ch-test`?

bin/ch-test
```diff
@@ -0,0 +1,200 @@
...
+        make test-run
+    ;;
+    examples)
+        make test-test
+    ;;
+    all)
+        perm_dirs
+        make test-build
+        make test-run
+        make test-test
```
reidpr: How about just `make test`?

j-ogas:
> My overall impression was that it seems kind of complicated for what basically is a `cd` followed by `make`. However, I became less sure of that as I read on. I do think some of the suggestions could make things simpler.

Re. complexity: The summary flag here is for `test/run/ch-test.bats`. If we use ch-test exclusively with CI, travis tests passing would imply the that ch-test script is working. In which case we probably don't need the ch-test.bats file or the summary option, which will also simplify things.

bin/ch-test
```diff
@@ -0,0 +1,200 @@
+#!/bin/bash
+
+libexec="$(cd "$(dirname "$0")" && pwd)"
+. "${libexec}/base.sh"
+. "${libexec}/version.sh"
+
+check_env () {
```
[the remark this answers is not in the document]
j-ogas: The issue here is that prefix needs `tar` or `img` appended depending on which variable is being exported. For example, using your approach we have something like this:

```bash
  7 check_env () {
  8     local var=$1
  9     local default=$2
 10     if [[ -n ${!var} ]]; then
 11         printf '%s set and will be used: %s=%s\n' "$var" "$var" "${!var}"
 12         export $var=${!var}
 13     elif [[ -n $prefix ]]; then
 14         printf '%s unset, using prefix: %s=%s\n' "${var}" "${var}" "$prefix"
 15         export $var=$prefix
 16     else
 17         printf '%s unset, using default: %s=%s\n' "${!var}" "${!var}" "$default"
 18         export $var=$defualt
 19     fi
 20     exit 0
}
```

Do you have any suggestions on a simple way to do this for the middle `elif` value? Also, I'm not sure that is any less complicated than:

```bash
check_env () {
    if [[ -n ${!1} ]]; then
        printf "%s=%2" "$1" "${!1}"
    fi
}
[...]
# Set environment
[[ -n $prefix ]]         || prefix=/var/tmp
[[ -n $scope ]]          || scope=standard
[[ -n $CH_TEST_TARDIR ]] || export CH_TEST_TARDIR=${prefix}/tar
[[ -n $CH_TEST_IMGDIR ]] || export CH_TEST_IMGDIR=${prefix}/dir
[[ -n $CH_TEST_SCOPE ]] || export CH_TEST_SCOPE=$scope
```

bin/ch-test
```diff
@@ -0,0 +1,200 @@
...
+    if [[ -n $2 ]]; then
+        printf "prefix: warning: %s set and will be used\n" "$1"
+    fi
+}
+
+fatal () {
+    printf '%s\n\n' "$1" 1>&2
+    exit 1
+}
```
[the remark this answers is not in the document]
j-ogas: Not yet. #454 is still a work in progress (handling debug printing at the base.sh level for ch-fromhost is the current holdup).

bin/ch-test
```diff
@@ -0,0 +1,200 @@
...
+        export CH_TEST_PERMDIRS="$prefix"
+    fi
+fi
+
+# Avoid cryptic bats error (#347) that occurs when setting a CH_TEST variable
+# to a path where there are two or more directories that do not exist.
+if [[ -z $summary ]]; then
+    [[ -d $CH_TEST_TARDIR ]] || mkdir -p "$CH_TEST_TARDIR"
+    [[ -d $CH_TEST_IMGDIR ]] || mkdir -p "$CH_TEST_IMGDIR"
+fi
```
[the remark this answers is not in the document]
j-ogas: The test suite errors after checking `CH_TEST_[TAR|IMG]DIR` but before creating it:
```
$ export CH_TEST_TARDIR=/var/tmp/fizz/buzz
$ make test-build
bats build.bats build_auto.bats build_post.bats
/home/jogas/git/charliecloud/test/bats/libexec/bats-exec-suite: line 20: let: count+=: syntax error: operand expected (error token is "+=")
```
I believe this occurs due to readlink in common.bash:
```
ch_imgdir=$(readlink -ef "$CH_TEST_IMGDIR")
ch_tardir=$(readlink -ef "$CH_TEST_TARDIR")
```

j-ogas:
Also related to #329.

j-ogas:
> I think overall this looks solid, but I'd like to discuss and walk through it in person. Can we do that next week?

Sounds good.

> One thing I thought to add: There are currently a few setup/teardown tests, I think mostly to manage the test directories. I think these maybe should be moved to `ch-test`?

Agreed.

reidpr:
Add check/workaround for #278?

doc-src/ch-test_desc.rst
```diff
@@ -3,159 +3,178 @@ Synopsis
...
-    Specify prefered image builder. Default: :code:`$CH_BUILDER`.
+  :code:`mk-perm-dirs`
+    Create the filesystem permissions directories. Requires
+    :code:`--perm-dirs`.
 
-  :code:`--dry-run`
-    Print test suite phase details without executing.
+  :code:`clean`
+    Delete test artifacts on disk and all images in builder storage (whether
+    or not Charliecloud-related), except for the permissions test fixtures.
```
j-ogas: Spicy. Not sure I'd like `clean` to wipe my docker/buildah image cache. Am I misunderstanding this?

doc-src/ch-test_desc.rst
```diff
@@ -3,159 +3,178 @@ Synopsis
...
-* Builder storage (e.g., layer cache). This goes wherever the builder puts it.
+  :code:`--img-dir DIR`
+    Set unpacked images directory to :code:`DIR`. In a multi-node allocation,
+    this directory may not be shared between nodes. Default:
+    :code:`$CH_TEST_IMGDIR` if set, or :code:`/var/tmp/img`.
 
-* Compressed image directory, i.e., :code:`--tar-dir DIR`.
+  :code:`--pack-dir DIR`
+    Set packed images directory to :code:`DIR`. Default:
+    :code:`$CH_TEST_TARDIR` if set, or :code:`/var/tmp/pack`.
```
j-ogas: Perhaps we should go ahead and update `CH_TEST_TARDIR` to `CH_TEST_PACKDIR` as well? I know it is long, but I'm also wondering if `CH_TEST_IMGDIR` should be `CH_TEST_UNPACKDIR`.

bin/ch-test
```diff
@@ -0,0 +1,322 @@
...
+        method='environment variable'
+    else
+        builder_choose
+        method='default'
+    fi
+    printf 'setting builder to %s (%s)\n' "$CH_BUILDER" "$method"
+    if [[ $CH_BUILDER == 'ch-grow' ]]; then
+        set_chvar CH_GROW_STORAGE 'ch-grow storage' '' "$CH_GROW_STORAGE" /var/tmp/ch-grow_
+    fi
+}
```
reidpr: I still feel like there's probably a better way to do this than have a function specific for a single variable, but also that you've probably put enough work into it and we should leave it alone for now.

bin/ch-test
```diff
@@ -0,0 +1,322 @@
...
+usage=$(cat <<EOF
+Run the Charliecloud test suite.
+
+Usage:
+
+  $ $(basename "$0") [PHASE] [OPTIONS]
+EOF
+)
+
+check_builder () {
```
reidpr: I tend to name things most-significant-first, i.e. this would be called `builder_check` and below `set_builder` would become `builder_set`. The reasoning is that then when you alphabetize them, related functions end up near one another. I suggest making this change.

doc-src/ch-test_desc.rst
reidpr: I guess the reasoning to leave the variables along is that it's a pretty pervasive change that alters an existing API, so I wanted to leave it alone for now, in case the new vocabulary is not satisfactory.

doc-src/ch-test_desc.rst
reidpr: Your understanding is correct. Why don't we leave it alone for now (i.e., don't wipe builder storage). I would like an easy way to do that though; our current `test/docker-clean.sh` isn't very user-facing.

test/travis.bash
```diff
@@ -42,31 +42,40 @@ esac
...
 make
 bin/ch-run --version
-version=$(cat VERSION.full)
 
 if [[ $INSTALL ]]; then
     sudo make install PREFIX="$PREFIX"
-    cd "$PREFIX/libexec/charliecloud-$version"
+    ch_test="${PREFIX}/bin/ch-test"
+else
+    ch_test=$(readlink -f bin/ch-test)
```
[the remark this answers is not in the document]
j-ogas: Removing the absolute path here will cause travis to error (https://travis-ci.org/hpc/charliecloud/jobs/604006996?utm_medium=notification&utm_source=email). 

```bash
147 top_dir=$(readlink -f "${ch_bin%/*}")
[...]
173 # Determine source vs prefix installation.
174 if [[ -d "${top_dir}/test" ]]; then
175     # source install
176     testdir=${top_dir}/test
177 else
178     # prefix installation
179     testdir=${top_dir%/charliecloud*}/libexec/charliecloud-$(version 2>&1)/test
180 fi
```


The check for `${top_dir}/test` fails (it shouldn't, it is a source install). I feel like this is because `base.sh` uses `dirname "$0"` instead of `dirname "${BASH_SOURCE[0]}`. When we do `ch_test=$(readlink -f bin/ch-test` in the `travis.bash`, the script the problem is resolved (https://travis-ci.org/hpc/charliecloud/jobs/604009862?utm_medium=notification&utm_source=github_status).

bin/ch-test
j-ogas: > I still feel like there's probably a better way to do this than have a function specific for a single variable...

I completely agree. Unfortunately I haven't been able to come up with anything better. Would you like to take a shot at it?

Issue #446, opened 2019-05-15
handle install when test suite is not built


j-ogas:
Related to #389 and #381.

When valid version of python is not detected the top level Makefile will not descend into the test suite and attempt and to build it. However, the top level Makefile still references files, e.g., libsotest.so via `make install` that have not yet been built:

```
#       shared library tests
    install -d $(TEST)/sotest $(TEST)/sotest/bin $(TEST)/sotest/lib
    install -pm 755 -t $(TEST)/sotest test/sotest/libsotest.so.1.0 \
                                      test/sotest/sotest
    install -pm 644 -t $(TEST)/sotest test/sotest/files_inferrable.txt \
                                      test/sotest/sotest.c
    ln -sf ./libsotest.so.1.0 $(TEST)/sotest/libsotest.so
    ln -sf ./libsotest.so.1.0 $(TEST)/sotest/libsotest.so.1
    install -pm 755 -t $(TEST)/sotest/bin test/sotest/bin/sotest
    install -pm 755 -t $(TEST)/sotest/lib test/sotest/lib/libsotest.so.1.0
```

j-ogas:
I propose we do not include the tests and examples in the absence of python. The two thoughts behind this are: 1)  the test suite is dependent on python and 2) we can eliminate the `BuildRequires /usr/bin/python3` from the base charliecloud package.

wiene:
I've just stumbled across this issue:
```
install: cannot stat 'test/sotest/libsotest.so.1.0': No such file or directory
install: cannot stat 'test/sotest/sotest': No such file or directory
```
What are your plans concerning PR #453?

j-ogas:
@wiene We plan to merge it. I recently merged changes from master and requested another review. I suspect it will be merged soon.

PR #453, opened 2019-05-30
separate test make install

j-ogas:
Addresses #446.


j-ogas:
Couple of notes: 1) the `test/Makefile` makes use of `cp` instead of `install`, 2) I used 8 space tabs, which seems to have caused changes in block indention.

Makefile
```diff
@@ -150,52 +149,9 @@ install: all
...
-	    install -pm 644 -t $(TEST)/bats test/bats/CONDUCT.md \
-	                                    test/bats/LICENSE \
-	                                    test/bats/README.md && \
-	    install -d $(TEST)/bats/libexec && \
-	    install -pm 755 -t $(TEST)/bats/libexec test/bats/libexec/* && \
-	    install -d $(TEST)/bats/bin && \
-	    ln -sf ../libexec/bats $(TEST)/bats/bin/bats && \
-	    ln -sf bats/bin/bats $(TEST)/bats; \
+#	install test suite and examples if the right python is found
+	if (command -v "$$(head -1 test/make-auto | sed -E 's/^.+ //')"); then \
```
reidpr: Don't we use this `command` business a second time somewhere? Should probably put it in a variable.

test/Makefile
```diff
@@ -1,14 +1,14 @@
 export LC_ALL := C
 images_ch := chtest/Build \
-             $(sort $(wildcard ./Build.*)) \
-             $(sort $(wildcard ./Dockerfile.*)) \
-             $(sort $(wildcard ./Docker_Pull.*))
+		 $(sort $(wildcard ./Build.*)) \
+		 $(sort $(wildcard ./Dockerfile.*)) \
+		 $(sort $(wildcard ./Docker_Pull.*))
```
reidpr: Spurious tab characters introduced in several places. The only place we want ASCII 9 in the file is in Make recipes where it is required, and there only once.

test/Makefile
```diff
@@ -86,3 +87,87 @@ sotest/libsotest.so.1.0: sotest/libsotest.c
...
+INSTALL_PREFIX := $(if $(DESTDIR),$(DESTDIR)/$(PREFIX),$(PREFIX))
+# LIBEXEC_DIR is modeled after FHS 3.0 and
+# # https://www.gnu.org/prep/standards/html_node/Directory-Variables.html. It
+# # contains any executable helpers that are not needed in PATH. Default is
+# # libexec/charliecloud which will be preprended with the PREFIX.
+LIBEXEC_DIR	?= libexec/charliecloud-$(VERSION)
+LIBEXEC_INST 	:= $(INSTALL_PREFIX)/$(LIBEXEC_DIR)
+LIBEXEC_RUN 	:= $(PREFIX)/$(LIBEXEC_DIR)
+EXAMPLES 	:= $(LIBEXEC_INST)/examples
+TEST            := $(LIBEXEC_INST)/test
```
reidpr: Most of these variables should be passed in from the parent Make rather than re-computed. The manual section on "recursive make" may be helpful on this common pattern.

test/Makefile
```diff
@@ -1,23 +1,23 @@
...
+images :=   $(images_ch) $(images_eg)
 
 # Favor embedded Bats, if installed, over system Bats.
 export PATH := $(CURDIR)/bats/bin:$(PATH)
 
 # Used by "make all" at top level to build these files for "make install".
 .PHONY: all
 all: build_auto.bats run_auto.bats \
-     sotest/bin/sotest sotest/lib/libsotest.so.1.0
+	sotest/bin/sotest sotest/lib/libsotest.so.1.0
```
reidpr: All the above indentation should stay unchanged.

test/Makefile
```diff
@@ -86,3 +87,87 @@ sotest/libsotest.so.1.0: sotest/libsotest.c
...
+	cd ../ && for i in examples/*; do \
+	cp -rp $$i $(EXAMPLES); \
+	done
+	chmod 755 $$(find $(EXAMPLES) -type d)
+	chmod 755 $(EXAMPLES)/serial/hello/hello.sh \
+	          $(EXAMPLES)/syscalls/pivot_root \
+	          $(EXAMPLES)/syscalls/userns \
+	          $(EXAMPLES)/*/*/*.sh;
+#	tests
+	cd ../ && cp -rp test $(TEST)
```
reidpr: You might comment somewhere why we are using `cp(1)` rather than `install(1)`.

j-ogas:
White space issues resolved; reverted `cp` lines to `install`. Adding ready label.

Issue #451, opened 2019-05-23
refactor scripts to use error function


j-ogas:
We have a lot of repetitive `exit` and `stdout` redirection in our scripts that would be handled more elegantly with a function:

j-ogas:
Was this addressed or partially addressed with #698?

reidpr:
Yes, I think this issue is now “refactor the other scripts to use it”.

PR #454, opened 2019-06-03
add output functions fatal, info, debug

j-ogas:
Addresses #451.

### DIFF
--- a/bin/base.sh
+++ b/bin/base.sh
@@ -6,6 +6,10 @@ ch_bin="$(cd "$(dirname "$0")" && pwd)"
 libexec="$(cd "$(dirname "$0")" && pwd)"
 . "${libexec}/version.sh"
 
+fatal () {
+    printf '%s\n\n' "$1" 1>&2
+    exit 1
+}
 
 parse_basic_args () {
     for i in "$@"; do
--- a/bin/base.sh
+++ b/bin/base.sh
@@ -15,12 +15,12 @@ debug () {
 }
 
 fatal () {
-    echo "$progname: $1" 1>&2
+    echo "${progname}: ${1}" 1>&2
     exit 1
 }
 
 info () {
-    echo "$progname: $1" 1>&2
+    echo "${progname}: ${1}" 1>&2
 }
 
 parse_basic_args () {
--- a/bin/base.sh
+++ b/bin/base.sh
@@ -1,14 +1,26 @@
 set -e
 
 # shellcheck disable=SC2034
-ch_bin="$(cd "$(dirname "$0")" && pwd)"
+ch_bin=$(cd "$(dirname "$0")" && pwd)
 
-libexec="$(cd "$(dirname "$0")" && pwd)"
+libexec=$(cd "$(dirname "$0")" && pwd)
 . "${libexec}/version.sh"
 
+progname=$(basename "$0")
+
+debug () {
+    if [ "$verbose" ]; then
+        echo "$1" 1>&2
+    fi
+}
+
 fatal () {
-    printf '%s\n\n' "$1" 1>&2
+    echo "$progname: $1" 1>&2
     exit 1
+}
+
+info () {
+    echo "$progname: $1" 1>&2
 }
 
 parse_basic_args () {
--- a/bin/base.sh
+++ b/bin/base.sh
@@ -9,7 +9,9 @@ libexec=$(cd "$(dirname "$0")" && pwd)
 progname=$(basename "$0")
 
 debug () {
-    echo "$1" 1>&2
+    if [ -n "$verbose" ]; then
+        echo "${progname}: ${1}:" 1>&2
+    fi
 }
 
 fatal () {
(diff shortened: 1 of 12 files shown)


bin/ch-tar2dir
```diff
@@ -51,7 +45,7 @@ for ext in '' .tar.gz .tar.xz .tgz; do
     fi
     tarball=$c
     if [ -n "$ext" ]; then
-        echo "found: ${tarball}" 1>&2
+        echo "found: $tarball" 1>&2
```
reidpr: Recall style is to enclose variable expansion in curly brackets unless it's the only thing in the string.

examples/mpi/mpihello/slurm.sh
```diff
@@ -9,16 +9,15 @@ tar=$1
 imgdir=$2
 img=${2}/$(basename "${tar%.tar.gz}")
 
-if [[ -z $tar ]]; then
-    echo 'no tarball specified' 1>&2
+fatal () {
+    printf '%s\n\n' "$1" 2>&1
```
reidpr: I think the `slurm.sh` scripts should not be updated, since they are intended to be examples and ought to work outside Charliecloud.

j-ogas: I did question whether this script should also be altered. The primary justification for doing updating this script is consistency. The new syntax is much sharper and easier read. FWIIW, the example script should still run outside of charliecloud. Thoughts?

bin/base.sh
```diff
@@ -1,11 +1,25 @@
...
+ch_bin=$(cd "$(dirname "$0")" && pwd)
 
-libexec="$(cd "$(dirname "$0")" && pwd)"
+libexec=$(cd "$(dirname "$0")" && pwd)
 . "${libexec}/version.sh"
 
+progname=$(basename "$0")
+
+debug () {
+    echo "$1" 1>&2
```
reidpr: `debug` should be silent unless we're in verbose mode. Previously it tested `$verbose` but I'm open to argument on how to make it happen.

bin/ch-fromhost
```diff
@@ -140,7 +125,7 @@ queue_files () {
                 ;;
         esac
         if is_so "$f"; then
-            debug "found shared library: ${f}"
+            [ -z "$verbose" ] || debug "found shared library: ${f}"
```
reidpr: Aha, here's where that test went. It needs to go inside `debug`; part of the point of that function is to insulate the caller from needing to test `$verbose`.

examples/mpi/mpihello/slurm.sh
reidpr: I think we should keep the examples simple, and it's simpler to have direct if statements rather than encapsulating them in another function for the user to think about. You might check with @heasterday. Do we promote this kind of error handling in user job scripts?

test/run/ch-run_uidgid.bats
```diff
@@ -100,29 +100,29 @@ setup () {
     # busybox seems to use the same list
     case $status in
         0)      # "success"
-            printf 'RISK\tsuccessful mount\n'
+            info 'RISK    successful mount'
```
reidpr: Here's a case where `info` is not appropriate. The `RISK`/`SAFE` lines are output, not log messages; e.g., it will break if we add the program name to each `info` line, which is a common thing to do for logging.

Also, I want the `\t` in there instead of a literal tab character so it's obvious it's a tab.

bin/base.sh
j-ogas: The only script that uses (and sets) the `--verbose` argument is `ch-fromhost`. So the reference to `$verbose` in the `base.sh` is undefined.

bin/ch-fromhost
j-ogas: The current implementation is admittedly ugly but it keeps `debug` defined in base.sh without having to alter much else.

We could add `--verbose` to `parse_basic_args` and remove that argument handling, along with `-v`, from ch-fromhost. Thoughts?

reidpr: > We could add `--verbose` to `parse_basic_args` in `base.sh` and remove that argument handling, along with `-v`, from ch-fromhost. Thoughts?

Currently `parse_basic_args` either does nothing or prints something and exits. If we add an argument that does not cause an exit, that argument must be consumed. If it's the first argument in `$@` that's easy because we can `shift` it off, but I don't know what to do if it's in the middle. Recall that problem was nearly impossible in POSIX sh when we were working on `ch-build2dir`.

reidpr: But, it would work to expect the caller to deal with `--verbose` and set `$verbose` if needed.

j-ogas: Hmm, yes it isn't clear to me how to gracefully handle the verbose argument and set the verbose variable at the POSIX shell base.sh level.

I tinkered with a kludge in which the variable was handled twice, once when setting `verbose` in base.sh and again in ch-fromhost where the case essentially does nothing (see recent commits). It works but is it pretty horrendous.

Perhaps ch-fromhost should just use its own defined debug function? I'll think about this.

reidpr:
Reviewing this conversation again, I think it would be reasonable to expect the caller to set `$verbose` (`$ch_verbose`?), and all the scripts other than `ch-fromhost` just use the default. I'm not sure what the right default would be though.

Once that question is addressed, are we good to update this and get it merged?

j-ogas:
This PR is ancient now. Perhaps we may revisit in the future. Closing.

Issue #428, opened 2019-04-26
test that we have the right man pages


reidpr:
Man pages don't show up in `man/` automatically nor in see also sections. Test that:

* We have a man page for each command, plus `charliecloud(1)`, and no others.
* Each man page appears in "See also" section in `charliecloud(1)`.
* Blurbs in `doc-src/conf.py` match `command-usage.rst` match the command help text.

PR #457, opened 2019-06-10
check man page sanity

reidpr:
Addresses #428.

### DIFF
--- a/doc-src/command-usage.rst
+++ b/doc-src/command-usage.rst
@@ -8,9 +8,6 @@ Charliecloud commands. Its content is identical to the commands' man pages.
    :depth: 1
    :local:
 
-.. WARNING: The one-line summaries below are duplicated in list man_pages in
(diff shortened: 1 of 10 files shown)


doc-src/command-usage.rst
```diff
@@ -19,15 +16,14 @@ Charliecloud commands. Its content is identical to the commands' man pages.
 ch-build
 ++++++++
 
-Wrapper for :code:`docker build` that works around some of its annoying
-behaviors.
+Wrapper for :code:`docker build` with various enhancements.
```
j-ogas: Prefer the former.

test/docs-sane
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env python
```
j-ogas: We should probably "versionize" this in in our spec file, e.g., `%{versionize_script python3 test/docs-sane}`

Issue #184, opened 2018-06-13
completely unprivileged Charliecloud workflow


reidpr:
Possibilities:

* img (#154)

j-ogas:
I have been investigating this topic. Thus far, I have come across a variety of tools that each seem to tackle a piece of the puzzle but offer no comprehensive solution. 

- [Skopeo](https://github.com/projectatomic/skopeo) — gives us the ability to copy images from distribution based registries (dockerhub, etc.) and container storage backends (docker daemon). For example, I was able to use skopeo to pull Alan McPherson's CoMD file from Dockerhub from within an unprivileged container. However, I did run into problems when trying to use Skopeo behind a proxy.

- [umoci](https://github.com/openSUSE/umoci) — intends to be a complete manipulation tool for OCI images. It _should_ be usable with Skopeo; umoci should be able to unpack and repack images from json blob and manifest produced from Skopeo. However, I have yet to successfully test this (I receive ann error stating that CoMD image is not an OCI image).

- [img](https://github.com/genuinetools/img) — a stand alone, daemon-less unprivileged image builder

We should note that all of these tools will implement dependencies (like go). Alternatively, we could bundle tools like these within a Charliecloud container and save the resulting images produced from the tools via `ch-run` arguments (`-w`, etc.). Just a thought.

reidpr:
More possibilities:

* [buildah](https://github.com/projectatomic/buildah)
* Docker inside `ch-run -u 0`.

j-ogas:
Running `docker build` inside a debian:stretch container  with Docker installed via `ch-run -u 0` produced the following:

```
root@localhost:/# docker build -t spack .
error checking context: 'no permission to read from '/dev/autofs''.
root@localhost:/# systemctl status docker
Failed to connect to bus: No such file or directory
```

I also attempted to containerize buildah (utility container #191 ) but the _inception_ didn't take. The "dream within a dream" failed (runc within charliecloud):
```
ERRO[0000] error reading allowed ID mappings: error reading subuid mappings for user "jogas" 
and subgid mappings for group "jogas": No subuid ranges found for user "jogas"
```
I took a shot in the dark and put `jogas:1000:1020` into both `/etc/subuid` and `/etc/subgid`:
```
ERRO[0000] error opening "allow" to /proc/18521/setgroups: write /proc/18521/setgroups: 
operation not permitted 
ERRO[0000] (unable to determine exit status)
```

reidpr:
[...]

PR #460, opened 2019-06-11
Add support for unprivileged container builds with Buildah

rstyd:
This PR addresses #364 and #184.  There are a few travis tests that do not currently pass.


reidpr:
Also addresses #479.

reidpr:
Looks like we don't need this now that #431 and #483 are merged. Thanks for the prototype!

PR #462, opened 2019-06-16
Add missing CFLAGS, CPPFLAGS and LDFLAGS for some examples/tests

wiene:
This patch ensures that CFLAGS, CPPFLAGS and LDFLAGS settings are
respected when compiling/linking (lib)sotest, pivot_root and userns.

I was kindly made aware of this issue by the [blhc](https://ruderich.org/simon/blhc) tool.

### DIFF
--- a/examples/syscalls/Makefile
+++ b/examples/syscalls/Makefile
@@ -10,4 +10,4 @@ clean:
 $(BINS): Makefile
 
 %: %.c
-	gcc $(CFLAGS) $< -o $@
+	gcc $(CFLAGS) $(CPPFLAGS) $(LDFLAGS) $< -o $@
--- a/test/Makefile
+++ b/test/Makefile
@@ -81,8 +81,8 @@ sotest/lib/libsotest.so.1.0: sotest/libsotest.so.1.0
 # the container. (Issue #227.)
 
 sotest/sotest: sotest/sotest.c sotest/libsotest.so.1.0
-	gcc -o $@ -L./sotest -lsotest $^
+	gcc -o $@ $(CFLAGS) $(CPPFLAGS) $(LDFLAGS) -L./sotest -lsotest $^
 sotest/libsotest.so.1.0: sotest/libsotest.c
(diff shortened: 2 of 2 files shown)


examples/syscalls/Makefile
```diff
@@ -10,4 +10,4 @@ clean:
 $(BINS): Makefile
 
 %: %.c
-	gcc $(CFLAGS) $< -o $@
+	gcc $(CFLAGS) $(CPPFLAGS) $(LDFLAGS) $< -o $@
```
reidpr: These are C programs, so do we really want `$CPPFLAGS`?

wiene: Since the C compiler uses the C preprocessor, I think that it makes sense to also include C preprocessor flags.

wiene:
> I'd also be open to a PR adding `blhc` to the test suite. Do you think that's a good idea?

So far I have no experience with using `blhc` outside a Debian environment where it is part of the CI machinery. Citing `blhc`'s [README](https://ruderich.org/simon/gitweb/?p=blhc/blhc.git;a=blob_plain;f=README;hb=c16e86feba1469ae7ccac10e425953b055408a44) file (for version 0.09):

> At the moment it works only on Debian and derivatives but it should be easily
> extendable to other systems as well.
> 
> [...]
> 
> The build log must contain the following line which marks the beginning of the
> real compile process (output of dpkg-buildpackage):
> 
>     dpkg-buildpackage: ...
> 
> If it's not present no compiler commands are detected. In case you don't use
> dpkp-buildpackage but still want to check a build log, adding it as first line
> should work fine.

So in summary, I feel uncertain how easy or difficult it is to integrate it in the Charliecloud test suite.

examples/syscalls/Makefile
reidpr: Oops, I'm sorry. I thought `CPPFLAGS` was for C++.

reidpr:
> So in summary, I feel uncertain how easy or difficult it is to integrate it in the Charliecloud test suite.

OK. This part of the code doesn't change too much, and it sounds like you would catch it eventually in the Debian package, so that seems like a low priority to add it upstream?

Issue #440, opened 2019-05-03
Add test for Openmpi errors which don't result in failure


heasterday:
If a plugin such as UCX fails during initialization OpenMPI will fall back to tcp resulting in the test suite still passing but simply taking longer to run. This means we currently may not catch regressions in the images until a test is ran outside of the suite or a particularly observant tester notices the suite took longer than usual. 

I plan to add a basic test to check for this.

reidpr:
Related but not duplicate: #289.

PR #463, opened 2019-06-17
Add basic check for non-fatal MPI errors

heasterday:
Addresses #440


reidpr:
Back to you. Travis now fails, though I sure suspect that is something on their end since the change I made doesn't seem like it should matter. Merging from master may help.

heasterday:
Looks like python3-setuptools was needed for travis to be happy, which a merge from master resolved. This should be good now.

reidpr:
Great! Thanks for your patience.

PR #476, opened 2019-07-03
fixed typo in ch-grow doc

pagrubel:
just fixing a typo


reidpr:
Thanks. I'll just correct this on `ch-run-oci` for expediency.

PR #478, opened 2019-07-05
Makefile: Allow to override DOCDIR.

olifre:
This is needed to enforce a naming scheme
e.g. from a package manager on this path
(may contain downstream revisions for example).
Default unchanged.

Proposed at:
https://github.com/gentoo/gentoo/pull/12084


reidpr:
Thanks @olifre.

Is the name `DOCDIR` a standard convention, and that's why you propose changing `DOC`?

olifre:
> Is the name `DOCDIR` a standard convention, and that's why you propose changing `DOC`?

Exactly. :smile: The switch should be done now that the variable is "exposed" and externally configurable, so it matches expectation better. 
The standard is here:
https://www.gnu.org/prep/standards/html_node/Directory-Variables.html

Cheers,
Oliver

PR #481, opened 2019-07-06
Fix hardcoded calls to gcc.

olifre:
To be open for choice of different C compilers,
honour CC variable.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -10,7 +10,7 @@ images_eg := $(sort $(wildcard ../examples/*/*/Build)) \
              $(sort $(wildcard ../examples/*/*/Docker_Pull)) \
              $(sort $(wildcard ../examples/*/*/Docker_Pull.*))
 images := $(images_ch) $(images_eg)
-sotests := sotest/bin/sotest sotest/lib/libsotest.so.1.0
+sotests := check-no-icc sotest/bin/sotest sotest/lib/libsotest.so.1.0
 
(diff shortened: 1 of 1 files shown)


reidpr:
So you're right of course, but there's a bit of history there. The reason it's GCC is that Intel CC likes to link everything with extra Intel shared libraries, which don't get copied into the images and the test suite fails. I suppose we could just puke if `CC=icc`.

olifre:
If it's only `icc`, I think this would be the better choice. The problem with hardcoding `CC` is that some distributions complain about packages doing that, and theoretically, users may want to use e.g. `clang` as system compiler. 

Then we should probably do that both for the `test/Makefile` and `examples/syscall/Makefile`, right?

reidpr:
> Then we should probably do that both for the `test/Makefile` and `examples/syscall/Makefile`, right?

Yes, probably. I believe `bin/Makefile` does it already (implicitly).

olifre:
It seems the call in `examples/syscall/Makefile` was removed in the meantime, and only `test/Makefile` remains. I have rebased on current master and changed the PR to:
- Remove the hardcoded call to `gcc` and use `$(CC)` instead. 
- Error out if the `$(CC)` ends with `icc`. 

This is right now only done in `test/Makefile` since that seems to be the only explicit call.

reidpr:
Thanks @olifre! I'll merge this after 0.10 which we are trying to push out the door.

Issue #479, opened 2019-07-05
fully unprivileged Buildah


reidpr:
Follow-on to #431 to introduce the fully unprivileged (non-setuid) Buildah builds.

reidpr:
Tasks:

* [ ] add builder `buildah` to scripts
* [ ] add builder `buildah` to documentation
* [ ] document how to get the right version of Buildah
* [ ] pass tests locally
* [ ] pass Travis

PR #483, opened 2019-07-10
fully unprivileged Buildah

reidpr:
Addresses #479.

### DIFF
--- a/test/travis-install.bash
+++ b/test/travis-install.bash
@@ -17,6 +17,13 @@ fi
 # Project Atomic PPA provides buggy Buildah for Xenial, and we need Kevin's
 # patched version, so build from source.
 if [[ $CH_BUILDER = buildah* ]]; then
+    if [[ $CH_BUILDER = buildah ]]; then
+        buildah_repo=https://github.com/hpc/buildah
+        buildah_branch=chown-error-tolerant-patch
+    else
+        buildah_repo=https://github.com/containers/buildah
+        buildah_branch=v1.9.0
+    fi
     sudo add-apt-repository -y ppa:alexlarsson/flatpak
     sudo apt-get update
     sudo apt-get -y install libapparmor-dev libdevmapper-dev libglib2.0-dev \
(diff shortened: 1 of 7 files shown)


reidpr:
Passes Travis and full-scope on my box.

test/travis-install.bash
```diff
@@ -27,10 +34,9 @@ if [[ $CH_BUILDER = buildah* ]]; then
     mkdir /usr/local/src/go
     pushd /usr/local/src/go
     export GOPATH=$PWD
-    git clone https://github.com/containers/buildah \
-              src/github.com/containers/buildah
+    git clone $buildah_repo src/github.com/containers/buildah
```
kpelzel: Clone and checkout can be done in a single line with "--branch", but the current way should work fine as well.
```suggestion
    git clone $buildah_repo src/github.com/containers/buildah --branch $buildah_branch
```

test/travis.yml
```diff
@@ -67,9 +69,9 @@ jobs:
       env: CH_BUILDER=docker
 
     - <<: *stage_install
-      env: CH_BUILDER=buildah-runc    TARBALL=export       INSTALL=yes
+      env: CH_BUILDER=buildah         TARBALL=export       INSTALL=yes
     - <<: *stage_install
-      env: CH_BUILDER=buildah-setuid  TARBALL=export       INSTALL=yes
+      env: CH_BUILDER=buildah-runc    TARBALL=export       INSTALL=yes
```
kpelzel: Did you want buildah-setuid to also be included in the install stage?

reidpr: Good catch. No, because Travis does at most five jobs in parallel, so I'm trying to keep the stages to five jobs wide so they complete in one interval rather than two. Also, the Buildah tests take a lot longer.

test/travis-install.bash
reidpr: Yeah, you're right. One gotcha is that you can't give a commit hash to `--branch`, for reasons that seem silly to me. Because it works already and for that reason, I'm going to leave it as is, but good catch.

PR #491, opened 2019-07-19
Fix Vagrantfile to reflect change in charliecloud test location

rstyd:
The location of the test suite in vagrant images has been changed from `/usr/local/libexec/charliecloud` to `/usr/local/libexec/charliecloud-<version>`. The Vagrantfile has not been updated to reflect this change. I would prefer this solution to draw the version from `VERSION` in the root directory so it will not need to be updated with each new release, but I've been unable to find a suitable way to do that.

### DIFF
--- a/packaging/vagrant/Vagrantfile
+++ b/packaging/vagrant/Vagrantfile
@@ -83,8 +83,8 @@ Vagrant.configure("2") do |c|
     rpm --install ius-release.rpm
     yum -y install git2u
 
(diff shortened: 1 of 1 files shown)


rstyd:
> The vagrant file should also be updated to make ch-grow work out of the box IMO. However, that may be a different PR.

I agree. I think that warrants an issue. I'll go make it.

rstyd:
Looks good to me I approve.

Issue #494, opened 2019-07-24
ch-grow lark error: wrong "lark" dependency listed in documentation.


j-ogas:
I appear to have broken ch-grow in my test box environment. It no longer works. Has this issue been seen before?

```
$ pip3 list
[...]
lark                     0.0.4
lark-parser              0.7.1
[...]
$ ./bin/ch-run --version
0.10
$ ./bin/ch-build --builder ch-grow -t debian9 -f ./test/Dockerfile.debian9 ./test
building with: ch-grow
Traceback (most recent call last):
  File "/home/jogas/upstream/bin/ch-grow", line 177, in <module>
    class Main_Loop(lark.Visitor):
AttributeError: module 'lark' has no attribute 'Visitor'
```

j-ogas:
The issue was that I installed the _wrong lark_.

The incorrect way to install lark (what I did):

```
$ pip3 install lark-parser --user
```

The correct way:
```
$ pip3 install lark_parser --user
```
```

**Comment by j-ogas** [comment]
Turns out I'm not crazy (I hope). We appear to list `lark-parser` as the correct dependency in our install documentation.

**Comment by reidpr** [comment]
I'm not sure what's going on here. `lark-parser` (with hyphen) works for me:

```
$ pip3 uninstall lark-parser
[...]
  Successfully uninstalled lark-parser-0.7.1
$ bin/ch-grow --storage=$CH_TEST_TARDIR/_ch-grow -t foo -f test/Dockerfile.alpine39 test
missing dependency: Python module "lark"
$ pip3 install lark-parser
Collecting lark-parser
Installing collected packages: lark-parser
Successfully installed lark-parser-0.7.1
$ bin/ch-grow --storage=$CH_TEST_TARDIR/_ch-grow -t foo -f test/Dockerfile.alpine39 test
growing: /scratch/tarballs/_ch-grow/img/foo:latest
  2 FROM alpine:3.9
  4 RUN ['/bin/sh', '-c', 'apk add --no-cache bc']
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.9/community/x86_64/APKINDEX.tar.gz
(1/5) Installing ncurses-terminfo-base (6.1_p20190105-r0)
(2/5) Installing ncurses-terminfo (6.1_p20190105-r0)
[...]

PR #495, opened 2019-07-24
Fix incorrect lark_parser module name

rstyd:
This PR addresses #494.


rstyd:
The issue is actually the documentation itself. The documentation states we need the `lark-parser` module, but the module is actually called `lark_parser`. It's very subtle.

j-ogas:
Right, but @reidpr just showed that `lark-parser` works for him. I'm wondering if it is a pip3 version issue?

rstyd:
Ohhh. Now that is odd.

reidpr:
Let's please discuss the bug on the bug's issue.

PR #490, opened 2019-07-18
Handle ARG instruction in Dockerfile.

cmbiwer:
This PR allows `ch-grow` to parse the `ARG` instruction in a `Dockerfile`.

An example `Dockerfile`:
```
FROM centos:7
ARG TEST_MSG
RUN echo My test message is: ${TEST_MSG}
```

This will echo the message if built with: `docker build --build-arg TEST_MSG="ONE TWO THREE" --file Dockerfile .`

Now with `ch-grow` can do the same with: `ch-grow --build-arg TEST_MSG="ONE TWO THREE" --file Dockerfile .`

Follows the same appending CLI format as `docker build`, i.e. can do `ch-grow --build-arg ARG1=A --build-arg ARG2=b --build-arg ARG3=c` and so on.

PR #502, opened 2019-08-01
handle ARG instruction in Dockerfile

reidpr:
This extends PR #490 by @cmbiwer and also addresses issue #501.

The extensions are:

* generalize `ARG` and `--build-arg` syntax per Docker and Buildah documentation
* add tests
* document `--build-arg`

The one reduction in functionality is to accept exactly one `KEY[=VALUE]` per `--build-arg`. My reading of both the Docker and Buildah docs is that they have this behavior. Also, if we consume any number of variables, constructions like `ch-grow [...] --build-arg foo=bar baz` will consume `baz` as a build argument when it's intended as the context directory.

@cmbiwer, does this extended PR meet your needs? I apologize for creating a duplicate PR; I can't push to your branch.

### DIFF
--- a/bin/ch-grow
+++ b/bin/ch-grow
@@ -56,8 +56,9 @@ copy: "COPY"i ( _WS copy_chown )? ( copy_shell ) _NEWLINES
 copy_chown: "--chown" "=" /[^ \t\n]+/
 copy_shell: _WS WORD ( _WS WORD )+
 
-arg: "ARG"i _WS ( arg_set ) _NEWLINES
-arg_set: WORD
+arg: "ARG"i _WS ( arg_bare | arg_equals ) _NEWLINES
+arg_bare: WORD
+arg_equals: WORD "=" ( WORD | STRING_QUOTED )
 
 env: "ENV"i _WS ( env_space | env_equalses ) _NEWLINES
 env_space: WORD _WS LINE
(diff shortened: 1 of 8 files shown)


bin/ch-grow
```diff
@@ -505,12 +544,16 @@ def INFO(*args, **kwargs):
...
    DEBUG("executing: %s" % args)
-   INFO("\033[33m", end="")
+   color("33m", sys.stdout)
    cp = subprocess.run(args, env=env, stdin=subprocess.DEVNULL)
-   INFO("\033[0m", end="")
+   color("0m", sys.stdout)
    if (cp.returncode):
       FATAL("%s failed with return code %d" % (args[0], cp.returncode))
 
+def color(color, fp):
```
rstyd: Should the color stuff technically be in a different commit? I'm fine with keeping it here.

reidpr: Remember we squash-merge, so this will all become one commit.

You're right that it's a second concern and would be tidier in another PR. However, `ch-grow` output can't be tested without fixing #501, which is why it's here.

cmbiwer:
@reidpr Yes, looks like this PR still meets my use case. Thanks!

bin/ch-grow
rstyd: Sounds good to me!

rstyd:
Approved.

Issue #485, opened 2019-07-12
ch-run as root: root with non-zero gid refused (skip if requirements are not met)


j-ogas:
The `run/ch-run_escalated.bats` test `ch-run as root: root with non-zero gid refused ` has caused confusion for some time now. Section 2.2.2 of the run phase documentation states that a user needs to be a member of two groups. The actual requirement is that they need a sudoers config to allow groups other than root (GID 0) without a password.

In any case, the output from the error isn't helpful in describing why the test failed. This test should skip if the condition for the test isn't met.

```
✗ ch-run as root: root with non-zero gid refused
   (in test file run/ch-run_escalated.bats, line 70)
      `[[ $output = *'error ('* ]]' failed
   Sorry, user vagrant is not allowed to execute '/usr/bin/ch-run -v
--version' as root:vagrant on localhost.localdomain.
```

loveshack:
Note that the failure was under an account that satisfied the requirements (although README.TEST in the rpm packaging doesn't mention it). Also I get other failures, like in https://bugzilla.redhat.com/show_bug.cgi?id=1690046, if I set CH_TEST_DONT_SUDO.  Sorry I don't have time to debug it at present.

j-ogas:
@loveshack Does the sudoers config have `(ALL:ALL)`? For example, if the secondary group was `charlie`:

```
%charlie ALL=(ALL:ALL) NOPASSWD: ALL
```

reidpr:
It shouldn't require no password, just be able to sudo to the other group. That configuration (password required) works on my dev box.

That second part is an unusual configuration though, and the bug report stands regardless of the password requirement IMO.

loveshack:
This is in an el7 vagrant box (for which I can provide a recipe if
necessary); I can't run on bare metal under a Fedora-ish OS with user
namespaces, and I got other errors on bare Debian.  The vagrant account
under which things run is installed with

  $ sudo cat /etc/sudoers.d/vagrant
  %vagrant ALL=(ALL) NOPASSWD: ALL

like the same as the default %wheel configuration.  Changing that to
ALL:ALL does fix "74 ch-run as root", but it still fails on subsequent
tests.

PR #516, opened 2019-08-16
skip test "root with non-zero gid refused" if sudo not configured

reidpr:
Addresses #485.

@loveshack, does this skip the test (instead of failing) in your configuration?

### DIFF
--- a/test/run/ch-run_escalated.bats
+++ b/test/run/ch-run_escalated.bats
@@ -61,9 +61,20 @@ load ../common
 }
 
 @test 'ch-run as root: root with non-zero gid refused' {
-    scope quick
+    scope standard
     [[ -n $ch_have_sudo ]] || skip 'sudo not available'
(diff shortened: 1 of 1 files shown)


loveshack:
I'm now confused because I can't reproduce the original issue with the
development source or 0.9.10, although I think the Vagrant configuration
is the same.

However, I see a bunch of failures from "make test" with the development
sources, which differ on CentOS 7 and Debian 10 without most of the
optional test things installed.  I'll do some investigation.

reidpr:
Thank you @loveshack.

I think we'll proceed with merging this one. Please let us know if you spot a regression.

Regarding the other errors, the test suite should be passing at all times on master, so please do file bug(s) with what you learn.

loveshack:
You wrote: 

> Thank you @loveshack.
>
> I think we'll proceed with merging this one. Please let us know if you
> spot a regression.

It didn't seem to make any difference when I tried, so looks fine.  I'm
still puzzled.

PR #519, opened 2019-08-20
Add SC1007 to shellcheck ignored errors

loveshack:
This removes a failure seen on EL7 and Debian 10, assuming you'd prefer to ignore the warning rather than change the source.

### DIFF
--- a/test/build.bats
+++ b/test/build.bats
@@ -64,7 +64,7 @@ load common
     for i in "$ch_bin"/ch-*; do
         echo "shellcheck: ${i}"
         [[ ! $(file "$i") = *'shell script'* ]] && continue
-        shellcheck -e SC1090,SC2002,SC2154 "$i"
+        shellcheck -e SC1007,SC1090,SC2002,SC2154 "$i"
     done
     # libraries for user executables
     for i in "$ch_libexec"/*.sh; do


loveshack:
Actually, there are others that need fixing.  I'll amend when I have them passing.

loveshack:
I think what I'm seeing with shellcheck is just due to using an old version (0.3 on EL7), and there are quite a few changes to get it passing.
I can fix the issues with that, but suspect you'd rather ignore shellcheck if it's installed and too old.
Should I close this?

reidpr:
Yes, we would want to skip ShellCheck if it's too old. We don't have a documented minimum, but I've got 0.6.0 on my development box and I believe Travis is similar.

I looked at [SC1007](https://github.com/koalaman/shellcheck/wiki/SC1007) and it doesn't seem like one I'd want to ignore.

reidpr:
If you feel the issue is adequately covered by the bug I just created, then tag this PR `erroneous` and close it. Otherwise, let's discuss further.

loveshack:
> I looked at
> [SC1007](https://github.com/koalaman/shellcheck/wiki/SC1007) and it
> doesn't seem like one I'd want to ignore.

For what it's worth, I guess that more recent shellcheck spots the
difference between

x=  y

and

x=  # comment

which is why you weren't seeing the issue.

loveshack:
Closing as suggested.

PR #528, opened 2019-08-22
executables show help if no arguments

### LINKED ISSUE #503: executables should show help if run with no arguments
If you run ch-run-oci without arguments you get a python traceback instead of a help message:

```
ana@green:~> ch-run-oci 
  File "/usr/bin/ch-run-oci", line 365, in <module>
    main()
  File "/usr/bin/ch-run-oci", line 33, in main
    args = args_parse()
  File "/usr/bin/ch-run-oci", line 82, in args_parse
    bundle_ = bundle_from_cid(args_.cid)
AttributeError: 'Namespace' object has no attribute 'cid'
```
By the way, the command wrapped `ch-run` also shows an error  when it's run without arguments:

```
ana@green:~> ch-run
ch-run[19635]: NEWROOT and/or CMD not specified (ch-run.c:139)
``` 

It would be nice to get a short help message.

In case it's not evident, this is with version 0.10

rstyd:
This PR resolves #503. It makes slight modifications to ch-pull2[dir,tar] to achieve this goal.


test/build.bats
```diff
@@ -43,6 +43,11 @@ load common
         echo "$output"
         [[ $status -eq 0 ]]
         [[ $output = *'sage:'* ]]
+        # Executables print usage when run without arugments
+        run "$i"
+        echo "$output"
+        [[ $status -eq 0 ]]
+        [[ $output = *'sage:'* ]]
```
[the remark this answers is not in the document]
rstyd: For the first question, this behavior is because of some arcane GNU argparse administrivia. 
To print the full help text in the case of no arguments, one does:
https://github.com/hpc/charliecloud/blob/ee62ef321ece920e359c8f72b54681ee98e32e38/bin/ch-run.c#L442-L443

`ARGP_HELP_STD_HELP` prints the full help then exits with success. `ARGP_HELP_STD_USAGE` prints a message and exits with failure, but it only shows a partial message. All the other options outlined [here](https://www.gnu.org/software/libc/manual/html_node/Argp-Help-Flags.html) print a partial message.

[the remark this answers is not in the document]
rstyd: For the second, I based the test off of the --help test. The *sage catches `Usage` and `usage`. In this case, it's important since argparse.print_help which is used in all of the Python programs prints `usage` while everything else prints `Usage`. This could be changed.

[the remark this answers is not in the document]
j-ogas: Makes sense. Resolving.

bin/base.sh
```diff
@@ -27,6 +27,9 @@ builder_choose () {
 }
 
 parse_basic_args () {
+    if [ "$#" -eq 0 ]; then
+        usage 0
```
reidpr: I believe standard practice is to exit with failure if there is a usage error, which no arguments is, even if we are also printing help.

bin/ch-run.c
```diff
@@ -438,6 +439,9 @@ static error_t parse_opt(int key, char *arg, struct argp_state *state)
    case 'w':
       args->c.writable = true;
       break;
+   case ARGP_KEY_NO_ARGS:
+      argp_state_help(state, stderr, ARGP_HELP_STD_HELP);
```
reidpr: Looking briefly at the `argp_state_help` flags, I think we can get what we want (full help message, exit with failure), but it's not `ARGP_HELP_STD_HELP` by itself.

bin/ch-run.c
```diff
@@ -438,6 +439,9 @@ static error_t parse_opt(int key, char *arg, struct argp_state *state)
...
+
```
reidpr: Remove extra blank line.

bin/ch-checkns.c
```diff
@@ -52,6 +52,7 @@
 
 
 const char usage[] = "\
+\n\
```
reidpr: Why a newline added here?

bin/ch-checkns.c
```diff
@@ -74,9 +75,13 @@ void fatal_errno(int line)
...
 {
 
+   if (argc == 1) {
+      fprintf(stderr, usage);
+      exit(EXIT_SUCCESS);
+   }
    if (argc >= 2 && strcmp(argv[1], "--help") == 0) {
       fprintf(stderr, usage);
-      return 0;
+      exit(EXIT_SUCCESS);
```
reidpr: It's not clear to me if there's any difference, but regardless, can you justify this change?

bin/ch-grow
```diff
@@ -140,6 +140,11 @@ environment variables:
                    help="print version and exit")
    ap.add_argument("context", metavar="CONTEXT",
                    help="context directory")
+
+   if len(sys.argv) < 2:
```
reidpr: Use parentheses around condition per house style.

bin/ch-pull2dir
```diff
@@ -18,13 +18,8 @@ EOF
...
 
-if [ "$1" = "--help" ]; then
-    usage 0
-fi
-if [ "$1" = "--version" ]; then
-    version
-    exit 0
-fi
+parse_basic_args "$@"
+
```
reidpr: Good catch!

bin/ch-run.c
```diff
@@ -36,7 +36,8 @@ Example:\n\
   $ ch-run /data/foo -- echo hello\n\
   hello\n\
 \n\
-You cannot use this program to actually change your UID.";
+You cannot use this program to actually change your UID.\n\
+\n";
```
reidpr: Why newlines added?

bin/ch-checkns.c
rstyd: The newline was intended to make the message similar to the other which have a newline there as well. The message is quite large, and I think it improves readability.

bin/ch-run.c
rstyd: Also to look similar to the other messages.

bin/ch-checkns.c
```diff
@@ -74,9 +74,13 @@ void fatal_errno(int line)
...
 {
 
+   if (argc == 1) {
+      fprintf(stderr, usage);
+      return 1;
+   }
    if (argc >= 2 && strcmp(argv[1], "--help") == 0) {
       fprintf(stderr, usage);
-      return 0;
+      return 1;
```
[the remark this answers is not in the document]
rstyd: D'oh. That is correct

bin/ch-ssh.c
```diff
@@ -35,13 +36,17 @@ int main(int argc, char *argv[])
...
    char *args[ARGS_MAX+1];
 
+   if (argc == 1) {
+      fprintf(stderr, usage);
+      exit(EXIT_FAILURE);
+   }
    if (argc >= 2 && strcmp(argv[1], "--help") == 0) {
       fprintf(stderr, usage);
-      return 0;
+      exit(EXIT_SUCCESS);
```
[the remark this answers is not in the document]
rstyd: Some of the C programs used exit and some returned 0. This makes them all look the same and makes the code more self-documenting.

test/build.bats
```diff
@@ -43,6 +43,11 @@ load common
         echo "$output"
         [[ $status -eq 0 ]]
         [[ $output = *'sage:'* ]]
+        # Executables print usage when run without arugments
```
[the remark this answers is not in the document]
rstyd: I need to use `:set spell` in vim more often. :)

bin/ch-ssh.c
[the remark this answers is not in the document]
reidpr: OK.

We should probably use `0`/`1` vs. `EXIT_FAILURE`/`EXIT_SUCCESS` consistently, as well as `return` (which only works in `main()`) vs. `exit()`. I *believe* (correct me if I'm wrong) what we have now is `return 0` and `return 1` in `main()`, and `exit(0)` and `exit(1)` elsewhere. Let's leave it there for now, to remain consistent with the rest of the code.

I'd happily consider future issues to change `return` to `exit()`, as well as `0` and `1` to `EXIT_FAILURE` and `EXIT_SUCCESS`. They should be separate issues I think, and we should decide the outcome before beginning work.

Issue #543, opened 2019-09-12
ch-docker2squashfs: missing output directory should fail fast


sameershende:
If the output directory is missing, ch-docker2squashfs should complain immediately:

```
[sameer@gorgon bin]$ ./ch-docker2squash e4s_base:1.1 /tmp/ecp
builder: docker
exporting
adding environment
-rw-r--r--. 1 sameer paraducks 3.3G Sep 12 09:50 /tmp/ch-docker2squash.yH2tT1/e4s_base:1.1.tar
creating new image /tmp/ch-docker2squash.yH2tT1/e4s_base:1.1
/tmp/ch-docker2squash.yH2tT1/e4s_base:1.1 unpacked ok
can't squash: /tmp/ecp does not exist
```

Issue #542, opened 2019-09-12
ch-docker2squash does not work with Docker images that have a / in their name.


sameershende:
```
[sameer@gorgon bin]$ ./ch-docker2squash ecpe4s/ubi7_ppc64le_base:1.1 /tmp/ecp
builder: docker
exporting
adding environment
-rw-r--r--. 1 sameer paraducks 3.3G Sep 12 09:34 /tmp/ch-docker2squash.nL5t1e/ecpe4s.ubi7_ppc64le_base:1.1.tar
can't read: /tmp/ch-docker2squash.nL5t1e/ecpe4s/ubi7_ppc64le_base:1.1.tar
can't read: /tmp/ch-docker2squash.nL5t1e/ecpe4s/ubi7_ppc64le_base:1.1.tar.tar.gz
can't read: /tmp/ch-docker2squash.nL5t1e/ecpe4s/ubi7_ppc64le_base:1.1.tar.tar.xz
can't read: /tmp/ch-docker2squash.nL5t1e/ecpe4s/ubi7_ppc64le_base:1.1.tar.tgz
can't read: /tmp/ch-docker2squash.nL5t1e/ecpe4s/ubi7_ppc64le_base:1.1.tar.tar
no input found
```

PR #545, opened 2019-09-12
Builder2squash fixes

Brofessional:
fix issue where tar replaces slashes in tag names with periods
fix issue where ch-builder2tar does not complain about missing output directory before making temp files


Brofessional:
Adresses #542 and #543

bin/ch-builder2squash
```diff
@@ -36,18 +36,25 @@ if [ "$#" -lt 2 ]; then
 fi
 
 builder_choose
-echo "builder: ${CH_BUILDER}"
```
reidpr: What's the reasoning to remove this chatter?

bin/ch-builder2squash
```diff
@@ -36,18 +36,25 @@ if [ "$#" -lt 2 ]; then
...
 
 image=$1
 outdir=$2
 shift 2
 
+if [ ! -e "$outdir" ]; then
```
reidpr: I'd recommend `-d` here, to be more specific. You might then change the error message to something along the lines of "${outdir} is not a directory".

bin/ch-builder2squash
```diff
@@ -36,18 +36,25 @@ if [ "$#" -lt 2 ]; then
...
+    exit 1
+fi
+
 # mktemp is used for intermediate files to avoid heavy metadata loads on
 # certain filesystems
 temp=$(mktemp -d --tmpdir ch-builder2squash.XXXXXX)
 # Get image as a directory
 "${ch_bin}/ch-builder2tar" -b "$CH_BUILDER" --nocompress "$image" "$temp"
+
+image=$(echo "$image" | sed 's/\//./g')
```
reidpr: Is there any way to share the same naming code as wherever else does this substitution?

bin/ch-builder2squash
```diff
@@ -36,18 +36,25 @@ if [ "$#" -lt 2 ]; then
...
+
 "${ch_bin}/ch-tar2dir" "${temp}/${image}.tar" "$temp"
 # Create squashfs, and clean up intermediate files and folders.
 "${ch_bin}/ch-dir2squash" "${temp}/${image}" "$outdir" "$@"
-rm -rf --one-file-system "$temp"
```
reidpr: Leave the newline on the last line in the file.

bin/ch-builder2squash
Brofessional: Because `ch-builder2squash` calls `ch-builder2tar` which already has that line so you would get a parrot of the builder selected.

bin/ch-builder2squash
Brofessional: As in assign it the variable `tar` like in `ch-builder2tar`?

That is certainly possible. We would still need this call to identify the output of `ch-builder2tar`. So I could assign a variable to the tarball in addition to the image directory it will produce. 

Unless I'm wildly misunderstanding what you're asking me to do.

reidpr: What I'm thinking is that somewhere we already have code that translates a Docker image tag to something suitable for a filesystem, so I propose find that code and put it in a function in `base.sh`.

PR #546, opened 2019-09-12
HSN bandwidth test: Don't run on single node; clean up

reidpr:
The HSN bandwidth test `mpibench-openmpi/using the high-speed network (host launch)` was supposed to be skipped in single-node environments, but it wasn't. Further, it seemed to busy-hang on my development box.

This PR fixes the test for a multi-node environment.

In terms of clean up, I:

* removed `common.bash:disable_hsn_openmpi` because it's only used in one place
* moved the test for `/dev/infiniband` directory to the top of the test
* made the test fail (not skip) if above directory not present
* copy-edited some of the comments

Full scope test passes on my dev box, but I haven't tested it in a Slurm allocation. Could one of you do this, please?


examples/mpi/mpibench/test.bats
```diff
@@ -83,18 +80,22 @@ check_process_ct () {
...
+# plugins enabled against the performance just using TCP. Pass if HSN
+# performance is at least double TCP.
 @test "${ch_tag}/using the high-speed network (host launch)" {
     multiprocess_ok
-    [[ $SLURM_NNODES = "1" ]] && skip "Multinode only"
+    [[ $ch_multinode ]] || skip "multinode only"
     [[ $ch_cray ]] && skip "Cray doesn't support running on tcp"
+    # Verify we have proper HSN devices present. (Note that -d tests for
+    # directory, not device.)
+    [[ -d /dev/infiniband ]]
```
heasterday: This will cause the test to fail on Ethernet only clusters e.g. woodchuck, do we plan to not test on such systems?

reidpr: In principle an Ethernet system could be using RDMA, though we do not have any such systems.

Our Ethernet clusters use TCP, making the test a no-op. We'd want to skip in that case, I think.

heasterday: Apologies if I'm confused, in this PR it would fail on the [[ -d /dev/infiniband ]] check and in  master after the first unnecessary run of the benchmark it skips in disable_hsn_openmpi(). Which case would it fail because performance would be equal?

reidpr: You're right. It would also fail in performance if we modified the line in question to let Ethernet systems through.

heasterday: Ahh I see what you are saying, I wasn't suggesting removing the line, I was suggesting adding the skip back if the directory didn't exist. I wanted to see if this was intended behavior before suggesting it though.

reidpr: The reason I changed it from skip to fail is that I wanted to make an explicit choice about systems that have an HSN but it's not IB or OPA. But I can't think of a way to test whether there is an HSN present but it's Ethernet, since testing for Ethernet will match anything. Do you have any ideas?

examples/mpi/mpibench/test.bats
```diff
@@ -83,18 +80,22 @@ check_process_ct () {
...
 # This test compares OpenMPI's point to point bandwidth with all high speed
-# plugins enabled against the performance just using tcp. If the performance of
-# the former is not at least double that of tcp the former's performance is
-# assumed to be lacking.
+# plugins enabled against the performance just using TCP. Pass if HSN
+# performance is at least double TCP.
 @test "${ch_tag}/using the high-speed network (host launch)" {
     multiprocess_ok
-    [[ $SLURM_NNODES = "1" ]] && skip "Multinode only"
+    [[ $ch_multinode ]] || skip "multinode only"
```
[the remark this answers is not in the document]
reidpr: Hmm, that seems like a bug. I wonder if we should refuse to run in a Slurm allocation smaller than two nodes.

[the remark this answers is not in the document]
heasterday: I agree it's a logic bug. That was the thought behind the original check of SLURM_NNODES = 1, perhaps this check could be used in common.bash where ch_multinode is set to only set ch_multinode when in Slurm allocations greater than one node.

examples/mpi/mpibench/test.bats
heasterday: Makes sense. On the initial commit I only thought as far as RoCE which I believe also uses the device file /dev/infiniband/uverbsX so it would have passed the check for /dev/infiniband. My knowledge of Ethernet only system is quite lacking though so I don't have any ideas. If you have some example systems in mind I can look into it though.

test/common.bash
```diff
@@ -296,19 +299,25 @@ ch_cores_total=$((ch_nodes * ch_cores_node))
...
     fi
+    if [[ $ch_nodes -eq 1 ]]; then
+        ch_multinode=
+        ch_multiprocess=
+        ch_mpirun_2=false
+    else
+        ch_multinode=yes
+        ch_multiprocess=yes
+        ch_mpirun_2='srun -n2'
+    fi
```
heasterday: The downside of setting `ch_multiprocess=` here is this means now a minimum of 2 nodes are required to run any of the host launch tests. Even though I would recommend folks use a guest launch in single node allocations I do think it is beneficial to test host launch. 

In general I feel if we detect Slurm env variables we should be able to run multiple processes. We could also still run `srun -n2`. If we want `srun -n2` to truly be only used for 2 ranks on 2 seperate nodes any tests using it would need to check `ch_multinode` and we may want to change it to `srun -N2 -n2` to be more explicit about what we are doing.

reidpr: I think `ch_multiprocess=` is a typo and I meant `ch_multiprocess=yes`.

reidpr: Can you make that fix as you test?

heasterday:
@reidpr Made the changes we discussed and it looks good when I tested two nodes with slurm, one node with slurm, one node with mpirun.

heasterday:
Incorporated what we discussed at the meeting today. We may want more explanation about the skip vs fail decision.

Issue #553, opened 2019-09-24
remove embedded Bats


reidpr:
We've been discussing simplifying the test suite by removing the embedded Bats.

Currently, `make install` does not install the embedded Bats (`test/Makefile:134` has an extra `test/`). Thus, the Vagrantfile cannot pass tests, because the system Bats doesn't work (#552) and the embedded Bats isn't installed.

Rather than fix the install, why don't we go ahead and just remove the embedded Bats.

@j-ogas, opinions?

Issue #552, opened 2019-09-24
tests fail with Bats 0.4.0, which is latest in EPEL


reidpr:
The test suite fails with Bats version 0.4.0, which is what's currently in EPEL (sstephenson/bats@3b33a5a):

```
 ✗ broken image errors
   (in test file run/ch-run_misc.bats, line 582)
     `rm "${img}/${f}"' failed
   ch-run[19016]: can't execve(2): /bin/true: No such file or directory (charliecloud.c:476 2)
   required: etc/group etc/passwd
   rm: cannot remove '/tmp/bats.tmp.charlie/broken-image/etc/group etc/passwd': No such file or directory
```

The problem is that Bats' `run` command does not restore `$IFS`, so word breaking no longer works after you call it. This was fixed in PR sstephenson/bats#90, but the EPEL package does not have that PR.

The embedded Bats via our Git submodule works because it's Git master and does have the PR, but we want to get rid of embedded Bats.

We want the test suite to pass using readily available Bats, which it currently does not on RHEL/CentOS/Fedora.

I see two options:

1. [Wrap](https://stackoverflow.com/a/32425874) Bats `run` to restore `$IFS`. This would make EPEL Bats work, but adds cruft to our code.
2. Require a Bats with the bug fix in question. This would have less cruft but require:
    * Update the Bats RPM. Currently it is unmaintained, and may be [replaced](https://src.fedoraproject.org/rpms/bats) with Bats-core, which does seem to have the bug fix. (Bats is no longer maintained upstream, and Bats-core is a maintained fork.)
    * Test that the Bats in use does not have the bug.
    * Document the required Bats version, which is more complicated than stating a simple version number.

reidpr:
Note that Debian Stretch (9) and Buster (10) both have the buggy Bats as well, and it seems Debian Bullseye (11) will have Bats-core.

PR #556, opened 2019-09-25
remove Bats submodule and work around Bats bug

reidpr:
Addresses #552 and #553.

### DIFF
--- a/test/travis.yml
+++ b/test/travis.yml
@@ -18,10 +18,9 @@ compiler: gcc
 # it up in parallel. It would be nice if "make test-build" could be done
 # serially before splitting into parallel jobs.
 #
-#   TARBALL=              # build in Git checkout & use embedded Bats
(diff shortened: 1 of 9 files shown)


test/travis.yml
```diff
@@ -72,12 +71,12 @@ jobs:
       env: CH_BUILDER=buildah         TARBALL=export       INSTALL=yes
     - <<: *stage_install
       env: CH_BUILDER=buildah-runc    TARBALL=export       INSTALL=yes
+    - <<: *stage_install
+      env: CH_BUILDER=buildah-setuid  TARBALL=export       INSTALL=yes
```
j-ogas: Is this related to removing the bats submodule? If not, perhaps we should make a separate issue, which we can also close with this PR. Thoughts?

reidpr: Probably? I removed the `TARBALL=export-bats` one, which freed up a space in the stage, so I added this one.

Issue #559, opened 2019-09-26
ch-grow should list all missing dependencies


j-ogas:
Related to #447.

`ch-grow` should have a `--dependencies` flag that will check the for the presence of each dependency, e.g., python3, lark-parser, skopeo, and umoci, and report all missing dependencies.

Issue #550, opened 2019-09-23
ch-grow --print-storage requires a context directory


j-ogas:
I don't think a context directory should be needed to get the `CH_GROW_STORAGE` variable.

```
$ ch-grow --print-storage
usage: ch-grow [-h] [--build-arg KEY=VALUE] [-f DOCKERFILE] [-n] [--no-cache]
               [--parse-only] [--print-storage] [-s DIR] [-t TAG] [--verbose]
               [--version]
               CONTEXT
ch-grow: error: the following arguments are required: CONTEXT
```

PR #560, opened 2019-09-26
ch-grow: new options --dependencies, clean up --print-storage

reidpr:
This addresses #550 and #559.

I have not comprehensively tested all combinations of missing or broken dependency.

### DIFF
--- a/bin/ch-grow
+++ b/bin/ch-grow
@@ -40,7 +40,7 @@ try:
    lark.Visitor  # AttributeError if wrong module
 except (ImportError, AttributeError) as x:
    if (isinstance(x, ImportError)):
-      depfails.append(("missing", 'Python module "lark"'))
+      depfails.append(("missing", 'Python module "lark-parser"'))
    elif (isinstance(x, AttributeError)):
       depfails.append(("bad", 'found Python module "lark"; need "lark-parser"'))
    else:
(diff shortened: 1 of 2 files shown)


bin/ch-grow
```diff
@@ -584,6 +596,20 @@ def cp(src, dst):
    else:
       shutil.copy2(src, dst)
 
+def dependencies_check():
+   # Check some more dependencies. If any dependency problems found, here or
+   # above (e.g., lark module checked at import time), the complain and exit.
```
j-ogas: [...], "then" complain [...]

bin/ch-grow
```diff
@@ -584,6 +596,20 @@ def cp(src, dst):
...
+      depfails.append(("missing", CH_RUN))
+   if (not shutil.which("skopeo")):
+      depfails.append(("missing", "skopeo"))
+   if (not shutil.which("umoci")):
+      depfails.append(("missing", "umoci"))
+   for (p, v) in depfails:
+      ERROR("%s dependency: %s" % (p, v))
+   if (len(depfails) > 0):
+      sys.exit(1)
+
```
j-ogas: Should we be enforcing some minimum version for these dependencies?

reidpr: Yes, but I don't know what they are for the moment.

PR #139, opened 2018-03-30
add ParaView to test suite

reidpr:
This addresses issue #125.

PR #564, opened 2019-10-03
Update LAMMPS image and test mpi4py with LAMMPS example.

heasterday:
Related to internal issue #139. I've tested this multinode on an ARM system, Two CTS-1 systems, and our Cray XC40 but folks giving it a run on any further environments of interest would be appreciated. Please note that you will want to refresh your OpenMPI base image to include PR #561 as it won't happily run without it.

This could use a tidy


examples/mpi/lammps/Dockerfile
```diff
@@ -1,29 +1,41 @@
...
-WORKDIR /
+WORKDIR /usr/local/src
 
 # Packages for building.
 RUN apt-get install -qy --no-install-recommends \
+    cmake \
     git \
     patch \
-    python-dev
+    python-dev \
```
reidpr: Can we use Python 3? (`python3-foo` and then `pip3` later)

examples/mpi/lammps/Dockerfile
```diff
@@ -1,29 +1,41 @@
...
+    python-pip \
+    python-setuptools 
 
+# Building mpi4py from source to ensure it is built against our MPI build
+# Building numpy from source to work around issues seen on Aarch64 systems
+RUN pip install --no-binary=mpi4py,numpy mpi4py numpy
```
reidpr: I think you can say `--no-binary=:all:`? Or are there dependencies where you'll accept binary packages? If the latter, say that briefly.

examples/mpi/lammps/Dockerfile
```diff
@@ -1,29 +1,41 @@
...
-# Patch instead of including in.large.melt in full because LAMMPS is GPL.
-COPY melt.patch /lammps/examples/melt
-RUN patch -p1 -o /lammps/examples/melt/in.large.melt \
-               < /lammps/examples/melt/melt.patch
+ENV LAMMPS_VERSION 19Sep19
+RUN wget -nv http://lammps.sandia.gov/tars/lammps-$LAMMPS_VERSION.tar.gz
+RUN tar xf lammps-$LAMMPS_VERSION.tar.gz
+RUN mkdir lammps-${LAMMPS_VERSION}.build
+RUN    cd lammps-${LAMMPS_VERSION}.build \
+    && cmake -DCMAKE_INSTALL_PREFIX=/usr \
```
reidpr: Inconsistent with `/usr/local/src` above.

examples/mpi/lammps/Dockerfile
```diff
@@ -1,29 +1,41 @@
...
+             -DPKG_DIPOLE=yes \
+             -DPKG_KSPACE=yes \
+             -DPKG_POEMS=yes \
+             -DPKG_PYTHON=yes \
+             -DPKG_USER-REAXC=yes \
+             -DPKG_USER-MEAMC=yes \
+             -DLAMMPS_MACHINE=mpi \
+          ../lammps-${LAMMPS_VERSION}/cmake \
+    && make -j $(getconf _NPROCESSORS_ONLN) install
+RUN ln -s /usr/local/src/lammps-${LAMMPS_VERSION}/ /lammps
```
reidpr: Do we need a top-level directory for it?

examples/mpi/lammps/Dockerfile
```diff
@@ -1,29 +1,41 @@
...
+# Patch in.melt instead of including in.large.melt because LAMMPS is GPL.
+# Patch simple.py to uncomment mpi4py calls and disable file output.
+COPY melt-and-simple.patch /lammps
+RUN patch -d / -p1 < /lammps/melt-and-simple.patch
```
reidpr: I'd suggest keeping two patches for the separate targets.

examples/mpi/lammps/test.bats
```diff
@@ -79,15 +79,20 @@ lammps_try () {
...
-# This test busy-hangs after several:
+# Previously running the python example in /examples/python resulted in a busy
+# hang with the error:
 #
 #   FOO error: local variable 'foo' referenced before assignment
 #   Inside simple function
 #
-# Perhaps related to --join?
+# Using mpi4py to launch the examples in /python/examples does not appear to
+# exhibit this behavior so using simple.py instead
```
reidpr: I think you can probably skip explaining the previous behavior.

heasterday: I'll remove.

examples/mpi/lammps/Dockerfile
heasterday: I'll change to python3 and retest.

examples/mpi/lammps/Dockerfile
heasterday: I'll change to `--no-binary :all: mpi4py numpy` as I'm not seeing a behavior difference between the two.

examples/mpi/lammps/Dockerfile
heasterday: Good catch, I'll change to `&& cmake -DCMAKE_INSTALL_PREFIX=/usr/local \`

examples/mpi/lammps/Dockerfile
heasterday: I will change to python3 and retest

examples/mpi/lammps/Dockerfile
heasterday: The top level directory `/lammps` is carried over behavior from the previous iteration. It make working with the path quite a bit cleaner for the relevant tests both by being shorter and abstracting the particular lammps version away (other solutions to second problem exist). This behavior is inconsistent among the tests though. If you would like to do away with the `/lammps` symlink I can adjust the relevants tests.

examples/mpi/lammps/Dockerfile
heasterday: Since there are three targets, in.melt, in.simple, and simple.py are you thinking a patch for the former and one patch for the second two? Melt.patch could be restored to what is in git master and a second could be created that patches in.simple and simple.py. A downside of this would be that the patches would be inconsistently doing similar things. The melt.patch creates a new file for its result and simple.patch would be an in-place patch.

heasterday: Per the meeting today we decided two separate patches, both in-place.

examples/mpi/lammps/Dockerfile
heasterday: This required moving the install target to /usr/ as the resulting python3 lammps module wasn't being found. This is consistent with the direction we plan to go with the CentOS 8 base images .

heasterday:
@reidpr Made some changes required for python3 and incorporated your feedback. 

Tested with python3 on two seperate CTS platforms let me know if you have other systems you want it tested on.

Issue #558, opened 2019-09-26
Tidy ch-checkns


j-ogas:
`ch-checkns` needs to re-mount read-only using existing flags (see: issue #9, PR #407).

`ch-checkns` should also provide better output when failing (perhaps a URL to the file, which has excellent documentation).

j-ogas:
Related to PR #447.

PR #567, opened 2019-10-10
update ch-checkns

j-ogas:
Addresses #558.

### DIFF
--- a/bin/ch-checkns.c
+++ b/bin/ch-checkns.c
@@ -62,12 +62,14 @@ Example:\n\
   ok\n\
 \n";
 
-#define TRY(x) if (x) fatal_errno(__LINE__)
+#define TRY(x) if (x) fatal_errno(__LINE__, #x)
 
 
-void fatal_errno(int line)
+void fatal_errno(int line, char *str)
 {
    printf("error at line %d, errno=%d\n", line, errno);
+   printf("%d: if(%s)\n", line, str);
+   printf("see: https://github.com/hpc/charliecloud/blob/master/bin/ch-checkns.c\n");
    exit(1);
 }
 
@@ -132,7 +134,10 @@ int main(int argc, char *argv[])
       TRY (errno);
 
    /* Re-mount the image read-only. */
-   TRY (mount(NULL, "/newroot", NULL, MS_REMOUNT | MS_BIND | MS_RDONLY, NULL));
+   unsigned long flags = path_mount_flags("/newroot") | MS_REMOUNT
+                                                      | MS_BIND
+                                                      | MS_RDONLY;
+   TRY (mount(NULL, "/newroot", NULL, flags, NULL));
 
    /* Finally, make our "real" newroot into the root filesystem. */
    TRY (chdir("/newroot"));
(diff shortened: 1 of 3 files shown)


bin/ch-checkns.c
```diff
@@ -132,7 +134,10 @@ int main(int argc, char *argv[])
       TRY (errno);
 
    /* Re-mount the image read-only. */
-   TRY (mount(NULL, "/newroot", NULL, MS_REMOUNT | MS_BIND | MS_RDONLY, NULL));
+   unsigned long flags = path_mount_flags("/newroot") | MS_REMOUNT
+                                                      | MS_BIND
+                                                      | MS_RDONLY;
+   TRY (mount(NULL, "/newroot", NULL, flags, NULL));
```
reidpr: Does this fit on one line? If not, I'd recommend wrapping as follows:

```
unsigned long flags = (  path_mount_flags("/newroot")
                       | MS_REMOUNT | MS_BIND | MS_RDONLY);
```

bin/charliecloud.h
```diff
@@ -90,6 +90,7 @@ extern int verbose;
 
 void containerize(struct container *c);
 void msg(int level, char *file, int line, int errno_, char *fmt, ...);
+unsigned long path_mount_flags(char *path);
```
reidpr: If we're going to use stuff from `charliecloud.c`, which it looks like we were already doing, I think we should use everything available. I believe there's a lot of duplication in `ch-checkns.c`.

bin/ch-checkns.c
```diff
@@ -62,12 +61,14 @@ Example:\n\
   ok\n\
 \n";
 
-#define TRY(x) if (x) fatal_errno(__LINE__)
+#define TRY(x) if (!(x)) fatal_(__FILE__, __LINE__, errno, #x)
```
reidpr: Why reverse the sense of this macro? It seems to make the calls noisier.

bin/ch-checkns.c
```diff
@@ -106,41 +107,42 @@ int main(int argc, char *argv[])
...
-   TRY (chroot("."));
+   TRY (0 == chroot("."));
 
    /* Make a place for the old (intermediate) root filesystem to land. */
    if (mkdir("/newroot/oldroot", 0755) && errno != EEXIST)
       TRY (errno);
 
    /* Re-mount the image read-only. */
-   TRY (mount(NULL, "/newroot", NULL, MS_REMOUNT | MS_BIND | MS_RDONLY, NULL));
+   unsigned long flags = path_mount_flags("/newroot") | MS_REMOUNT | MS_BIND | MS_RDONLY;
```
reidpr: I prefer to put variable declarations at the beginning of blocks, because (1) then any variable is either in- or out-of-scope for the whole scope and (2) I can never remember if C makes the variable in scope for the whole block or just from declaration to the end.

bin/ch-checkns.c
j-ogas: I initially forgot that C's `if(0)` is `false`. I wanted the fatal function to execute when the syscalls return failure (non-zero). 

I personally think `TRY (0 == syscall())` is clearer than `TRY (!syscall())`

reidpr:
@j-ogas, if you're happy, I'll merge.

Issue #568, opened 2019-10-14
buildah 1.11.2 errors with ch-run-oci


j-ogas:
Related to #447 #517.

`ch-run-oci` has /tmp hard coded as the path to check for bundle arguments:

```bash
 97 def bundle_from_cid(cid):
 98    m = re.search(r"^buildah-buildah(.+)$", cid)
 99    if (m is None):
100       fatal("cannot parse container ID: %s" % cid)
101    return "/tmp/buildah%s" % m.groups(1
```

Buildah 1.11.2 appears to use `/var/tmp`, which results in the following error:
```
error running container: error creating container for [/bin/sh -c apk add --no-cache bc]: ch-run-oci[27473]: bundle argument "/var/tmp/buildah909090388" differs from inferred bundle "/tmp/buildah909090388"
```

PR #517, opened 2019-08-16
add ch-grow dependencies to Vagrantfile

j-ogas:
Addresses #493.

PR #570, opened 2019-10-17
ch-run-oci: check multiple paths for inferred bundles

j-ogas:
Addresses #568; needed for #447 and #517.

### DIFF
--- a/bin/ch-run-oci
+++ b/bin/ch-run-oci
@@ -13,6 +13,7 @@ import sys
 import time
 import types
 
+BUNDLE_PREFIX = ["/tmp", "/var/tmp"]
 CH_BIN = os.path.dirname(os.path.abspath(
            inspect.getframeinfo(inspect.currentframe()).filename))
 OCI_VERSION_MIN = "1.0.1-dev"
@@ -79,15 +80,15 @@ def args_parse():
       cmd = CH_BIN + "/ch-run"
       os.execl(cmd, cmd, "--version")
 
-   bundle_ = bundle_from_cid(args_.cid, ["/tmp", "/var/tmp"])
+   bundle_ = bundle_from_cid(args_.cid, BUNDLE_PREFIX)
    if ("bundle" in args_ and args_.bundle != bundle_):
-      fatal("bundle argument \"%s\" differs from available inferred bundle \
-             paths: \"%s\"" % (args_.bundle, bundle_))
+      fatal("bundle argument \"%s\" differs from inferred bundle \"%s\""
+            % (args_.bundle, bundle_))
    args_.bundle = bundle_
 
    pid_file_ = pid_file_from_bundle(args_.bundle)
    if ("pid_file" in args_ and args_.pid_file != pid_file_):
-      fatal("pid_file argument \"%s\" differs from inferred \"%s\"" \
+      fatal("pid_file argument \"%s\" differs from inferred \"%s\""
             % (args_.pid_file, pid_file_))
    args_.pid_file = pid_file_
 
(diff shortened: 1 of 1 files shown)


bin/ch-run-oci
```diff
@@ -79,27 +79,29 @@ def args_parse():
       cmd = CH_BIN + "/ch-run"
       os.execl(cmd, cmd, "--version")
 
-   bundle_ = bundle_from_cid(args_.cid)
+   bundle_ = bundle_from_cid(args_.cid, ["/tmp", "/var/tmp"])
```
reidpr: I don't believe this is the right place to define the list. Set a constant at the top and use it in `bundle_from_cid()`.

bin/ch-run-oci
```diff
@@ -79,27 +79,29 @@ def args_parse():
...
    if ("bundle" in args_ and args_.bundle != bundle_):
-      fatal("bundle argument \"%s\" differs from inferred bundle \"%s\""
-            % (args_.bundle, bundle_))
+      fatal("bundle argument \"%s\" differs from available inferred bundle \
+             paths: \"%s\"" % (args_.bundle, bundle_))
```
reidpr: Revert? This looks left over from when the list was passed around.

bin/ch-run-oci
```diff
@@ -79,27 +79,29 @@ def args_parse():
...
    args_.bundle = bundle_
 
    pid_file_ = pid_file_from_bundle(args_.bundle)
    if ("pid_file" in args_ and args_.pid_file != pid_file_):
-      fatal("pid_file argument \"%s\"% differs from inferred \"%s\""
+      fatal("pid_file argument \"%s\" differs from inferred \"%s\"" \
```
reidpr: This removes an erroneous `%`? Good, but the trailing ` \` is not needed because it's in parentheses.

bin/ch-run-oci
```diff
@@ -79,27 +79,29 @@ def args_parse():
...
-
-def bundle_from_cid(cid):
+def bundle_from_cid(cid, prefix):
    m = re.search(r"^buildah-buildah(.+)$", cid)
    if (m is None):
       fatal("cannot parse container ID: %s" % cid)
-   return "/tmp/buildah%s" % m.groups(1)
-
+   for p in prefix:
+      p = p + "/buildah%s" % m.group(1)
```
reidpr: 1. Don't re-use `p` here.
2. I suggest a single way of string interpolation on this line rather than two (concatenation, printf-style).

bin/ch-run-oci
```diff
@@ -79,27 +79,29 @@ def args_parse():
...
+      if os.path.exists(p):
+         return p
+   fatal("bundle doesn't match an inferred path (%s))" % list(prefix))
```
reidpr: 1. I think because we're being so weird here, it's important to list all the paths that we tried, not just the prefixes.
2. Use colon style, e.g.:

```
fatal("can't infer bundle path; none of these exist: %s" % " ".join(paths)
```

j-ogas:
Not sure which version of python3 Travis is using but I can't replicate the travis error with python 3.7 or 3.6:

```
$ ./bin/ch-run-oci --version
0.12~pre+chrunoci.8339665
$ git rev-parse HEAD
833966516def50f0e62619e1a372700f370808b2
$ python3 --version
Python 3.7.2
$ conda deactivate
$ python3 --version
Python 3.6.8
$ ./bin/ch-run-oci --version
0.12~pre+chrunoci.8339665
```

bin/ch-run-oci
```diff
@@ -79,27 +80,30 @@ def args_parse():
       cmd = CH_BIN + "/ch-run"
       os.execl(cmd, cmd, "--version")
 
-   bundle_ = bundle_from_cid(args_.cid)
+   bundle_ = bundle_from_cid(args_.cid, BUNDLE_PREFIX)
```
reidpr: Just use the `BUNDLE_PREFIX` constant directly in `bundle_from_cid()`, without adding an argument (unless I'm missing something).

Issue #581, opened 2019-11-08
remove test/bin symlink


reidpr:
`common.bash` uses `test/bin`, a symlink to `bin`, to find the relevant executables. This functionality should be moved to `ch-test` and the symlink removed.

ID credit: @rstyd

Issue #576, opened 2019-10-30
delete test/{Makefile,make-auto}


reidpr:
Concerns:

* `Makefile` has a lot of procedural code for running tests, and it's ugly.
* `make-auto` is a Python program that doesn't do much any more and is the only reason (a) building requires Python and (b) the tests need to be built at all.
* Tests are slow because skipping is expensive (#186).

As a follow-on to #447, let's consider removing both of these files and folding their functionality into `ch-test`. I think procedure might be:

* Create a temporary directory. May as well use it at `$BATS_TMPDIR`, in which case it needs to be under `/tmp` because of #278. 
* Copy all the `.bats` files to this directory, but filter them to remove out-of-scope tests. (Note there are filename collisions.)
* Generate the derived `.bats` files in the same directory. I believe all the variables could be substituted with `sed` or maybe `awk`.
* Run the tests; working directory would need to remain `test.
* Remove temporary directory.

We might want a `--keep-tmp` to leave this stuff around for examination.

PR #584, opened 2019-11-14
update ch-test with test/Makefile and test/make-auto logic

j-ogas:
Addresses #576 and #581.

Notes:
- This PR does not completely remove `test/Makefile`


bin/ch-test
```diff
@@ -169,12 +320,12 @@ win () {
     printf "\nAll tests passed.\n"
 }
 
-
-top_dir=$(readlink -f "${ch_bin%/*}")
+bin=$(readlink -f "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)")
+topdir=$(readlink -f "${bin%/*}")
```
reidpr: (1) I would set these up at the top of the script, since that's where the others do it.
(2) Recommend rename `topdir` to `ch_base` for consistency (or even add it to `base.sh`).
(3) Why `readlink`?

bin/ch-test
```diff
@@ -186,18 +337,35 @@ fi
 
 # Ensure BATS_TMPDIR is set to /tmp (issue #278).
 if [[ -n $BATS_TMPDIR ]]; then
-    if [[ ! $BATS_TMPDIR == '/tmp/'* ]]; then
+    if [[ "$BATS_TMPDIR" != '/tmp'* ]]; then
```
reidpr: (1) You can skip the double quotes because it's Bash `[[ ]]`.
(2) This tests if `$BATS_TMPDIR` *begins with* `/tmp` rather than *is equal to* `/tmp`.

bin/ch-test
```diff
@@ -186,18 +337,35 @@ fi
...
         fatal "BATS_TMPDIR: must be /tmp but found $BATS_TMPDIR (issue #278)"
     fi
 fi
 
+# Create a directory in user space to hold auto-generated test artifacts
```
reidpr: (1) Say "home directory" rather than "user space", which is a kernel concept.
(2) I think this belongs better in `$BATS_TMPDIR` rather than the home directory.
(3) What's the reasoning to keep the directory rather than deleting it?
(4) Note that `$BATS_TMPDIR` gets twiddled in `common.bash`. I wonder if that logic is better here?
(5) You'll want to make sure you have a fresh directory without any leftovers from a previous run.

bin/ch-test
```diff
@@ -186,18 +337,35 @@ fi
...
+tmpdir=${HOME}/.charliecloud/ch-test.tmp.$USER
```
reidpr: (1) Call it something more specific than `tmpdir`.
(2) Generally speaking, I recommend avoiding dot files, as they hide filesystem pollution. Principle: "[explicit is better than implicit](https://inventwithpython.com/blog/2018/08/17/the-zen-of-python-explained/)".

bin/ch-test
```diff
@@ -186,18 +337,35 @@ fi
...
+if [[ ! -d $tmpdir ]]; then
+    mkdir -p "$tmpdir"
+    chmod 700 "$tmpdir"
+    chmod 700 "$(dirname "$tmpdir")"
```
reidpr: (1) Don't mess with `$tmpdir`'s parent.
(2) Don't use `-p`, since you know it doesn't exist.
(3) I think the `mkdir`/`chmod` combination sets up a race condition and maybe other issues. Use `-m 700` instead.

bin/ch-test
```diff
@@ -186,18 +337,35 @@ fi
...
+fi
+
+if [[ $(stat -c %a "$tmpdir") != '700' ]]; then
+    fatal "BATS_TMPDIR: $tmpdir doesn't have octal perms 700"
+fi
+
+if [[ $(stat -c %a "$(dirname "$tmpdir")") != '700' ]]; then
+    fatal "BATS_TMPDIR: $(dirname "$tmpdir") doesn't have  octal perms 700"
+fi
```
reidpr: I think you do not need to check this for the case where you created the directory. But, can you reduce the number of code paths to always create it?

bin/ch-test
```diff
@@ -186,18 +337,35 @@ fi
...
+
+
 # Determine source vs prefix installation.
```
reidpr: I think better jargon is "build directory" and "installed".

bin/ch-test
```diff
@@ -247,6 +415,7 @@ while [[ $# -gt 0 ]]; do
         require_unset scope
         scope_check "$1"
         scope=$1; shift
+        scope_value=$(scope_to_digit "$scope")
```
reidpr: `_value` isn't good because both are values. `scope_int`?

bin/ch-test
```diff
@@ -292,9 +461,13 @@ if [[ $phase == *'perm'* ]] && [[ ${CH_TEST_PERMDIRS[*]} == skip ]]; then
     fatal "phase $phase: CH_TEST_PERMDIRS: can't be 'skip'"
 fi
 
+if [[ -z $scope_value ]]; then
+    scope_value=$(scope_to_digit "$CH_TEST_SCOPE")
+fi
+
```
reidpr: Can you reduce the number of code paths to remove this conditional? I think you do not need to set `scope_value` during argument parsing.

bin/ch-test
```diff
@@ -364,5 +542,9 @@ case $phase in
         ;;
     clean)
         dirs_unpriv_rm
+        if [[ -d $tmpdir ]]; then
+            echo "removing $tmpdir"
+            rm -rf --one-file-system "$tmpdir"
```
reidpr: If you're going to `rm -rf`, make it safer by checking for a sentinel.

doc-src/ch-test_desc.rst
```diff
@@ -67,7 +67,8 @@ allows trading off thoroughness versus time.
     :code:`--perm-dirs`.
 
   :code:`clean`
-    Delete packed and unpacked images directories.
+    Delete automatically-generated test files, packed and unpacked images
+    directories.
```
reidpr: Add Oxford comma.

test/chtest-make-auto
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+
```
reidpr: No `base.sh`?

test/chtest-make-auto
```diff
@@ -0,0 +1,102 @@
...
+fatal () {
+    printf "error: %s\n" "$1" >&2
+    exit 1
+}
+
+tag_from_path () {
+    parent_dir=$(dirname "$path_")
+    basicname=${basename_%%.*}
```
reidpr: Call it `basename_`?

test/chtest-make-auto
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+
+fatal () {
+    printf "error: %s\n" "$1" >&2
+    exit 1
+}
+
+tag_from_path () {
+    parent_dir=$(dirname "$path_")
```
reidpr: I would call this `dir` because `parent` makes me ask "parent of what". I think only directories have parents. Files have enclosing or containing directories.

test/chtest-make-auto
```diff
@@ -0,0 +1,102 @@
...
+        extension=''
+    else
+        extension=${extension/\./} # remove dot
+    fi
+
+    case $basicname in
+    Build|Dockerfile|Docker_Pull)
+        if [[ $parent_dir == '' ]]    || \
+           [[ $parent_dir == '.' ]]   || \
+           [[ $parent_dir == test ]]; then
```
reidpr: Format as:

```
if    [[ $parent_dir == '' ]] \
   || [[ $parent_dir == '.' ]] \
   || [[ $parent_dir == test ]]; then
```

Principle: operators are more visible at beginning of line.

test/chtest-make-auto
```diff
@@ -0,0 +1,102 @@
...
+                echo "$extension"
+            fi
+        else
+            if [[ -z $extension ]]; then
+                echo $(basename "$parent_dir")
+            else
+                echo $(basename "${parent_dir}-${extension}")
+            fi
+        fi
+    ;;
```
reidpr: Indent.

test/chtest-make-auto
```diff
@@ -0,0 +1,102 @@
...
+
+template_print () {
+    template="./make-auto.d/$1"
+
+    cat "$template" | sed "s@%(addr)s@$addr@g" \
+                    | sed "s@%(basename)s@$basename_@g" \
+                    | sed "s@%(dirname)s@$dirname_@g" \
+                    | sed "s@%(path)s@$path_@g" \
+                    | sed "s@%(scope)s@$scope@g" \
+                    | sed "s@%(tag)s@$tag@g"
```
reidpr: I think you can reduce this to a single `sed` or `awk` invocation, and remove `cat`.

test/chtest-make-auto
```diff
@@ -0,0 +1,102 @@
...
+}
+
+mode=$1; shift
+scope=$CH_TEST_SCOPE
+
+if [[ -z $CH_TEST_SCOPE ]]; then
+    fatal "CH_TEST_SCOPE not set"
+fi
+
+cd ../test
```
reidpr: Make the CWD assumption explicit, or eliminate it.

test/chtest-make-auto
```diff
@@ -0,0 +1,102 @@
...
+if [[ $mode != tag ]]; then
+    printf "# Do not edit this file; it's autogenerated\n\n"
+    printf "load %s/common.bash\n\n" $(dirname "${BASH_SOURCE[0]}")
+fi
+
+while [[ "$#" -gt 0 ]]; do
+    path_=$1;shift
+    basename_=$(basename "$path_")
+    dirname_=$(dirname "$path_")
+    tag=$(tag_from_path)
```
reidpr: Give `tag_from_path` explicit rather than implicit arguments.

test/chtest-make-auto
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
```
reidpr: Script needs a comment explaining what it does and how to use it.

test/chtest-make-auto
```diff
@@ -0,0 +1,102 @@
...
+        exit 0
+    fi
+
+    if [[ $mode == build ]]; then
+        if (echo "$basename_" | grep Build &> /dev/null); then
+            template_print 'build_custom.bats.in'
+        elif (echo "$basename_" | grep Docker &> /dev/null); then
+            if (echo "$basename_" | grep Dockerfile &> /dev/null); then
+                template_print 'build.bats.in'
+            elif (echo "$basename_" | grep Docker_Pull &> /dev/null); then
```
reidpr: Weren't we going to eliminate the `Docker_Pull` logic and replace the one test with a `Build` script?

bin/ch-test
```diff
@@ -150,6 +149,158 @@ scope_check () {
     esac
 }
 
+scope_to_digit () {
```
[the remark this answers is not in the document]
j-ogas: Good point. The purpose of this is to compare scopes, e.g., `if [[ img_scope -gt scope ]]; then filter...`. This may not be the best approach. And alternative approach would be to make a filter function like:

```bash
if [[ $CH_TEST_SCOPE == quick ]]; then 
    filter 'scope' full standard
elif [[ $CH_TEST_SCOPE == standard ]]; then
    filter 'scope' fill
[...]
filter () {
   mode=$1; shift
   while [[ $# -gt 0 ]]; do
       case $mode in
       scope)
           # do something
           ;;
       builder-exclude)
           # do something
           ;;
       arch-exclude)
           # do something
           ;;
       esac
       shift
   done
}
```

bin/ch-test
```diff
@@ -19,10 +19,10 @@ builder_check () {
     printf 'checking builder ...\n'
     case $CH_BUILDER in
     ch-grow)
-        if ! ("${ch_bin}/ch-grow" --dependencies); then
```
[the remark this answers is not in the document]
j-ogas: To remain posix base.sh can't use `"${BASH_SOURCE[0]}"`. When I'd use ch-test from a prefix installation, e.g., `/usr/local/bin/ch-test`, ch-test's ch_bin would be my cwd. This behavior was corrected by making ch-test use its own bin variable using BASH_SOURCE[0].

bin/ch-test
```diff
@@ -150,6 +149,158 @@ scope_check () {
...
+
+    # shellcheck disable=SC2068
+    for i in ${images_all[@]}; do
+        if [[ $(test_make_check_image "$i") == ok ]]; then
+            echo "found: $i"
+            images+=( "$i" )
+        fi
+    done
+
+    make_auto="${testdir}/chtest-make-auto"
```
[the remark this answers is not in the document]
j-ogas: Absolutely.

bin/ch-test
j-ogas: Guarantee absolute path. If ch-test is sitting in some bin directory that is a symlink to somewhere else, I want to refer to that somewhere else directly. Not sure my preference for absolute path is justifiable (it very well could be hand-wavy), it just _feels_ right.

bin/ch-test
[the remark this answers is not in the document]
reidpr: OK. I wonder why the other scripts don't have the problem. I would prefer not to have the special case; maybe create an issue and we'll figure it out later?

bin/ch-test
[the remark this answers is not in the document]
reidpr: You definitely don't want to pack three functions into one with a "mode" selector. I think your current approach is probably fine; just needs some clarification.

bin/ch-test
```diff
@@ -150,6 +149,158 @@ scope_check () {
...
+                 find . -name 'Dockerfile.*';           \
+                 find . -name 'Docker_Pull.*';          \
+                 find ../examples -name  Build;         \
+                 find ../examples -name 'Build.*';      \
+                 find ../examples -name  Dockerfile;    \
+                 find ../examples -name 'Dockerfile.*'; \
+                 find ../examples -name  Docker_Pull;   \
+                 find ../examples -name 'Docker_Pull.*' )
+
+    # shellcheck disable=SC2068
```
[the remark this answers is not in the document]
reidpr: Aha. I think maybe the issue is that it is not really an array. Does `for i in $images_all` work?

If you want to take into account filenames with whitespace in them, then `-print0` and some Bash voodoo may work. But, filenames with whitespace are broken all over the code.

bin/ch-test
reidpr: I think you do not need it in that case. "Guarantee absolute path" is going to be rarely needed.

bin/ch-test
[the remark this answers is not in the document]
j-ogas: Of course now I can't reproduce. Reverting change for now.

test/chtest-make-auto
j-ogas: The thought here is that this script belongs in libexec and shouldn't be used directly outside of ch-test. I'll go ahead and comment on that.

Should I make a libexec directory now and move ch-test (along with base.sh there)?

reidpr: I'm creating a `libexec` directory in my `autotools2` branch. I'm a little skeptical of creating it in a 2nd branch for fear of merge conflicts. So I'd suggest leaving it where it is for now.

All that said, it still needs a comment. :)

j-ogas: > Should I make a libexec directory now and move ch-test (along with base.sh there)?
I am reluctant spending time moving stuff around if autotools is just going to refactor everything anyway.

reidpr: Sorry, to be more clear, no I don't think you should do those moves. I agree with you on the reasoning.

bin/ch-test
```diff
@@ -150,6 +149,158 @@ scope_check () {
...
+        skip*) # skip has the highest value to ensure comparison always fails
+            echo 3
+            ;;
+        *)
+            fatal "scope '$scope' invalid"
+    esac
+}
+
+test_build () {
+    echo 'executing build phase tests ...'
```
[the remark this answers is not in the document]
j-ogas: This doesn't function doesn't build tests, it runs the build phase tests. Perhaps "run build phase tests ..."?

bin/ch-test
```diff
@@ -150,6 +149,158 @@ scope_check () {
...
+            echo "found: $i"
+            images+=( "$i" )
+        fi
+    done
+
+    make_auto="${testdir}/chtest-make-auto"
+    printf '\n'
+    case $phase in
+    build)
+        printf 'generating build_auto.bats ...\n'
```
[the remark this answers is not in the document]
j-ogas: I do not think the users should be thinking about bat files, however, I think they should be aware an important test file artifact is being generated.

bin/ch-test
```diff
@@ -150,6 +149,158 @@ scope_check () {
...
+    *)
+        return 0
+        ;;
+    esac
+}
+
+test_make_check_image () {
+    img=$1
+    img_exclude_arch=$(cat "$img" | grep 'ch-test-arch-exclude: ' \
+                                  | sed 's/.*: //' | awk '{print $1}')
```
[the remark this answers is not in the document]
j-ogas: I have very limited `awk` skills. I wonder if optimizing this should be a separate issue? If not, I will have to spend some time tinkering with awk and reading the man page (god forbid).

test/chtest-make-auto
```diff
@@ -17,21 +33,21 @@ tag_from_path () {
...
 
     case $basicname in
-    Build|Dockerfile|Docker_Pull)
-        if [[ $parent_dir == '' ]]    || \
-           [[ $parent_dir == '.' ]]   || \
-           [[ $parent_dir == test ]]; then
+    Build|Dockerfile)
+        if    [[ $dir == '' ]]  \
+           || [[ $dir == '.' ]] \
+           || [[ $dir == test ]]; then
```
reidpr: `[[ ... ]]` should put OR and AND operators within the brackets, `[ ... ]` outside.

test/chtest-make-auto
```diff
@@ -40,8 +56,7 @@ tag_from_path () {
 template_print () {
     template="./make-auto.d/$1"
 
-    cat "$template" | sed "s@%(addr)s@$addr@g" \
-                    | sed "s@%(basename)s@$basename_@g" \
+    cat "$template" | sed "s@%(basename)s@$basename_@g" \
                     | sed "s@%(dirname)s@$dirname_@g" \
                     | sed "s@%(path)s@$path_@g" \
                     | sed "s@%(scope)s@$scope@g" \
```
reidpr: I'd say we should make an issue to change these and related pipes to a single sed/awk script.

test/common.bash
```diff
@@ -155,15 +155,6 @@ env_require CH_BUILDER
...
     env_require CH_GROW_STORAGE
 fi
-# Set path to the right Charliecloud. This uses a symlink in this directory
-# called "bin" which points to the corresponding bin directory, either simply
-# up and over (source code) or set during "make install".
-#
-# Note that sudo resets $PATH, so if you want to run any Charliecloud stuff
-# under sudo, you must use an absolute path.
-ch_bin="$(cd "$(dirname "${BASH_SOURCE[0]}")/bin" && pwd)"
-ch_bin="$(readlink -f "${ch_bin}")"
```
reidpr: OK, so this `readlink` was already in there. I wonder why ....

j-ogas: I believe the readlink _here_ was to make sure that the test/bin symlink was valid.

Issue #232, opened 2018-08-27
update Spark example


reidpr:
Currently the Spark example is version 2.0.2, which is pretty old (November 2016). Currently, 2.3.1 is the newest available.

There is some re-working to do in order to deal with the new Hive data store (see comments in `examples/other/spark/Dockerfile`).

While `SPARK_NO_DAEMONIZE=1` is available starting in 2.1.0, I think we may want to continue letting the start scripts daemonize in order to demonstrate that we can deal with this. (See also #230.)

reidpr:
We'll need to update the tutorial too, since it uses Spark.

PR #585, opened 2019-11-18
Update spark

rstyd:
Addresses #232. 

Updated the example to Spark 2.4.4. This ended up being much, much easier than it had been for Spark 2.1.0. The primary difference appears to be that the spark scripts now assume spark is installed in `/opt`.  I replaced the cloudfront link with an Apache link because the cloudfront mirror does not appear to have the most recent version.

### DIFF
--- a/examples/other/spark/Dockerfile
+++ b/examples/other/spark/Dockerfile
@@ -9,10 +9,15 @@ RUN    apt-get update \
 # We want ch-ssh
 RUN touch /usr/bin/ch-ssh
 
-# Download and install Spark.
-# We aren't using SPARK_NO_DAEMONIZE to make sure can deakwith daemonized spark. 
-# Spark is installed to /opt/spark which is spark's new default location
-# We disapprove of spark's master/slave terminology, but it's what the scripts are called
+# Download and install Spark. Notes:
+#
+# 1. We aren't using SPARK_NO_DAEMONIZE to make sure can deal with daemonized
+#    applications.
+#
+# 2. Spark is installed to /opt/spark, which is Spark's new default location.
+#
+# 3. We disapprove of Spark's master/slave terminology, but it's what the
+#    scripts are called, so we don't see a way to avoid it currently.
 
 ENV URLPATH https://www.apache.org/dist/spark/spark-2.4.4/
 ENV DIR spark-2.4.4-bin-hadoop2.7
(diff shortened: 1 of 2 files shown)


examples/other/spark/Dockerfile
```diff
@@ -11,29 +11,23 @@ RUN touch /usr/bin/ch-ssh
 
 # Download and install Spark.
 #
```
reidpr: Explain three things in this comment block:

1. Why we aren't using `SPARK_NO_DAEMONIZE`.
2. Why we are using `/opt/spark`.
3. That we disapprove of the master/slave terminology, but that's what the scripts are called, so we have to use it. (This is true, right? Are there better aliases?)

examples/other/spark/Dockerfile
```diff
@@ -11,29 +11,23 @@ RUN touch /usr/bin/ch-ssh
...
-# stack trace from pyspark. Workarounds exist, including cd to /tmp first or
-# configure hive-site.xml [2], but I'm not willing to put up with that crap
-# for demo purposes. Maybe it will be fixed in a 2.1 point release.
-#
-# [1]: http://spark.apache.org/docs/latest/sql-programming-guide.html#upgrading-from-spark-sql-20-to-21
-# [2]: https://community.cloudera.com/t5/Advanced-Analytics-Apache-Spark/Spark-displays-SQLException-when-Hive-not-installed/td-p/37954
-ENV URLPATH http://d3kbcqa49mib13.cloudfront.net
-ENV DIR spark-2.0.2-bin-hadoop2.7
+# Updated to spark-2.4.4
+
```
reidpr: Remove these two comment lines.

examples/other/spark/Dockerfile
```diff
@@ -11,29 +11,23 @@ RUN touch /usr/bin/ch-ssh
...
+ENV URLPATH https://www.apache.org/dist/spark/spark-2.4.4/
+ENV DIR spark-2.4.4-bin-hadoop2.7
 ENV TAR $DIR.tgz
+# [2]: https://community.cloudera.com/t5/Advanced-Analytics-Apache-Spark/Spark-displays-SQLException-when-Hive-not-installed/td-p/37954
```
reidpr: What's this comment for?

rstyd: This is a vestigial comment from before when this didn't work.

Issue #590, opened 2019-12-09
Add skopeo support to ch-pull2tar


rstyd:
ch-pull2tar currently only supports docker.  We would like it to fall-back to skopeo if docker is not installed.

reidpr:
Per discussion this morning, we are going to remove skopeo and umoci support (#675), making this issue moot.

PR #593, opened 2019-12-16
Update ch-pull2tar to use skopeo or docker

rstyd:
Addresses #590.


pagrubel:
I tried this out with neo4j, seems to work well.

bin/ch-pull2tar
```diff
@@ -20,12 +20,46 @@ EOF
...
 image=$1
 dest=$2
+tar="$image".tar.gz
 
-docker_ pull "$image"
-"${ch_bin}/ch-builder2tar" "$image" "$dest"
+while [ $# -gt 0 ]; do
+    opt=$1; shift
+    case $opt in 
+        -p|--pull-from)
```
reidpr: Call it `--pull-with`.

bin/ch-pull2tar
```diff
@@ -20,12 +20,46 @@ EOF
...
 dest=$2
+tar="$image".tar.gz
 
-docker_ pull "$image"
-"${ch_bin}/ch-builder2tar" "$image" "$dest"
+while [ $# -gt 0 ]; do
+    opt=$1; shift
+    case $opt in 
+        -p|--pull-from) 
+            if [ "$1" = "docker" ]; then
```
reidpr: This should be a `case` statement, and copy `$1` to a meaningful variable name.

bin/ch-pull2tar
```diff
@@ -20,12 +20,46 @@ EOF
...
+                docker=yes
```
reidpr: I'd have a single variable to contain the puller (i.e., values `docker` or `skopeo`). Eventually we are going to have ` `CH_PULLER` environment variable or similar, so prepare for that.

bin/ch-pull2tar
```diff
@@ -20,12 +20,46 @@ EOF
...
+                skopeo=yes
+            else 
+                echo "$1 is an invalid option" 1>&2
+                exit 1
+            fi
+            ;;
+    esac
+done
+
+if [ -n "$docker" ] || ( command -v docker >/dev/null 2>&1 ); then
```
reidpr: If someone says `--pull-with=docker` and there is no docker found, that should be an error. The current behavior simply falls through to skopeo.

bin/ch-pull2tar
```diff
@@ -20,12 +20,46 @@ EOF
...
+            elif [ "$1" = "skopeo" ]; then 
+                skopeo=yes
+            else 
+                echo "$1 is an invalid option" 1>&2
+                exit 1
+            fi
+            ;;
+    esac
+done
+
```
reidpr: Roughly here, add logic to select a default puller if none specified; also, say which puller was selected and why.

bin/ch-pull2tar
```diff
@@ -20,12 +20,46 @@ EOF
...
+if [ -n "$docker" ] || ( command -v docker >/dev/null 2>&1 ); then
+    echo "Docker"
+    docker_ pull "$image"
+    # shellcheck disable=SC2154
+    "${ch_bin}/ch-builder2tar" -b docker "$image" "$dest"
+elif [ -n "$skopeo" ] || ( command -v skopeo >/dev/null 2>&1 ); then 
+    echo "Skopeo"
```
reidpr: skopeo starts with a lower-case S. 🙄

bin/ch-pull2tar
```diff
@@ -20,12 +20,46 @@ EOF
...
+    temp=$(mktemp -d --tmpdir ch-pull2tar.XXXXXX)
+    skopeo copy docker://"$image" oci:"$temp"/oci:"$image"
```
reidpr: Format the strings as: `"docker://${image}"` and `"oci:${temp}/oci:${image}"`. Other strings elsewhere need a similar transformation.

bin/ch-pull2tar
```diff
@@ -20,12 +20,46 @@ EOF
...
+    umoci unpack --rootless --image "$temp"/oci:"$image" "$temp"/img
+    ( cd "$temp"/img/rootfs/
+    # shellcheck disable=SC2035
+    tar czf ../"$tar" * )
```
reidpr: 1. Put this subshell on a single line, for clarity.
2. Don't disable the ShellCheck warning; eliminate it. In this case, you don't need the glob. Instead, just tar `rootfs`; Charliecloud can deal with a top-level directory in the tarball.

bin/ch-pull2tar
```diff
@@ -20,12 +20,46 @@ EOF
...
+    cp "$temp"/img/"$tar" .
+    rm -rf "$temp"
```
reidpr: Add `--one-file-system` for safety.

test/Build.skopeo-umoci
```diff
@@ -1,37 +0,0 @@
-#!/bin/bash
```
reidpr: Why delete this file?

bin/ch-pull2tar
rstyd: Should I put in the plumbing for that right now, or should we just defer that to later?

test/Build.skopeo-umoci
rstyd: I removed it because the functionality it tested is now included in this script and tested by the  new ch-pull2tar test.

bin/ch-pull2tar
reidpr: I'd say yes. I think it will make your work on this PR easier.

test/Build.skopeo-umoci
reidpr: Sounds good.

reidpr: Sounds good!

bin/ch-pull2tar
rstyd: Will do!

bin/ch-pull2tar
rstyd: I chose this particular method because everything else was giving me the `implausibly old timestamp` error. I'll look for an alternative. I just tried ( cd $temp/img;  tar czf ../image.tar.gz rootfs/ ) and that resulted in:
```
[rustyd@mingus ch-pull2tar.GnLSWi]$ ch-tar2dir image.tar.gz .
creating new image ./image
tar: rootfs: implausibly old time stamp 1969-12-31 19:00:00
./image unpacked ok
```

reidpr: I think the solution is to find the directory that has the bad timestamp (`rootfs`?) and `touch(1)` it. We've seen this before; I think it's a bug in the base image.

rstyd: It ended up being a permissions issue all along.

rstyd:
The conflict is just that those two tests no longer exists in this version.

reidpr:
Per discussion this morning, we are not going to merge this. Thanks for your hard work though!!

Issue #580, opened 2019-11-08
test "lint shell scripts": fails with ShellCheck 0.7.0


rstyd:
The `lint shell scripts` test fails if shellcheck 0.7.0 is installed. 
```
 ✗ lint shell scripts
   (in test file build.bats, line 105)
     `| shellcheck -s bash -e SC1090,SC2002,SC2154,SC2164 -' failed
   shellcheck: version '0.7.0', major '0', minor '7'
   (Skip unrelated tests) 
   shellcheck: ./build_post.bats

   In - line 3:
   @test 'ARG and ENV'
   ^-- SC1073: Couldn't parse this bats @test. Fix to allow more checks.
                      ^-- SC1072:  Fix any mentioned problems and try again.

   For more information:
     https://www.shellcheck.net/wiki/SC1072 --  Fix any mentioned problems and t...
     https://www.shellcheck.net/wiki/SC1073 -- Couldn't parse this bats @test. F...
```
Shellcheck is complaining about a bats script. We explicitly guard against this behavior in `lint shell scripts`. It appears the new version breaks this. 

https://github.com/hpc/charliecloud/blob/af933999f39a9d99114f226e9d9796ad14ea34fb/test/build.bats#L97-L106

Here's a link to the static binary for shellcheck.0.7.0: https://storage.googleapis.com/shellcheck/shellcheck-stable.linux.x86_64.

Issue #577, opened 2019-10-31
test "lint shell scripts" in build.bats needs cleanup


reidpr:
It's missing files and confusing. I think we can boil it down to two classes:

* shell programs: `*.{ba,}sh`, or executable and have a shell shebang line
* `.bats` files, which need preprocessing

PR #612, opened 2020-01-08
clean up shell script lint

reidpr:
Addresses #577 and #580.

This PR adds some shell scripts to lint that were missed previously, so there are a few adjustments throughout.


test/Build.centos7xz
```diff
@@ -19,9 +19,9 @@
 
 set -ex
 
-srcdir=$1
+#srcdir=$1  # unused
```
rstyd: What's the reasoning for leaving these variables commented out instead of removing them?

test/build.bats
```diff
@@ -67,60 +67,82 @@ load common
...
+                   \(    -name .git \
+                      -o -name build-aux \) -prune \
+                -o \( -name '*.sh' -print \) \
+                -o \( -name '*.bash' -print \) \
+                -o \( -type f -exec awk '/^#!\/bin\/(ba)?sh/ {print FILENAME}
+                                         {nextfile}' {} + \) )
+    # Bats scripts. Use sed to do two things:
+    #
+    # 1. Make parseable by ShellCheck by removing "@test '...'". This does
+    #    remove the test names, but line numbers are still valid.
```
rstyd: What is the effect of removing the test names?

test/build.bats
```diff
@@ -67,60 +67,82 @@ load common
...
-    done
-    # BATS scripts
+    done < <( find "$ch_base" \
+                   \(    -name .git \
+                      -o -name build-aux \) -prune \
+                -o \( -name '*.sh' -print \) \
+                -o \( -name '*.bash' -print \) \
+                -o \( -type f -exec awk '/^#!\/bin\/(ba)?sh/ {print FILENAME}
+                                         {nextfile}' {} + \) )
+    # Bats scripts. Use sed to do two things:
```
j-ogas: Can we align the indention to line up 'name' (the common argument) as we do in `ch-test`?

test/Build.centos7xz
reidpr: To be consistent, i.e., always mention all three arguments in the same order. Commenting them out satisfies ShellCheck.

test/build.bats
reidpr: ShellCheck warnings will not be quite as clear. But the line numbers are correct, so you can still go look easily.

test/build.bats
reidpr: Well, the two groups of `-name` each line up within the group, unless I messed up. I don't quite follow. Can you write what indentation you propose?

Issue #602, opened 2019-12-20
re-implement "make web"


reidpr:
After PR #589, `make web` no longer works. Fix it as soon as 0.13 is out the door.

PR #618, opened 2020-01-14
restore script to publish web docs

reidpr:
Addresses #602.

Note that some of the documentation which appears to have disappeared was moved to a more logical home on the internal wiki.


doc/publish
```diff
@@ -0,0 +1,72 @@
...
+commit=$(cd .. && git rev-parse --short HEAD)${dirty}
+set -x
+git add --all
+git commit -a -m "docs for commit $commit"
+git push origin gh-pages
+set +x
+
+# Done.
+echo 'Done.'
+echo 'Typos found: 1'
```
rstyd: Do we intend to say that we found 1 typo each time?

doc/publish
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+# This script builds the documentation and then publishes it to the web. See
+# the internal documentation for usage and how to set it up.
```
rstyd: Internal hear means it won't be on github?

doc/publish
reidpr: Yes, it's an Easter egg. ;)

doc/publish
reidpr: Yes.

reidpr: Yes.

doc/publish
rstyd: Fair enough!

Issue #586, opened 2019-11-27
change ch-test behavior when a supported builder is not installed


j-ogas:
Without a valid builder installed we get:
```
[vagrant@charliebox ~]$ ch-test
ch-test version 0.12~pre+82eee98

ch-run: 0.12~pre+82eee98 /usr/local/bin/ch-run
bats:   0.4.0 /usr/bin/bats
tests:  /usr/local/libexec/charliecloud-0.12~pre+82eee98/test

phase:                build run examples
scope:                standard (default)
builder:              ch-grow (default)
ch-grow storage:      /var/tmp/ch-grow (default)
use generic sudo:     no (default)
unpacked images dir:  /var/tmp/images (environment)
packed images dir:    /var/tmp/tarballs (environment)
fs permissions dirs:  skip (environment)

checking namespaces ...
ok

checking builder ...
/usr/local/bin/ch-test: line 22: /usr/local/bin/ch-grow: No such file or directory
error: builder: ch-grow: missing dependencies
```

reidpr:
I wonder if the way to address this is to:

1. Add `--builder=none` and `CH_BUILDER=none`. `ch-test` accepts these and does its best (it should mostly work because the `chtest` image doesn't require a builder, but many tests are skipped). `ch-build` also accepts them, but errors because it really does require a builder.
2. Make `none` be the default builder if no external builders are available and `ch-grow` doesn't have its dependencies.

PR #621, opened 2020-01-15
Support tests without a builder

j-ogas:
Addresses: #586

### DIFF
--- a/bin/ch-test
+++ b/bin/ch-test
@@ -71,12 +71,18 @@ builder_check () {
         bl=$(command -v docker)
         bv=$(docker_ --version | awk '{print $3}' | sed -e 's/,$//')
         ;;
+    none)
+        ;;
     *)
         fatal "builder: $CH_BUILDER: not supported"
         ;;
     esac
-    printf 'found: %s %s\n\n' "$bl" "$bv"
-    version_check 'builder' "$min" "$bv"
+    if [[ $CH_BUILDER == none ]]; then
+        printf 'found: none\n\n'
+    else
+        printf 'found: %s %s\n\n' "$bl" "$bv"
+        version_check 'builder' "$min" "$bv"
+    fi
 }
 
 builder_set () {
@@ -205,7 +211,7 @@ test_build () {
 test_examples () {
     local tag
     local phase=examples
-    if [[ $CH_TEST_SCOPE == quick ]]; then
+    if [[ $CH_TEST_SCOPE == quick || $CH_BUILDER == none ]]; then
         return
     fi
     echo
@@ -277,6 +283,8 @@ test_make_auto () {
 test_make_check_image () {
     img=$1
     dir=$(basename "$(dirname "$img")")
+    builder_required=$(cat "$img" | grep -F 'ch-test-builder-required: ' \
+                                  | sed 's/.*: //' | awk '{print $1}')
     exclude_arch=$(cat "$img" | grep -F 'ch-test-arch-exclude: ' \
                               | sed 's/.*: //' | awk '{print $1}')
 
@@ -299,6 +307,22 @@ test_make_check_image () {
             return 1
         fi
     done
+
+    # Exlude images that don't match a required builder.
+    case $builder_required in
+    any|yes) # 'yes' and 'any' behave the same. The later is preferred for
+             # clarity in Dockerfiles that require a builder but exclude
+             # specific builder (which is handled below).
+        if [[ $CH_BUILDER == none ]]; then
+            return 1
+        fi
+        ;;
+    *)
+        if [[ $builder_required != "$CH_BUILDER" ]]; then
+            return 1
+        fi
+    esac
+
     # Exclude images that are not compatible with CH_BUILDER.
     for b in $exclude_builder; do
         if [[ $b == "$CH_BUILDER" ]]; then
--- a/bin/ch-test
+++ b/bin/ch-test
@@ -278,9 +278,10 @@ test_make_auto () {
 }
 
 test_make_check_image () {
+    local include
     img=$1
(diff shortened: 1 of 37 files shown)


test/Build.skopeo-umoci
```diff
@@ -13,6 +13,7 @@
 # issue #332 are still tested.
 #
 # ch-test-scope: standard
+# ch-test-builder-required: none
```
reidpr: So this one only works with `--builder=none`? Shouldn't it work with all builders, because no builder is used?

test/build_post.bats
```diff
@@ -8,6 +8,8 @@ load common
     # seconds for ch-grow.
     if [[ $CH_BUILDER = ch-grow ]]; then
         scope standard
+    elif [[ $CH_BUILDER = none ]]; then
+        skip 'no builder'
```
reidpr: Can this use `builder_required`?

test/run/ch-fromhost.bats
```diff
@@ -31,6 +31,7 @@ fromhost_ls () {
 
 @test 'ch-fromhost (Debian)' {
     scope standard
+    builder_required
```
reidpr: Can this be covered with `prerequisites_ok`? Because if there is no builder, then the image won't be built. Let's discuss.

test/common.bash
```diff
@@ -26,6 +26,12 @@ builder_ok () {
     #docker_tag_p "${1}:$(ch-run --version |& tr '~+' '--')"
 }
 
+builder_required () {
+    if [[ $CH_BUILDER == none ]]; then
+        skip 'builder required'
+    fi
+}
+
```
reidpr: The sense of this name implies, I think, that a builder is required, and the test will fail if it's not there. But, it just skips.

test/travis.yml
```diff
@@ -65,6 +65,8 @@ jobs:
     - <<: *stage_quick
       env: CH_BUILDER=docker CH_TEST_SCOPE=quick
 
+    - <<: *stage_builders
+      env: CH_BUILDER=none
```
reidpr: I'd like to keep the stages to 5 jobs, because that's what we can run in parallel, so the 6th job makes it take twice as long. I wonder if we should omit `buildah-setuid`, because that seems perhaps the least interesting of the Buildah flavors?

j-ogas: For our purposes I think testing without a builder is more important than testing the buildah setuid subset.

test/run/ch-fromhost.bats
j-ogas: I'm not sure `prerequisites_ok` works anymore. The only sentinel tarball that is generated is `missing.pq_missing`, other dockerfiles/build are just not included in the generation of build|run_auto.bats anymore.

bin/ch-test
```diff
@@ -71,6 +71,10 @@ builder_check () {
         bl=$(command -v docker)
         bv=$(docker_ --version | awk '{print $3}' | sed -e 's/,$//')
         ;;
+    none)
+        bl=
+        bv=
```
j-ogas: An issue I have with this is the test suite output:

```
checking namespaces ...
ok

checking builder ...
found:

creating /var/tmp/tar
[...]
```

I'd prefer an explicit output here instead implicit.

reidpr: Yes, I think you're right.

examples/Dockerfile.exhaustive
```diff
@@ -4,18 +4,15 @@
...
+# ch-test-builder-include: ch-grow
 
 #FROM alpine
 #FROM alpine@cdf98d1859c1  # skopeo pukes on this
 FROM alpine:3.9
 #FROM alpine:3.9 AS stage1
 
 # FIXME: These should maybe be moved into Dockerfile.argenv?
-ENV chse_1 value1 with spaces and trailing space 
+ENV chse_1 value1 with spaces and trailing space
```
reidpr: The single trailing space here is significant and should remain.

test/Makefile.am
```diff
@@ -32,7 +32,6 @@ Build.ch-build2dir \
 Build.ch-pull2dir \
 Build.ch-pull2tar \
 Build.docker_pull \
-Build.missing \
```
reidpr: This should remain because it tests the `prerequisites_ok` function.

test/common.bash
```diff
@@ -96,7 +96,7 @@ need_squashfs () {
 }
 
 prerequisites_ok () {
-    if [[ -f $CH_TEST_TARDIR/${1}.pq_missing ]]; then
+    if [[ ! -f $CH_TEST_TARDIR/${1}.tar.gz ]]; then
```
reidpr: I think we should leave the `pq_missing` functionality alone. It's OK if this PR doesn't make use of it, but I think it's an important feature. Eventually we may want to hook into it for this.

Also, this will fail if using SquashFS, no?

bin/ch-test
```diff
@@ -102,12 +102,16 @@ builder_set () {
...
-# Create CH_TEST_TARDIR, avoiding #347.
+# Create CH_TEST_IMGDIR, avoiding #347.
 dir_img_mk () {
-    if [[ ! -d $CH_TEST_IMGDIR ]]; then
-        printf "creating %s\n\n" "$CH_TEST_IMGDIR"
-        mkdir "$CH_TEST_IMGDIR"
-    fi
+    dir_img_rm
+    printf "creating %s\n\n" "$CH_TEST_IMGDIR"
+    mkdir "$CH_TEST_IMGDIR"
```
j-ogas: Doesn't this mean the manual `run` phase following a manual `build` will always fail?

reidpr:
Thanks; merged. The "Revert" commit was a mis-click and should be ignored.

Issue #406, opened 2019-04-09
test suite: add Lustre ioctl test(s)


reidpr:
Some apps like to make Lustre ioctl calls on Lustre directories bind-mounted into the container. Does this work? Let's test it to see.

Not sure whether installing the Lustre utilities or making the ioctl calls directly in a toy app would be better.

Brofessional:
Sending Lustre ioctls does appear to work from inside a container.

I was able to run `lfs getstripe` and `lfs setstripe` from within the container on a bind mounted Lustre share.

Some notes:
Lustre can be built inside a container from source
The container must be writable, the `lfs` command appears to write some things in order to perform the work.

It could be taken a step further to make a script that manually calls the ioctls, but that seemed like more work than just building Lustre in a container and testing common commands.

We could add these tests to the suite, but I'm not sure how we would go about skipping them if testing on a machine with no Lustre.

Brofessional:
I managed to prune the Lustre client image to about 1GB in Docker and 320MB on disk after exporting to squashfs. 
I know it's been a while since we talked about it but I would like to make a case for using the C API or LFS.
I spent a lot of time looking at the Lustre code and such to try and figure out sending IOCTLs manually. 
1) You would need a set of cracked header files that would allow you to import them as the ones in Lustre won't allow it without modification. 
2) You would need to spend a lot of time figuring out the rather complex data types being sent through IOCTLs to make a test with meaningful results. Simply sending trash IOCTLs doesn't have any value. It turns out that sending IOCTLs to a distributed filesystem is complicated. There's a lot of legwork in knowing how to send an IOCTL that will do anything useful and not corrupt data or anything. 
3) The barrier to entry for the C API or LFS is rather low and can be easily scripted in a way that we can perform meaningful tests.
4) Sending IOCTLs all willy nilly can be dangerous, which is why Lustre doesn't want you doing it. Small mistakes can have big consequences.

I think sending IOCTLs is more work, less reliable, less meaningful, and silly when Lustre provides ways to interact with Lustre as intended.

reidpr:
Good arguments. How hard is it to do an image including the C API but not the LFS utilities?

Brofessional:
That could be possible if you build from source, but is unlikely. Installing via RPM installs them both. You could just dumpster the LFS command line binary in the image.

Brofessional:
[...]

PR #623, opened 2020-01-15
Add Lustre tests through LFS

Brofessional:
Add Lustre tests to the examples


Brofessional:
Addresses #406

examples/lustre/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN yum install -y epel-release
+
+# more dependencies
+RUN yum -y install libyaml-devel \
+                   wget \
+                   gcc
+
+
+RUN wget https://downloads.whamcloud.com/public/lustre/latest-release/el7/client/RPMS/x86_64/lustre-client-2.12.3-1.el7.x86_64.rpm && wget https://downloads.whamcloud.com/public/lustre/latest-release/el7/client/RPMS/x86_64/kmod-lustre-client-2.12.3-1.el7.x86_64.rpm
+
```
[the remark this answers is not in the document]
Brofessional: Lustre is not available in epel.
```
$ yum provides lfs
Loaded plugins: fastestmirror, ovl
Determining fastest mirrors
epel/x86_64/metalink                                                                                                           |  18 kB  00:00:00     
 * base: mirror.sesp.northwestern.edu
 * epel: fedora.mirrors.pair.com
 * extras: linux.mirrors.es.net
 * updates: mirror.lug.udel.edu
base                                                                                                                           | 3.6 kB  00:00:00     
epel                                                                                                                           | 5.3 kB  00:00:00     
extras                                                                                                                         | 2.9 kB  00:00:00     
updates                                                                                                                        | 2.9 kB  00:00:00     
(1/7): epel/x86_64/group_gz                                                                                                    |  90 kB  00:00:00     
(2/7): base/7/x86_64/group_gz                                                                                                  | 165 kB  00:00:00     
(3/7): epel/x86_64/updateinfo                                                                                                  | 1.0 MB  00:00:00     
(4/7): extras/7/x86_64/primary_db                                                                                              | 159 kB  00:00:00     
(5/7): base/7/x86_64/primary_db                                                                                                | 6.0 MB  00:00:01     
(6/7): epel/x86_64/primary_db                                                                                                  | 6.9 MB  00:00:00     
(7/7): updates/7/x86_64/primary_db                                                                                             | 5.9 MB  00:00:01     
base/7/x86_64/filelists_db                                                                                                     | 7.3 MB  00:00:01     
epel/x86_64/filelists_db                                                                                                       |  12 MB  00:00:02     
extras/7/x86_64/filelists_db                                                                                                   | 210 kB  00:00:00     
updates/7/x86_64/filelists_db                                                                                                  | 3.6 MB  00:00:00     
No matches found
```
There is only an x86_64 RPM, I will add the exclude aarch64 comment good catch.

examples/lustre/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# ch-test-scope: full
+# ch-test-builder-exclude: ch-grow  # deferred
```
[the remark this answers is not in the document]
Brofessional: It should. I think that was just copied from another place and didn't need to be there

bin/ch-test
```diff
@@ -426,6 +426,10 @@ while [[ $# -gt 0 ]]; do
     --sudo)
         use_sudo=yes
         ;;
+    --lustre)
+        export ch_lustre="$1"
+	shift
+        ;;
```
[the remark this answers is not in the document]
Brofessional: A single assignment is the way to go.

bin/ch-test
```diff
@@ -36,6 +36,7 @@ Additional options:
   --pack-dir DIR         packed imaged directory; same as \$CH_TEST_TARDIR
   --perm-dir DIR         permissions fixture dir, can repeat; \$CH_TEST_PERMDIRS
   --sudo                 enable tests that need sudo
+  --lustre DIR           run lustre tests in DIR
```
reidpr: How about "use DIR for run-phase Lustre tests"?

bin/ch-test
```diff
@@ -426,6 +427,10 @@ while [[ $# -gt 0 ]]; do
     --sudo)
         use_sudo=yes
         ;;
+    --lustre)
+        require_unset ch_lustre
+        export ch_lustre="$1"; shift
+        ;;
```
reidpr: 1. Similar to the other PR, add `--lustre=foo` processing.
2. You don't need to `export` and can omit the double quotes.

doc/ch-test_desc.rst
```diff
@@ -128,6 +128,9 @@ Additional arguments:
     generic :code:`sudo` capabilities. Note that the Docker builder uses
     :code:`sudo docker` even without this option.
 
+  :code:`--lustre DIR`
+    Enable lustre LFS tests under :code:`DIR`
```
reidpr: How about:

```
Use :code:`DIR` for run-phase Lustre tests
```
?

examples/lustre/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# ch-test-scope: full
+# ch-test-arch-exclude: aarch64
+# No lustre RPMS for aarch64
```
reidpr: This can go on the same line as the `ch-test-arch-exclude`.

examples/lustre/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+# more dependencies
+RUN yum -y install libyaml-devel \
+                   wget \
+                   gcc
+
+ENV LUSTRE_URL=https://downloads.whamcloud.com/public/lustre/latest-release/el7/client/RPMS/x86_64
+
+RUN wget ${LUSTRE_URL}/lustre-client-2.12.3-1.el7.x86_64.rpm && \
+    wget ${LUSTRE_URL}/kmod-lustre-client-2.12.3-1.el7.x86_64.rpm
```
reidpr: Format as:

```

RUN    wget ${LUSTRE_URL}/lustre-client-2.12.3-1.el7.x86_64.rpm \
    && wget ${LUSTRE_URL}/kmod-lustre-client-2.12.3-1.el7.x86_64.rpm
```

(This is so the `&&` doesn't get lost at the end of the line.)

examples/lustre/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+# install development tools
+RUN yum install -y epel-release
+
+# more dependencies
+RUN yum -y install libyaml-devel \
+                   wget \
+                   gcc
+
+ENV LUSTRE_URL=https://downloads.whamcloud.com/public/lustre/latest-release/el7/client/RPMS/x86_64
```
reidpr: Is it worth a `LUSTRE_VERSION` too?

examples/lustre/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN wget ${LUSTRE_URL}/lustre-client-2.12.3-1.el7.x86_64.rpm && \
+    wget ${LUSTRE_URL}/kmod-lustre-client-2.12.3-1.el7.x86_64.rpm
+
+RUN yum localinstall -y *.rpm
+
+WORKDIR /lustre
```
reidpr: This instruction will have no effect in Charliecloud, since it's the last one in the Dockerfile. It looks like you are using it for the directory creation side effect? If so, I'd recommend `RUN mkdir ...`.

examples/lustre/test.bats
```diff
@@ -0,0 +1,37 @@
+load "${CHTEST_DIR}/common.bash"
+
+setup () {
+    scope standard
+    if [[ -z "${ch_lustre}" ]]; then
+        skip 'no lustre to bind mount'
+    fi
+    prerequisites_ok lustre
```
reidpr: Put this on the line after `scope`.

examples/lustre/test.bats
```diff
@@ -0,0 +1,37 @@
+load "${CHTEST_DIR}/common.bash"
+
+setup () {
+    scope standard
```
reidpr: I wonder if full scope is better for this one?

examples/lustre/test.bats
```diff
@@ -0,0 +1,37 @@
...
+    if [[ -z "${ch_lustre}" ]]; then
```
reidpr: No need to quote in Bash; you can just say `if [[ -z $ch_lustre ]]; then`.

examples/lustre/test.bats
```diff
@@ -0,0 +1,37 @@
...
+        skip 'no lustre to bind mount'
```
reidpr: I think the logic should be:
1. If not in a Slurm allocation and no Lustre dir, then just skip.
2. If in a Slurm allocation and no Lustre dir, then `pedantic_fail`, to force the team to run the Lustre tests.

examples/lustre/test.bats
```diff
@@ -0,0 +1,37 @@
...
+    fi
+    prerequisites_ok lustre
+
+    # Check lustre directory exists
+    if [[ ! -e "$ch_lustre" ]]; then
+        skip "${ch_lustre} does not exist"
```
reidpr: Fail the test here.

examples/lustre/test.bats
```diff
@@ -0,0 +1,37 @@
...
+    fi
+    # Check lustre directory is a directory
+    if [[ ! -d "$ch_lustre" ]]; then
+        skip "${ch_lustre} is not a directory"
```
reidpr: Fail this one too.

examples/lustre/test.bats
```diff
@@ -0,0 +1,37 @@
...
+    fi
+}
+
+@test "${ch_tag}/write" {
+    binds=${ch_lustre}:/lustre
```
reidpr: 1. Is there a reason to use `/lustre` rather than existing `/mnt`?
2. I recommend to set `binds` once at the top of the file, to avoid repeating it.

examples/lustre/test.bats
```diff
@@ -0,0 +1,37 @@
...
+    ch-run -b "$binds" "$ch_img" -- touch /lustre/test_w.txt
+    ch-run -b "$binds" "$ch_img" -- echo "hello" /lustre/test_w.txt
```
reidpr: 1. These test creation and writing separately; is that your intent? If so, I recommend two separate tests. If not, I'd go with just hello.
2. Do you mean `echo hello > /lustre/test_w.txt` (with redirect)?

examples/lustre/test.bats
```diff
@@ -0,0 +1,37 @@
...
+}
+
+@test "${ch_tag}/read" {
+    binds=${ch_lustre}:/lustre
+    ch-run -b "$binds" "$ch_img" -- cat /lustre/test_w.txt > /dev/null
```
reidpr: 1. Is there a way to dump all the caches and read this from disk?
2. I recommend you check the content is as expected.

examples/lustre/test.bats
```diff
@@ -0,0 +1,37 @@
...
+}
+
+@test "${ch_tag}/set_stripe_get_stripe" {
```
reidpr: 1. Simplify title, e.g. "stripes" or "striping".
2. General style note: Test names are human readable and can use spaces, etc.

examples/lustre/test.bats
```diff
@@ -0,0 +1,37 @@
...
+    binds=${ch_lustre}:/lustre
+    ch-run -b "$binds" "$ch_img" -- mkdir /lustre/default_stripes
+    ch-run -b "$binds" "$ch_img" -- lfs getstripe /lustre/default_stripes
+    ch-run -b "$binds" "$ch_img" -- mkdir /lustre/four_stripes
+    ch-run -b "$binds" "$ch_img" -- lfs setstripe -c 4 /lustre/four_stripes
```
reidpr: Check that the number of stripes is as expected.

examples/lustre/Dockerfile
Brofessional: If we wanted to use a static version of Lustre yes. As we are only using the userspace tools I expect there won't be a problem with using the latest release. But that is dangerous thinking, as it very well could come back to bite us. We could settle on a version and update it sometimes as needed. If we do that we might want a system to make sure it doesn't get too stale.

examples/lustre/test.bats
Brofessional: Just for human brain continuity. So no technical reason. I think I will make a change to address this in a different way.

examples/lustre/test.bats
Brofessional: I will investigate. The C API you can. I'll see if there is a way with the lfs tool.

examples/lustre/Dockerfile
Brofessional: Behold as if by magic not having a lustre version has impacted me. I'll fix that.

examples/lustre/test.bats
Brofessional: Turns out you can set IO flags using dd

examples/lustre/test.bats
```diff
@@ -0,0 +1,76 @@
...
+        if [[ -z $ch_lustre ]]; then
+            skip 'no lustre to bind mount'
+        fi
+    fi
+
+    # Check lustre directory exists
+    if [[ ! -e $ch_lustre ]]; then
+        echo "${ch_lustre} does not exist"
+        exit 1
+    fi
```
reidpr: This is redundant with the next if statement and should be removed.

examples/lustre/test.bats
```diff
@@ -0,0 +1,76 @@
...
+    # Check lustre directory is a directory
+    if [[ ! -d $ch_lustre ]]; then
+        echo "${ch_lustre} is not a directory"
+        exit 1
+    fi
+
+    binds=${ch_lustre}:/mnt/0
+    work_dir=/mnt/0/charliecloud_testing_directory
```
reidpr: Set these at the top level, not in `setup`.

reidpr: I'd call it just `charliecloud_test`.

examples/lustre/test.bats
```diff
@@ -0,0 +1,76 @@
...
+    mkdir -p "${ch_lustre}/charliecloud_testing_directory"
```
reidpr: Also remove any existing files so you're starting with a clean slate. You're doing the right thing below, to avoid the dangerous `rm -Rf`. Move those commands into a function and call it both places.

examples/lustre/test.bats
```diff
@@ -0,0 +1,76 @@
...
+}
+
+@test "${ch_tag}/create dir" {
```
reidpr: Call it "create directory".

examples/lustre/test.bats
```diff
@@ -0,0 +1,76 @@
...
+    ch-run -b "$binds" "$ch_img" -- mkdir "${work_dir}/tryme"
```
reidpr: Call it `test_create_dir` or similar.

examples/lustre/test.bats
```diff
@@ -0,0 +1,76 @@
...
+}
+
+@test "${ch_tag}/create file" {
+    ch-run -b "$binds" "$ch_img" -- touch "${work_dir}/test_create.txt"
```
reidpr: Call it `test_create_file` or similar; I don't think you need an extension.

examples/lustre/test.bats
```diff
@@ -0,0 +1,76 @@
...
+}
+
+
+@test "${ch_tag}/delete file" {
+    ch-run -b "$binds" "$ch_img" -- rm -f "${work_dir}/test_create.txt"
```
reidpr: Omit `-f`; you want the test to fail if it doesn't work.

examples/lustre/test.bats
```diff
@@ -0,0 +1,76 @@
...
+}
+
+@test "${ch_tag}/write" {
+    ch-run -b "$binds" "$ch_img" -- sh -c "sleep 5 && echo hello | dd of=${work_dir}/test_write.txt"
```
reidpr: Explain why you are sleeping and why you're using `dd`.

IIRC, you're using `dd` to avoid the disk cache? If so, use an explicit option for that.

examples/lustre/test.bats
```diff
@@ -0,0 +1,76 @@
...
+@test "${ch_tag}/create file" {
+    ch-run -b "$binds" "$ch_img" -- touch "${work_dir}/test_create.txt"
+}
+
+
+@test "${ch_tag}/delete file" {
+    ch-run -b "$binds" "$ch_img" -- rm -f "${work_dir}/test_create.txt"
+}
+
+@test "${ch_tag}/write" {
```
reidpr: You could call these `write file` etc., to be consistent.

examples/lustre/test.bats
```diff
@@ -0,0 +1,76 @@
...
+    ch-run -b "$binds" "$ch_img" -- sh -c "sleep 5 && echo hello | dd of=${work_dir}/test_write.txt"
+}
+
+@test "${ch_tag}/read" {
+    out=$(ch-run -b "$binds" "$ch_img" -- dd if="${work_dir}/test_write.txt" iflag=nocache 2> /dev/null)
```
reidpr: Use Bats' `run` command for this. The idiom is to check the return code is what you expect, print the output, and then compare the actual output with expected, including stderr, using `diff -u`, which will fail if there is a difference. There should be examples elsewhere in the code to work from.

examples/lustre/test.bats
```diff
@@ -0,0 +1,76 @@
...
+    if [ "$out" != hello ]; then
+        echo "Content of ${work_dir}/test_write.txt invalid" && exit 1
+    fi
+}
+
+@test "${ch_tag}/striping" {
+    ch-run -b "$binds" "$ch_img" -- mkdir "${work_dir}/set_stripes"
+    ch-run -b "$binds" "$ch_img" -- lfs setstripe -c 1 "${work_dir}/set_stripes"
+    one_stripe=$(ch-run -b "$binds" "$ch_img" -- lfs getstripe "${work_dir}/set_stripes/" | grep stripe_count | gawk '{ print $2}')
```
reidpr: I think you can use the `run` and `diff` idiom here too.

examples/lustre/test.bats
Brofessional: Turns out that won't work here. `dd` displays the speed at which data was moved which will vary on every run, so there is no way to match the output of stderr here.

Brofessional: I can dump the standard error but that defeats the purpose of using the diff idiom here.

examples/lustre/test.bats
Brofessional: This is further complicated by different Lustre systems having different default settings. There is no way to match all the output using the run and diff idiom here on every system.

We could:
- Not use the run diff idiom here
- Capture all output for the "default" then compare other output to it to make sure it has changed

Brofessional: Since system defaults can change this won't really work here. We could set a stripe count, then get the stripe count using `run` then we could change the stripe count and make sure the diff fails.

Brofessional: But i might be missing something. I can't think of another way to really compare the unpredictable output. Any thoughts?

examples/lustre/test.bats
reidpr: If there's not too much varying output, you can do something like:
```
$ cat <<'EOF' | diff -u - <(echo "$output" | sed -E s'/(copied),.*$/\1/')
0+1 records in
0+1 records out
234 bytes copied
EOF
--- -	2020-02-26 15:46:21.656131355 -0700
+++ /dev/fd/63	2020-02-26 15:46:21.651184648 -0700
@@ -1,3 +1,3 @@
 0+1 records in
 0+1 records out
-234 bytes copied
+235 bytes copied
```

examples/lustre/test.bats
reidpr: How about something like this:

```
mkdir foo
old=$(lfs getstripe --stripe-count foo)
echo "old: $stripe_ct_old"
new_expected=$(($old * 2))
echo "new expected: $new_expected"
new_actual=$(lfs getstripe --stripe-count foo)
echo "new actual: $new_actual"
[[ $new_expected -eq $new_actual ]]
```

I think you're right that the `diff` trick is overkill here. The test will fail if any of the `lfs` invocations fail.

You might want to set a variable or function which contains all the `ch-run` options you need in this script, to reduce repetition.

examples/lustre/test.bats
Brofessional: Turns out there is a flag to drop the transfer speed on dd

Issue #565, opened 2019-10-04
make docs build with EPEL Sphinx


reidpr:
I believe we are close to making the docs build without error in EPEL's Sphinx (which is either 1.1.3 or 1.2.3 depending on if we choose the Python 2 or Python 3 version?). Docutils looks to be 0.11 or 0.14, and RTD theme 0.2.4.

If this is true, let's make those changes, because it makes packaging easier.

I do believe we should put a note saying that while the docs build with these old versions, it is not recommended.

reidpr:
@j-ogas proposes this patch:

```
   --- 0.10/doc-src/conf.py	                2019-09-03 18:42:01.250000000 -0600
   +++ 0.10-patch/doc-src/conf.py	2019-09-04 10:07:51.390000000 -0600
   @@ -91,8 +91,9 @@

   # The theme to use for HTML and HTML Help pages.  See the documentation for
   # a list of builtin themes.
   +import sphinx_rtd_theme
   html_theme = 'sphinx_rtd_theme'
   -
   +html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
   highlight_language = 'console'
```

j-ogas:
> @j-ogas proposes this patch:
> 
> ```
>    --- 0.10/doc-src/conf.py	                2019-09-03 18:42:01.250000000 -0600
>    +++ 0.10-patch/doc-src/conf.py	2019-09-04 10:07:51.390000000 -0600
>    @@ -91,8 +91,9 @@
> 
>    # The theme to use for HTML and HTML Help pages.  See the documentation for
>    # a list of builtin themes.
>    +import sphinx_rtd_theme
>    html_theme = 'sphinx_rtd_theme'
>    -
>    +html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
>    highlight_language = 'console'
> ```

FWIIW, this patch reverts a change introduced in commit d8ed5e6cd9a859186dbaa33610ab86fbbe1c57b7

PR #624, opened 2020-01-15
make docs build with old Sphinx

reidpr:
Addresses #565.

I tested this using the highest versions of Sphinx (1.2.3), docutils (0.14), and sphinx-rtd-theme (0.2.4) available in EPEL7, but installed using `pip3`. I have not tried it on actual CentOS or RHEL; can you please test that?

The number of changes to the `.rst` file was surprisingly small, but the results look OK.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -127,11 +127,50 @@ AC_DEFUN([CH_CHECK_VERSION], [
 AC_CHECK_FUNCS(asprintf, [], [
   AC_MSG_ERROR([asprintf() not found. Please report this bug.])])
 
-vmin_sphinx=1.4.9
+# Sphinx
+vmin_sphinx=1.2.3
 AX_WITH_PROG([SPHINX], [sphinx-build])
 AC_SUBST([SPHINX])
 CH_CHECK_VERSION([SPHINX], [$vmin_sphinx],
                  [--version | sed -E 's/sphinx-build //'])
+
+# Get the Sphinx Python
+AS_IF([test -n "$SPHINX"], [
+  AC_MSG_CHECKING([for sphinx-build Python])
+  sphinx_python=$(head -1 "$SPHINX" | sed -E 's/^..//')
+  AC_MSG_RESULT([$sphinx_python])
+])
+
+# "docutils" module
+vmin_docutils=0.14
+docutils=$sphinx_python  # FIXME: output is confusing
+AS_IF([test -n "$SPHINX"], [
+  # Sphinx requires docutils, so don't verify it's there.
+  CH_CHECK_VERSION([docutils], [$vmin_docutils],
+                   [-c 'import docutils; print(docutils.__version__)'])
+])
+
+# "sphinx-rtd-theme" module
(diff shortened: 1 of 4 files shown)


j-ogas:
The epel 7 package, `python3-sphinx`, which provides `sphinx-build` version 1.2.3, packages the executable in /usr/bin as `sphinx-build-3.6`. Thus, the configure sets documentation build to no when it fails to find `sphinx-build`.

reidpr:
> packages the executable in /usr/bin as `sphinx-build-3.6`.

Well that's convenient. 🙄

Debian packages have a notion of "alternatives" and would set up a `sphinx-build` symlink. Do the EPEL RPMs have an equivalent?

Also, one could configure with `--with-sphinx-build=/usr/bin/sphinx-build-3.6`; would that be acceptable?

j-ogas:
> Debian packages have a notion of "alternatives" and would set up a `sphinx-build` symlink. Do the EPEL RPMs have an equivalent?

Yes, if the alternatives are defined in the package spec file. However, the python3-sphinx package [spec file](https://src.fedoraproject.org/rpms/python3-sphinx/blob/epel7/f/python3-sphinx.spec) does not appear to have them.

FWIIW, this specific package is dead and only maintained on epel 7 (there is a different package for epel 8). 
> 
> Also, one could configure with `--with-sphinx-build=/usr/bin/sphinx-build-3.6`; would that be acceptable?

This seems reasonable to me. I can account for this conditionally in the spec file; still much cleaner than the doc patch we have now.

olifre:
I'm sorry I only notice this now, but this PR broke the build on Gentoo Linux (and potentially others). 
The problem is that:
```
sphinx_python=$(head -1 "$SPHINX" | sed -E 's/^..//')
```
does not work. On Gentoo, any python tool which is installed with multiple implementations is wrappered:
```
$ ls -la $(which sphinx-build)
lrwxrwxrwx 1 root root 31  4. Jan 18:44 /usr/bin/sphinx-build -> ../lib/python-exec/python-exec2
$ head /usr/lib/python-exec/python-exec2
#!/usr/bin/python-exec2c
# vim:fileencoding=utf-8:ft=python
# (c) 2012-2016 Michał Górny
# Released under the terms of the 2-clause BSD license.
#
# This is not the script you are looking for. This is just a wrapper.
# The actual scripts of this application were installed
# in subdirectories of /usr/lib/python-exec.
# You are most likely looking for one of those.
```
This is done so the system can runtime-switch implementation (like Debian's `update-alternatives`). 

I'm not sure about the best way to work around this here, at least checking the first line of `sphinx-build` is not portable enough.

olifre:
Maybe the "easiest" would be to have a flag to force the assumption that the `sphinx_python` is the same as the system's `python`?

reidpr:
Thank you @olifre! No statue of limitations. I agree that you've found a bug. Shall we discuss a solution in issue #629?

olifre:
Agreed, thanks :+1: .

PR #630, opened 2020-01-27
ch-test: add pedantic mode

reidpr:
This PR adds an option `--pedantic` to `ch-test`. In pedantic mode, we try hard to run all the tests we can, and fail instead of skipping. Pedantic mode defaults to yes if the user is listed in `README.rst` or we are in a CI environment.

One thing we don't insist on is `CH_TEST_PERMDIRS != skip`. I'm not thrilled about this, but some contributors may not have root, and in order to be effective with permissions fixtures, you need root.

### DIFF
--- a/bin/ch-test
+++ b/bin/ch-test
@@ -160,7 +160,7 @@ pedantry_set () {
     if [[ -n $ch_contributor || -n $TRAVIS ]]; then
         default=yes
     fi
-    vset ch_pedantic "$pedantic" '' $default $width 'pedantic mode'
+    vset ch_pedantic "$pedantic" '' $default "$width" 'pedantic mode'
     if [[ $ch_pedantic == no ]]; then
         ch_pedantic=  # proper boolean
     fi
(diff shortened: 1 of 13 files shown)


bin/ch-test
```diff
@@ -383,6 +411,16 @@ if [[ ! -d $CHTEST_EXAMPLES_DIR ]]; then
     fatal "examples not found: $CHTEST_EXAMPLES_DIR"
 fi
 
+# Is the user a contributor? (Obtaining an e-mail address this way is flaky,
+# but it works at Los Alamos, where the team is based.)
+email="$USER@$(hostname --domain)"
```
Brofessional: We should consider using `git config --get user.email` to get the users email address. I would never have pedantic mode enforced, because I usually develop in a VM that doesn't have my expected username or domain. I am likely not the only one in this case. I would expect despite past blunders that every one of us has our git configured to our contributor email in the global config.

reidpr: parks & rec mic drop

reidpr:
Note, this commit adds `$CHTEST_GITWD`, which I thought I'd need for using Git to obtain e-mail, but that turned out not to be true, but I was too lazy to remove it (and it does get used in one place).

PR #632, opened 2020-01-28
ch-test: accept long options with arg separated by equals

reidpr:
I keep typing `ch-test` options with an equals separator (`ch-test --scope=full`) rather than whitespace (`ch-test --scope full`). This makes it work.


reidpr:
Yeah, probably.

reidpr:
So, the man pages aren't consistent in general. I'm going to merge this and deal with the doc problem separately. See #641.

Issue #605, opened 2019-12-20
build: configure should distinguish between "not present" and "too old"


reidpr:
The current configure reports "not found" both for a dependency being too old and not found at all. E.g.:

```
$ buildah --version
buildah version 1.9.0 (image-spec 1.0.0, runtime-spec 1.0.0)
```

then configure reports:

```
Image builders via our wrappers
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  Buildah: no
    Buildah 1.10.1 ... not found
```

which is technically correct but confusing. Further above, the details about what was found is logged, but we should get the report right too.

Follow-up to PR #589. Thanks to @rstyd for reporting.

PR #633, opened 2020-01-29
Make configure distinguish between "not present" and "too old"

rstyd:
Addresses #605. Not sure of the best way to present this information.


rstyd:
The second format looks perfect! I'll work on implementing it.

reidpr:
OK. Make sure you understand the implications w.r.t. the documentation before you start coding.

configure.ac
```diff
@@ -116,12 +116,12 @@ AC_DEFUN([CH_CHECK_VERSION], [
       [AC_MSG_RESULT([ok ($vact)])
        AC_SUBST([$1_VERSION], [$vact])],
       [AC_MSG_RESULT([too old ($vact)])
+	   AC_SUBST([$1_VERSION_OLD], ["too old $vact"]),
```
reidpr: I like this approach, but let's take it a little further to simplify the report. In the current PR, report lines look like:

```
sphinx-build(1) ≥ $vmin_sphinx ... ${SPHINX_VERSION:-not found $SPHINX_VERSION_OLD} ${SPHINX}
```

Let's instead do this:

```
sphinx-build(1) ≥ $vmin_sphinx ... ${SPHINX_VERSION_NOTE}
```

That is, `CH_CHECK_VERSION` sets `FOO_VERSION_NOTE` to one of three values:

* If `FOO` is not found (i.e., `test -n "$prog"` fails), then set `FOO_VERSION_NOTE` to "not found".
* If `FOO` is found and version is OK, then set `FOO_VERSION_NOTE` to "ok (x.y.z)".
* If `FOO` is found and too old, then set `FOO_VERSION_NOTE` to "too old (x.y.z)".

`FOO_VERSION` goes away.

configure.ac
```diff
@@ -414,20 +414,18 @@ Building Charliecloud
 Image builders via our wrappers
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-  Buildah: ${have_buildah}
-    Buildah $vmin_buildah ... ${BUILDAH_VERSION:-not found}
+  Buildah ≥ $vmin_buildah ... ${BUILDAH_VERSION:-not found $BUILDAH_VERSION_OLD} $BUILDAH_VERISON_OLD
```
reidpr: This should stay two lines. Note how you've lost `$have_buildah`.

configure.ac
```diff
@@ -414,20 +414,18 @@ Building Charliecloud
...
     "lark-parser" module ... ${have_lark}
-    skopeo $vmin_skopeo ... ${SKOPEO_VERSION:-not found}
-    umoci $vmin_umoci ... ${UMOCI_VERSION:-not found}
+    skopeo ≥ $vmin_skopeo ... ${SKOPEO_VERSION:-not found $SKOPEO_VERSION_OLD}
+    umoci ≥ $vmin_umoci ... ${UMOCI_VERSION:-not found $UMOCI_VERSION_OLD}
     ch-run(1) ... ${have_ch_run}
 
-  Docker: ${have_docker}
-    Docker $vmin_docker ... ${DOCKER_VERSION:-not found}
+  Docker ≥ $vmin_docker ... ${DOCKER_VERSION:-not found $DOCKER_VERSION_OLD}
```
reidpr: Same here.

configure.ac
rstyd: I really like that approach. I'll implement it.

configure.ac
```diff
@@ -101,6 +101,7 @@ AC_ARG_ENABLE([ch-grow],
 # set $1_VERSION to the actual version.
 #
 # If given absolute path, and that file isn't executable, error out.
+
```
reidpr: 1. Remove this blank line.
2. Update comment to reflect new API.

configure.ac
```diff
@@ -114,10 +115,13 @@ AC_DEFUN([CH_CHECK_VERSION], [
...
       [AC_MSG_RESULT([ok ($vact)])
-       AC_SUBST([$1_VERSION], [$vact])],
+       AC_SUBST([$1_VERSION], [$vact]),
+       AC_SUBST([$1_VERSION_NOTE], ["ok $vact"])],
       [AC_MSG_RESULT([too old ($vact)])
+	   AC_SUBST([$1_VERSION_NOTE], ["too old $vact"]),
        AS_UNSET([$1])])
-  ])
+  ],
+  AC_SUBST([$1_VERSION_NOTE], ["Not found"]))
```
reidpr: Lower case "not found".

configure.ac
```diff
@@ -114,10 +115,13 @@ AC_DEFUN([CH_CHECK_VERSION], [
     vact=$($prog $3)
     AX_COMPARE_VERSION([$2], [le], [$vact],
       [AC_MSG_RESULT([ok ($vact)])
-       AC_SUBST([$1_VERSION], [$vact])],
+       AC_SUBST([$1_VERSION], [$vact]),
```
reidpr: The proposal was to remove `FOO_VERSION`. Would you like to argue to keep it?

configure.ac
```diff
@@ -114,10 +115,13 @@ AC_DEFUN([CH_CHECK_VERSION], [
...
+       AC_SUBST([$1_VERSION_NOTE], ["ok $vact"])],
```
reidpr: I think this and the below would look better with parens added — `ok ($vact)`?

configure.ac
```diff
@@ -114,10 +115,13 @@ AC_DEFUN([CH_CHECK_VERSION], [
     vact=$($prog $3)
     AX_COMPARE_VERSION([$2], [le], [$vact],
       [AC_MSG_RESULT([ok ($vact)])
```
reidpr: For both of these, if you put `AC_MSG_RESULT` second, I think you can use the variable set in `AC_SUBST`, and thereby write the string only once?

configure.ac
```diff
@@ -451,10 +455,10 @@ Building Charliecloud
     C99 compiler ... ${CC} ${CC_VERSION}
 
   documentation: ${have_docs}
-    sphinx-build(1) $vmin_sphinx ... ${SPHINX:-not found} ${SPHINX_VERSION}
+    sphinx-build(1) ≥ $vmin_sphinx ... ${SPHINX_VERSION_NOTE}
     sphinx-build(1) Python ... ${sphinx_python:-n/a}
-    "docutils" module $vmin_docutils ... ${docutils_VERSION:-not found}
-    "sphinx-rtd-theme" module $vmin_rtd ... ${rtd_VERSION:-not found}
+    "docutils" module $vmin_docutils ≥ ... ${docutils_VERSION_NOTE}
+    "sphinx-rtd-theme" module $vmin_rtd ≥ ... ${rtd_VERSION_NOTE}
```
reidpr: "≥" is in the wrong place.

configure.ac
rstyd: Ah. Sorry. `FOO_VERSION` is used by all of the AS_IF lines 
https://github.com/hpc/charliecloud/blob/03f546c7bbd19cdf9ef0a9f96761e37ea6b9119c/configure.ac#L342. 
Can that be done with the version note string? Cause it seems like it expects just a version number and the note has the message as well.

configure.ac
rstyd: Thanks! I had updated this while merging and misplaced it.

configure.ac
rstyd: Figured out how to do this properly with `FOO_VERSION_NOTE`.

reidpr: `AS_IF` can all use plain `$FOO` I think.

rstyd: I modified it from `AS_IF([test -n "$NVIDIA_CLI_VERSION"]` to `AS_IF([test  "$NVIDIA_CLI_VERSION" != "not found"]`

configure.ac
```diff
@@ -99,8 +99,13 @@ AC_ARG_ENABLE([ch-grow],
...
 #
-# If minimum version is less than actual version, unset $1. Otherwise, also
-# set $1_VERSION to the actual version.
+# If minimum version is less than actual version, set $1_VERSION_NOTE 
+# to "too old $version" and unset $1.
+#
+# If actual version is equal to or greater than minimum, set $1_VERSION_NOTE 
+# to "ok $version"
+# 
+# If dependency doesn't exist set $1_VERSION_NOTE to "not found"
```
reidpr: Need to also unset `$1` in this case.

configure.ac
```diff
@@ -349,22 +356,22 @@ AS_IF([   test $have_userns = yes],
 
 # image builders
 
-AS_IF([   test -n "$BUILDAH_VERSION"],
+AS_IF([   test "$BUILDAH_VERSION_NOTE" != "not found" ],
```
reidpr: Change these tests to e.g. `-n "$BUILDAH"`. Because `$BUILDAH_VERSION_NOTE` etc. is a human-readable message, we shouldn't use it in conditionals.

configure.ac
```diff
@@ -376,30 +383,30 @@ AS_IF([   test $have_buildah = yes \
       [have_any_builder=yes],
       [have_any_builder=no])
 
-AS_IF([test $have_any_builder = yes],
+AS_IF([   test $have_any_builder = yes],
       [have_ch_build=yes],
       [have_ch_build=no])
 
-AS_IF([test -n "$TAR"],
+AS_IF([   test "$TAR"],
```
reidpr: `-n` dropped here.

reidpr:
I pushed a couple of commits. The tidy cleans up some things and solves some bugs, in large part due to lack of clarify on my part earlier (e.g., bad base comment for you to work on, confusing indentation in M4 macro). Give them a try; if they look OK, I'll merge.

rstyd:
Just tried it. Looks good!

Issue #486, opened 2019-07-12
ch-test: add option to select packed image format regardless of what is installed


reidpr:

Brofessional:
It actually might be better to weave it into the bats test, that way if your tests are already built you could still force tar without having to rebuild.

PR #634, opened 2020-01-31
add variable to force tar over squashfs during tests

Brofessional:
addresses #486


bin/ch-test
```diff
@@ -481,6 +482,10 @@ while [[ $# -gt 0 ]]; do
         scope_check "$1"
         scope=$1; shift
         ;;
+    --fmt)
+	require_unset archive_fmt
```
reidpr: Indentation error.

bin/ch-test
```diff
@@ -481,6 +482,10 @@ while [[ $# -gt 0 ]]; do
         scope_check "$1"
         scope=$1; shift
         ;;
+    --fmt)
+	require_unset archive_fmt 
+        archive_fmt=$1; shift
```
reidpr: Also support the new equals argument separator (e.g., `--fmt=foo`). If you merge from master, you'll get it for the others and can use that for an example.

bin/ch-test
```diff
@@ -516,6 +521,8 @@ vset CH_TEST_IMGDIR   "$imgdir"        "$CH_TEST_IMGDIR"   /var/tmp/img \
                       21 'unpacked images dir'
 vset CH_TEST_TARDIR   "$tardir"        "$CH_TEST_TARDIR"   /var/tmp/tar \
                       21 'packed images dir'
+vset CH_ARCHIVE_FMT   "$archive_fmt"   "$CH_ARCHIVE_FMT"   squashfs \
```
reidpr: Call it `CH_TEST_PACK_FMT` or `CH_PACK_FMT`? Do you think the variable should be used only for testing, or generally, keeping in mind we'll be moving to your new user-friendly scripts soon?

doc/ch-test_desc.rst
```diff
@@ -122,6 +122,9 @@ Additional arguments:
     Implies :code:`--sudo`. Default: :code:`CH_TEST_PERMDIRS` if set;
     otherwise skip the filesystem permissions tests.
 
+  :code:`--fmt FMT`
+    Force a certain archive format to be used during testing (squashfs, tar)
```
reidpr: Maybe:

```
Use packed image format :code:`FMT` (:code:`squashfs` or :code:`tar`).
```

test/common.bash
```diff
@@ -133,6 +133,9 @@ scope () {
 }
 
 squashfs_ready () {
+    if [ "${CH_ARCHIVE_FMT}" != squashfs ]; then
```
reidpr: In Bash this can be written:

```
if [[ $CH_ARCHIVE_FMT != squashfs ]]; then
```

bin/ch-test
```diff
@@ -35,6 +35,7 @@ Additional options:
   --img-dir DIR          unpacked images directory; same as \$CH_TEST_IMGDIR
   --pack-dir DIR         packed imaged directory; same as \$CH_TEST_TARDIR
   --perm-dir DIR         permissions fixture dir, can repeat; \$CH_TEST_PERMDIRS
+  --fmt FMT              flattened image format (squashfs, tar)
```
reidpr: 1. Call it `--pack-format` or `--pack-fmt`?
2. Refer to it as "packed image" rather than "flattened image" (flattened is the directory).

bin/ch-test
reidpr: The default will need to be what is installed, not simply `squashfs`.

Issue #635, opened 2020-02-03
Charliecloud dependency gpgme for Spack


audreystott:
Hi guys, I posted a Charliecloud-related issue [here](https://github.com/spack/spack/issues/14605) but was wondering if you guys have come across this problem too seeing as it is a dependency of Charliecloud for Spack? Thanks!

j-ogas:
@audreystott Thanks for taking time to bring attention to this issue here and on Spack. Hopefully it will be addressed soon.

If you need an immediate workaround you could try the kludge in PR #636 (i.e., force spack to use the system gpg), and include `^python+libxml2` for the builder, e.g., `spack install charliecloud+builder ^python+libxml2`.

audreystott:
Thanks for the swift response Jordan, what a clever and easy fix!

reidpr:
With the merge of PR #616, we are no longer using skopeo to pull images. Was that the reason for the dependency? Is this issue now moot?

j-ogas:
This issue is now obsolete! Closing.

Issue #600, opened 2019-12-20
Spack test is failing again


reidpr:
Follow-up to PR #589.

j-ogas:
After changing the base image to CentOS 7 in my local branch, I can confirm the Spack package install of charliecloud with the `+builder` variant is still failing.

The error occurs when Spack tries to install expat.
```
==> Installing expat
==> Searching for binary cache of expat
==> No binary for expat found: installing from source
tar (child): lbzip2: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
/usr/bin/tar: Child returned status 2
/usr/bin/tar: Error is not recoverable: exiting now
==> Fetching https://github.com/libexpat/libexpat/releases/download/R_2_2_9/expat-2.2.9.tar.bz2
==> Staging archive: /tmp/root/spack-stage/spack-stage-expat-2.2.9-kbptkkmpipuohhzg5iqexa4eov2cpcav/expat-2.2.9.tar.bz2
==> Error: ProcessError: Command exited with status 2:
    '/usr/bin/tar' '-xf' '/tmp/root/spack-stage/spack-stage-expat-2.2.9-kbptkkmpipuohhzg5iqexa4eov2cpcav/expat-2.2.9.tar.bz2'
The command '/bin/sh -c spack install charliecloud+builder' returned a non-zero code: 1
```

The `tar (child): lbzip2 Cannot exec: No such file or directory` is of special note. I can't seem to install that package manually as the package URLs are broken.

```
$ ~/git/spack/bin/spack install lbzip2
==> Installing lbzip2
==> Searching for binary cache of lbzip2
==> No binary for lbzip2 found: installing from source
==> Fetching http://archive.lbzip2.org/lbzip2-2.5.tar.gz

curl: (22) The requested URL returned error: 503 Service Unavailable
==> Failed to fetch file from URL: http://archive.lbzip2.org/lbzip2-2.5.tar.gz
    URL http://archive.lbzip2.org/lbzip2-2.5.tar.gz was not found!
==> Fetching from http://archive.lbzip2.org/lbzip2-2.5.tar.gz failed.
==> Error: FetchError: All fetchers failed for spack-stage-lbzip2-2.5-kwtegx2tuiuy22rxzngownc7yzaqbnhr

/home/jogas/git/spack/lib/spack/spack/package.py:1110, in do_fetch:
       1107                raise FetchError("Will not fetch %s" %
[...]

PR #636, opened 2020-02-03
update spack example

j-ogas:
WIP.

Addresses #600.

TODO:
- [x] update when 0.14 is merged into spack

outdated comments:

<details>
TODO:

- Get feedback and add commentary from @nicholas-sly regarding undocumented dependencies, e.g., `patch`, `file`, `which`, and `bzip`;  
- Comment on `^python+libxml2`; and
- resolve gpg issue (see comment).

Hopefull this kludge below, related to #635, will be addressed with the [spack issue](https://github.com/spack/spack/issues/14605) reported by @audreyscott (thanks!).
```
# Add system gpg to packages.yaml to avoid the gpg filename too long error. See
# https://github.com/spack/spack/issues/14605
RUN ( echo "  gnupg:" \
 &&   echo "    paths:" \
 &&   echo "      gnupg@2.0.22: /usr" \
 &&   echo "    buildable: false" ) \
       >>/spack/etc/spack/defaults/packages.yaml
```

</details>


examples/spack/Dockerfile
```diff
@@ -59,23 +62,30 @@ RUN spack compiler find --scope system
...
+ &&   echo "    paths:" \
+ &&   echo "      gnupg@2.0.22: /usr" \
+ &&   echo "    buildable: false" ) \
+       >>/spack/etc/spack/defaults/packages.yaml
 
 # Test: Install Charliecloud.
-RUN spack spec charliecloud+builder
-RUN spack install charliecloud+builder
+RUN spack spec charliecloud+builder ^python+libxml2
+RUN spack install charliecloud+builder ^python+libxml2
```
j-ogas: This should be addressed in the charliecloud package.

@nicholas-sly can you comment on why we need to add the `^python+libxml2` dependency?

nicholas-sly: This is a factor of a circular dependency resolution between libxml2 and python, I believe.  The hope is that the new concretizer (still a couple months away) will fix this issue and remove that variant, but for now, it is necessary to build.

nicholas-sly:
The dependencies added are a mixture of requirements for Spack and diagnosis tools.  `patch` is necessary for Spack to install any packages that utilize patches during installation.  `which` may be used by Spack to determine the paths to binaries when loaded in using modules or elsewhere in Spack.  It may have just been used for diagnosis.  `file` may also be used by Spack, but may have just been diagnosis.  

For the above dependencies that 'may' be used by Spack, the best way to check is to remove them from the `yum install` lines and see if the container builds.

`bzip` along with the lines `echo`'d to `/spack/etc/spack/defaults/packages.yaml` are an attempt to just use the yum-provided bzip package with the appropriate lbzip2 library.  While Spack does have an lbzip2 package, it is broken in the develop branch and appears to be pretty unmaintained.  I have successfully fixed this on a local version to enable installing the lbzip2 package, but properly modifying the package dependencies up to charliecloud is taking further development (something like charliecloud -> expat -> tar -> bzip2 -> lbzip2).  Once that is merged into Spack, hopefully installing it only with Spack and without the yum-provided package will work.  Until then, using the "system"-provided version of the package seems to work.

reidpr:
Thanks for your analysis, Nick! That's really helpful.

j-ogas:
This PR will need to be updated once we have 0.14 is [merged](https://github.com/spack/spack/pull/15573#pullrequestreview-377980227) into spack. We won't need to mess with the gpgme or`^python+libxml2` kludges.

examples/spack/Dockerfile
```diff
@@ -1,6 +1,5 @@
-# FIXME: was full; doesn't work after autotools update
-# ch-test-scope: skip
-FROM debian9
+# ch-test-scope: full
+FROM centos:7
```
reidpr, line 2: This is the upstream CentOS 7 image. Should it be `centos7` to be ours, or `centos:8` or `centos8`?

examples/spack/Dockerfile
```diff
@@ -14,22 +13,26 @@ FROM debian9
 # [1]: https://spack.readthedocs.io/en/latest/getting_started.html
 # [2]: https://spack.readthedocs.io/en/latest/workflows.html#using-spack-to-create-docker-images
 
-# Packages needed to install Spack [1].
-#
```
reidpr, line 18: Why remove this comment?

examples/spack/Dockerfile
```diff
@@ -59,24 +62,16 @@ RUN spack compiler find --scope system
...
 RUN spack compiler list
 RUN spack compiler list --scope=system
 RUN spack compiler list --scope=user
 RUN spack compilers
-RUN spack spec netcdf
+RUN spack spec charliecloud
 
 # Test: Install Charliecloud.
-RUN spack spec charliecloud+builder
-RUN spack install charliecloud+builder
```
reidpr, line 70: Did we not want to test installing `ch-grow` any more with Spack?

examples/spack/Dockerfile
```diff
@@ -59,24 +62,16 @@ RUN spack compiler find --scope system
...
-
-# Test: Run Spack test suite. Distro package dependencies discovered by trial
-# and error.
-#
-# FIXME: Commented out because the suite fails. Last time I tried, 6 tests
-# failed and 1 "xfailed".
-#RUN apt-get install -y \
-#    python-pkg-resources
-#RUN spack test
```
reidpr, line 79: So this wasn't doing anything because it was commented out. Do we want to retain the documentation that the Spack test suite doesn't work?

examples/spack/Dockerfile
j-ogas: The builder variant has been removed. You always get ch-grow now :).

examples/spack/Dockerfile
j-ogas: Couple of thoughts: 1) it has been a very long time (2+ years?) since I personally tried to run spack unit tests in a container; and 2) even if they did work, I'm not so sure there is much value in including them in this example. This image takes quite a long time to build, and working unit tests would make it worse.

I would argue that unless we think we have something to gain by including them in the example, we don't need to comment on them.

examples/spack/Dockerfile
j-ogas: This appears to be a mistake. Good catch.

examples/spack/Dockerfile
j-ogas: I suppose our centos 7 and 8 images are a bit bulkier with the added sauce needed for openmpi and rpm building. My initial thought was to use a fresh upstream CentOS image but I don't think I have a strong opinion one way or another.

Issue #629, opened 2020-01-27
can't deduce Sphinx Python on Gentoo, maybe others


reidpr:
@olifre pointed out in comments on PR #624:

> [T]his PR broke the build on Gentoo Linux (and potentially others). 
The problem is that:
> ```
> sphinx_python=$(head -1 "$SPHINX" | sed -E 's/^..//')
> ```
> does not work. On Gentoo, any python tool which is installed with multiple implementations is wrappered:
> ```
> $ ls -la $(which sphinx-build)
> lrwxrwxrwx 1 root root 31  4. Jan 18:44 /usr/bin/sphinx-build -> ../lib/python-exec/python-exec2
> $ head /usr/lib/python-exec/python-exec2
> #!/usr/bin/python-exec2c
> # vim:fileencoding=utf-8:ft=python
> # (c) 2012-2016 Michał Górny
> # Released under the terms of the 2-clause BSD license.
> #
> # This is not the script you are looking for. This is just a wrapper.
> # The actual scripts of this application were installed
> # in subdirectories of /usr/lib/python-exec.
> # You are most likely looking for one of those.
> ```
> This is done so the system can runtime-switch implementation (like Debian's `update- alternatives`). 
> More details at:
> https://wiki.gentoo.org/wiki/Project:Python/python-exec
>
> I'm not sure about the best way to work around this here, at least checking the first line of `sphinx-build` is not portable enough. 

and also:

> Maybe the "easiest" would be to have a flag to force the assumption that the `sphinx_python` is the same as the system's `python`?

reidpr:
@olifre, would it work to provide a `configure` option `--with-sphinx-python` that takes the path to the proper Python interpreter?

olifre:
@reidpr This would work. By design, on Gentoo, an implementation is chosen at runtime (can be chosen system-wide, but also by the user, or even switched via environment variable). This only works for packages installed for the appropriate target, otherwise a fallback is made. 
Of course, at build-time, the implementation which is used is known, so I can pass the used interpreter as argument to {{configure}} without a problem. By design, it will always be the same implementation used for the other build steps (package dependencies ensure that sphinx and themes are build against the same implementation(s)). So that sounds like a good solution!

Issue #611, opened 2020-01-08
autogen.sh: provide suggestions for incomprehensible errors


reidpr:
Currently, if you lack one or more of the Autotools dependencies, the errors you get from `autogen.sh` make no sense. E.g. for missing Autoconf-Archive:

```
$ ./autogen.sh
+ rm -rf Makefile Makefile.in '*/Makefile' '*/Makefile.in' aclocal.m4 autom4te.cache bin/.deps bin/config.h bin/config.h.in bin/stamp-h1 build-aux config.log config.status configure
+ [[ '' != --clean ]]
+ aclocal
+ autoheader
+ autoreconf --install -Wall -Werror
configure.ac:35: error: possibly undefined macro: AC_MSG_ERROR
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:51: error: possibly undefined macro: AC_SUBST
configure.ac:66: error: possibly undefined macro: AC_MSG_RESULT
configure.ac:119: error: possibly undefined macro: AS_UNSET
autoreconf: /usr/bin/autoconf failed with exit status: 1
```

We should provide a tip at the top saying if you get incomprehensible errors, check such-and-such documentation.

Issue #643, opened 2020-02-13
docs do not build in parallel


Trophime:
Hi,
trying to upgrade the charliecloud package for Debian I run into this error:
```
configure.ac:35: error: possibly undefined macro: AC_MSG_ERROR
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:51: error: possibly undefined macro: AC_SUBST
configure.ac:66: error: possibly undefined macro: AC_MSG_RESULT
configure.ac:119: error: possibly undefined macro: AS_UNSET
autoreconf: /usr/bin/autoconf failed with exit status: 1
```

wiene:
I think this happens if the `autoconf-archive` package is missing.

Trophime:
Thanks it help but now I've got this error:
```
cat ../config.log | ./make-deps-overview > _deps.rst
make[3]: *** No rule to make target 'html/searchindex.js', needed by 'all-am'.  Stop.
make[3]: *** Waiting for unfinished jobs....
make[3]: Leaving directory '/build/charliecloud-0.13/doc'
make[2]: *** [Makefile:387: all-recursive] Error 1
make[2]: Leaving directory '/build/charliecloud-0.13'
dh_auto_build: make -j8 returned exit code 2
```

reidpr:
Thanks @Trophime, really pleased to hear that.

Is there somewhere we can see the complete build log, including configure output?

Trophime:
What is called:
```
```
	./configure --build=x86_64-linux-gnu --prefix=/usr --includedir=\${prefix}/include --mandir=\${prefix}/share/man --infodir=\${prefix}/share/info --sysconfdir=/etc --localstatedir=/var --disable-silent-rules --libdir=\${prefix}/lib/x86_64-linux-gnu --runstatedir=/run --disable-maintainer-mode --disable-dependency-tracking
```
[charliecloud_0.13-1_amd64.build.log](https://github.com/hpc/charliecloud/files/4199275/charliecloud_0.13-1_amd64.build.log)

reidpr:
Thank you @trophime! We will take a look. I develop on a Buster box, so I'll see if I can reproduce the issue.

reidpr:
p.s. I reformatted your comment a little for clarity.

Trophime:
I think I have found the issue.
To reproduce the problem:

make -j 2 -C libexec
make -j 2 -C doc

[...]

PR #659, opened 2020-02-21
autotools: add --with-sphinx-python, better docs

reidpr:
Addresses #611 and #629.

@olifre, does this fix the bug you reported in #629?

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -8,13 +8,29 @@ Usage:
 
   $ ./autogen.sh [--clean]
 
-Remove and rebuild Autotools files (./configure and friends).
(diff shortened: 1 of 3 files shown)


reidpr:
Great! We'll wait a little while to see if it fixes @olifre's bug before merging.

olifre:
@reidpr Indeed, it does :-). Just using:
```
--with-sphinx-python=$(which python)
```
now "does the right thing" in Gentoo (since the /usr/bin/python interpreter is a symlink to the underlying `python-exec2c` wrapper which then does the right thing, i.e. use the selected Python version). 
I'm now running into #643 and will follow up there. Thanks!

Issue #646, opened 2020-02-14
ch-test fix run and example phase image filtering.


j-ogas:
ch-test `run` and `example` phases should populate a list of tests to run based on what is found in`--pack-dir` and filter according to architecture and scope. Builder requirements should be ignored (run system may have no builder at all).

PR #661, opened 2020-02-22
ch-test improvements

j-ogas:
Addresses #646.


bin/ch-test
```diff
@@ -111,12 +111,20 @@ builder_set () {
 dir_img_mk () {
     dir_img_rm
     printf "creating %s\n\n" "$CH_TEST_IMGDIR"
-    mkdir "$CH_TEST_IMGDIR"
+    if [[ -n $SLURM_JOBID ]]; then
+        srun --ntasks-per-node=1 mkdir "$CH_TEST_IMGDIR"
+    else
+        mkdir "$CH_TEST_IMGDIR"
+    fi
```
j-ogas: The unpack assumptions are not squashfs friendly.

bin/ch-test
```diff
@@ -297,40 +305,102 @@ test_run () {
...
+                    if [[ $(test_make_tag_from_path "$j") == "$artifact" ]]; then
+                        if (test_make_check_image "$j"); then
+                            echo "found: $i"
+                            run_targets+=( "$j" )
+                        fi
+                    fi
+                done
+            fi
+        done
+        printf '\n'
```
j-ogas: For each artifact found in `--pack-dir` look for a test artifact that produces a matching tag. If we find one, then check the image for exclusion conditions.

bin/ch-test
reidpr: Are they? Creating and removing `$CH_TEST_IMGDIR` will eventually (not now) be redundant for SquashFS, but it won't be harmful.

reidpr: I think this if statement can be replaced with `$ch_mpirun_node`.

bin/ch-test
```diff
@@ -111,12 +111,20 @@ builder_set () {
...
 }
 
 dir_img_rm () {
     echo "removing $CH_TEST_IMGDIR"
-    rm -rf --one-file-system "$CH_TEST_IMGDIR"
+    if [[ -n $SLURM_JOBID ]]; then
+        srun --ntasks-per-node=1 rm -rf --one-file-system "$CH_TEST_IMGDIR"
+    else
+        rm -rf --one-file-system "$CH_TEST_IMGDIR"
+    fi
```
reidpr: Especially now that it's going through `srun`, good I think to validate that `$CH_TEST_IMGDIR` and `$CH_TEST_TARDIR` are what we think they are, or empty, before `rm -rf`ing them.

bin/ch-test
```diff
@@ -152,25 +160,39 @@ dir_tar_check_file () {
...
+    if [[ -n $SLURM_JOBID ]]; then
+        srun --ntasks-per-node=1 rm -rf --one-file-system "$CH_TEST_TARDIR"
+    else
+        rm -rf --one-file-system "$CH_TEST_TARDIR"
+    fi
+}
+
+dir_tmp_rm () {
+    echo "removing $TMP_"
+    rm -rf --one-file-system "$TMP_"
```
reidpr: Same — are we sure it's our `$TMP_`?

bin/ch-test
```diff
@@ -297,40 +305,102 @@ test_run () {
...
+        done
+        printf '\n'
         printf 'generate build_auto.bats ...\n'
-        "$make_auto" build "${targets[@]}" > "${TMP_}/build_auto.bats"
+        test_make_auto "$phase" "${build_targets[@]}" > "${TMP_}/build_auto.bats"
         printf 'ok\n\n'
         ;;
     run)
+        for i in "${CH_TEST_TARDIR}"/*{.tar.gz,.sqfs}; do
+            if [[ -e $i ]]; then
```
reidpr: Maybe better to check that you got the `*` back from the glob than whether the file exists? I'm not sure what's more idiomatic.

bin/ch-test
```diff
@@ -297,40 +305,102 @@ test_run () {
...
+                artifact=${i##*/}
```
reidpr: I'd call it something related to "packed image" rather than "artifact" — more specific.

bin/ch-test
reidpr: Make this a comment. :)

bin/ch-test
```diff
@@ -697,28 +812,45 @@ scope_int=$(scope_to_digit "$CH_TEST_SCOPE")
...
         dir_tar_mk
-        test_make_auto
+        test_make build
         test_build
         win
         ;;
     run)
         dir_img_mk
-        test_make_auto
+        test_make
```
reidpr: It might be clearer for `test_make` to always or never take an argument.

j-ogas: That was a bug, good catch.

bin/ch-test
j-ogas: `$ch_mpirun_node` is undefined here at the ch-test execution. I suppose we could source common.bash like the bat files; thoughts?

bin/ch-test
j-ogas: Any suggestions how to do this effectively without creating situations where the check produce false negatives? I would like to avoid situations where `ch-test clean` doesn't do anything because the check fails due to missing artifacts caused by forced test-suite cancel.

bin/ch-test
Brofessional: it appears it could be used multiple times. I think sourcing it would be helpful.

bin/ch-test
reidpr: Maybe a sentinel file? Or, if the directory is empty, we also remove it, since that's low cost.

test/common.bash
```diff
@@ -113,7 +113,10 @@ prerequisites_ok () {
 
 # Wrapper for Bats run() to work around Bats bug #89 by saving/restoring $IFS.
 # See issues #552 and #555 and https://stackoverflow.com/a/32425874.
-eval bats_"$(declare -f run)"
+# shellcheck disable=SC2091
+if $(declare -f run &> /dev/null); then
```
reidpr: I think here you want to test with `if ( type -t run > /dev/null 2>&1 )` rather than `declare`: https://stackoverflow.com/a/85903 (see first comment).

bin/ch-test
```diff
@@ -694,31 +813,48 @@ examples=$(find "$CHTEST_EXAMPLES_DIR"    -name 'Build' \
 
 scope_int=$(scope_to_digit "$CH_TEST_SCOPE")
 
+. "${CHTEST_DIR}/common.bash"
+
```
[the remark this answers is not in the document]
j-ogas: I can move it up a little bit, however, it will cause ch-test to exit with `CH_TEST_*` not set if sourced before CH_TEST_* variables are processed.

bin/ch-test
```diff
@@ -111,12 +111,18 @@ builder_set () {
...
+    $ch_mpirun_node mkdir "$CH_TEST_IMGDIR"
+    $ch_mpirun_node touch "$CH_TEST_IMGDIR/.sentinel"
 }
 
 dir_img_rm () {
+    if find "$CH_TEST_TARDIR" -mindepth 1 &> /dev/null | read -r; then
+        if [[ ! -e $CH_TEST_TARDIR/.sentinel ]]; then
+            fatal "$CH_TEST_IMGDIR missing sentinel file; manually delete."
+        fi
+    fi
```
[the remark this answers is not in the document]
j-ogas: The thought here was that we want to delete the directory if one of either cases was true: 1) directory is empty; or 2) directory is not-empty but has a sentinel file.

bin/ch-test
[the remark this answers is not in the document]
reidpr: What if you put it right under that `vset` stuff with a comment along the lines of "now that we have `CH_TEST_*` set, we can source `common.bash`?

bin/ch-test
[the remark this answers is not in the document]
reidpr: Aha! Put in a comment to that effect. In that case, join the clauses with `&&`.

bin/ch-test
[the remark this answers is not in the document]
j-ogas: It appears it needs to be below the BATS_TMPDIR check as common.bash changes this value; `$btnew`.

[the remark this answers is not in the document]
reidpr: OK. I recommend you stop listening to me and use your best judgement then. :) Just add a comment saying why it's where it is.

bin/ch-test
```diff
@@ -411,6 +489,60 @@ test_make_check_image () {
     fi
 }
 
+test_make_tag_from_path () {
```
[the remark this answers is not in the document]
j-ogas: I gave it a shot, not sure if it helps. See new commit.

Issue #563, opened 2019-10-01
Update images currently using debian 9 to CentOS 8


heasterday:
This should result in images that are closer to the environments we use and hopefuly work around glibc incompatibility issues we have seen with ch-fromhost. This is pending the release of an official CentOS 8 base docker image we can use.

Tracking issue: https://github.com/CentOS/sig-cloud-instance-images/issues/147

heasterday:
Progress is being posted via internal issue 145.

PR #665, opened 2020-02-24
Update images currently using debian 9 to CentOS 8

heasterday:
Addresses #563


heasterday:
Looks like the Travis failures are legit, I'll resolve.

reidpr:
I think you just need to adjust `Makefile.am` in the relevant directories, if you haven't figured this out yet.

heasterday:
I hadn't had time to look at it yet so I appreciate the suggestion. Looks like that resolved it.

heasterday:
Passed the tests in full outside of the failures reported in #677 and #678 which stem from the environment not contents of this PR.

examples/Dockerfile.centos8
```diff
@@ -0,0 +1,6 @@
+# ch-test-scope: standard
+FROM centos:8
+# Install the dnf ovl plugin to workaround rpmdb corruptions issues when using
+# overlayfs as the storage driver.
+RUN dnf -y install https://download.fedoraproject.org/pub/epel/playground/8/Everything/x86_64/os/Packages/d/dnf-plugin-ovl-0.0.3-1.epel8.playground.noarch.rpm
```
reidpr: Is this in EPEL and we can install it from there? Generally speaking, EPEL is super useful, so why don't we install it first.

examples/Dockerfile.exhaustive
```diff
@@ -21,7 +21,7 @@ ENV chse_4=value4 chse_5="value5 foo" chse_7=\"value7\"
 #ENV chse_4=value4 chse_5="value5 foo" chse_6=value6\ foo chse_7=\"value7\"
 
 RUN pwd
-WORKDIR /usr/local/src
+WORKDIR /usr/src
```
reidpr: A lot of the examples are changed to use prefix `/usr` instead of `/usr/local`. Why is that?

examples/Dockerfile.openmpi
```diff
@@ -58,88 +58,75 @@ FROM debian9
...
-#   https://packages.debian.org/buster/libpsm2-2
-#   https://lists.debian.org/debian-hpc/2017/12/msg00015.html
-ENV DEB_URL http://snapshot.debian.org/archive/debian/20181126T030749Z/pool/main
-ENV PSM2_VERSION 11.2.68-3
-RUN if [ "$(dpkg --print-architecture)" = "amd64" ] ; then \
-      wget -nv ${DEB_URL}/libp/libpsm2/libpsm2-2_${PSM2_VERSION}_amd64.deb \
-               ${DEB_URL}/libp/libpsm2/libpsm2-dev_${PSM2_VERSION}_amd64.deb ; \
+# Note that libpsm2 is x86-64 only
+RUN if [ "$(uname -i)" = "x86_64" ] ; then \
+      dnf install -y --enablerepo=PowerTools libpsm2 libpsm2-devel ; \
```
reidpr: What is PowerTools and how trustworthy is it?

examples/hello/Dockerfile
```diff
@@ -1,9 +1,10 @@
 # ch-test-scope: standard
-FROM debian:stretch
+# ch-test-builder-exclude: ch-grow  #see issues #472 and #498
```
reidpr: This example needs to still work with `ch-grow`. You could comment that we haven't converted it to CentOS for that reason, though.

test/run/ch-fromhost.bats
```diff
@@ -109,30 +109,33 @@ fromhost_ls () {
...
-@test 'ch-fromhost (CentOS)' {
+@test 'ch-fromhost (Debian)' {
     scope full
-    prerequisites_ok centos7
-    img=${ch_imgdir}/centos7
+    prerequisites_ok debian9
+    img=${ch_imgdir}/debian9
+
+    libpath=$(ch-fromhost --lib-path "$img")
+    echo "libpath: ${libpath}"
```
reidpr: I think I would leave this hardcoded in both actually, to make our expectations clear. If that path shuffles around, we should know.

examples/Dockerfile.centos8
heasterday: It is in EPEL in the "epel-playground" repository. Since up until the lustre tests we didn't pull anything from EPEL I didn't want to increase the size of resulting images for something we weren't using. If you feel it's good to have regardless I'll install it.

examples/Dockerfile.exhaustive
heasterday: In CentOS the linker doesn't search `/usr/local/lib` by default like Debian does so there were two options that I saw:
1. Add `/usr/local` to search path by adding something like `/etc/ld.so.conf.d/usrlocal.conf`
2. Move all of the examples to `/usr` instead

When we previously discussed this (a long time ago) we agreed that it was probably best to just move the examples.

examples/Dockerfile.openmpi
heasterday: It's an additional developer repository included in the base image. I can't speak for how reliable it is but I assumed that if it is included in the base image it was probably reliable enough.

According to the [fedora wiki](https://fedoraproject.org/wiki/EPEL)
>EPEL packages assume that the 'PowerTools' repository is enabled. You can do this with: dnf config-manager --set-enabled PowerTools

examples/hello/Dockerfile
heasterday: Sounds good, I will revert.

test/run/ch-fromhost.bats
heasterday: I'll change to hard coded paths.

examples/Dockerfile.centos8
reidpr: I think we should just go ahead and enable EPEL and PowerTools as one of the first steps. CentOS is way more usable with it.

That said, does `dnf` work well enough on overlayfs to enable EPEL and install that plugin?

examples/Dockerfile.exhaustive
reidpr: Aha. I wonder now if that was the right decision. For example, on another image I'm installing GCC from source into `/usr/local`. It seems to cause a bit of a clash with the system GCC, but at least it's in a separate place and doesn't overwrite a random subset of the system GCC. With the new GCC in `/usr/local`, I could in principle delete the system GCC, and it wouldn't interfere with the new one.

examples/Dockerfile.openmpi
reidpr: Sounds good.

examples/Dockerfile.exhaustive
heasterday: I'll change back to `/usr/local/` and add that location to the search path.

heasterday:
@reidpr I incorporated your suggestions and am retesting. Let me know if you have additional things.

heasterday:
Encountered a linking issue in testing, removing the "ready to merge?" label until resolved.

heasterday:
Resolved

examples/Dockerfile.centos8
```diff
@@ -0,0 +1,13 @@
+# ch-test-scope: standard
+FROM centos:8
+# Install the dnf ovl plugin to workaround rpmdb corruptions issues when using
+# overlayfs as the storage driver.
+# Enable the PowerTools repository as some packages in EPEL depend on it.
+RUN dnf -y install epel-release && dnf -y update \
+ && dnf -y install --enablerepo=epel-playground dnf-plugin-ovl \
+ && sed -i -e 's/enabled=0/enabled=1/g' /etc/yum.repos.d/CentOS-PowerTools.repo
```
reidpr: 1. Format as:

```
RUN    dnf -y install epel-release \
    && dnf -y update \
    && dnf -y install --enablerepo=epel-playground dnf-plugin-ovl \
    && sed -i -e 's/enabled=0/enabled=1/g' /etc/yum.repos.d/CentOS-PowerTools.repo
```

(There are several places this general change should be made.)

2. Is it more idiomatic to say `dnf config-manager --set-enabled ...`?

examples/Dockerfile.openmpi
```diff
@@ -58,62 +58,37 @@ FROM debian9
...
-#
-# Note that libpsm2 is x86-64 only:
-#   https://packages.debian.org/buster/libpsm2-2
-#   https://lists.debian.org/debian-hpc/2017/12/msg00015.html
-ENV DEB_URL http://snapshot.debian.org/archive/debian/20181126T030749Z/pool/main
-ENV PSM2_VERSION 11.2.68-3
-RUN if [ "$(dpkg --print-architecture)" = "amd64" ] ; then \
-      wget -nv ${DEB_URL}/libp/libpsm2/libpsm2-2_${PSM2_VERSION}_amd64.deb \
-               ${DEB_URL}/libp/libpsm2/libpsm2-dev_${PSM2_VERSION}_amd64.deb ; \
+# Note that libpsm2 is x86-64 only
```
reidpr: Recommend simply "libpsm2 is x86-64 only".

test/run/ch-fromhost.bats
```diff
@@ -29,26 +29,23 @@ fromhost_ls () {
...
 
     # --file
     fromhost_clean "$img"
     ch-fromhost -v --file sotest/files_inferrable.txt "$img"
     fromhost_ls "$img"
     test -f "${img}/usr/bin/sotest"
-    test -f "${img}${libpath}/libsotest.so.1.0"
-    test -L "${img}${libpath}/libsotest.so.1"
+    test -f "${img}/lib64/bind9-export/libsotest.so.1.0"
+    test -L "${img}/lib64/bind9-export/libsotest.so.1"
```
reidpr: That's a really bizarre place to put it. I wonder if we should put a comment acknowledging that and explaining why.

examples/Dockerfile.openmpi
heasterday: Good catch, I missed the dated comment.

examples/Dockerfile.centos8
heasterday: 1. Will do

2. Turns out `config-manager` is part of the `dnf-plugins-core` package and not available by default. Since this is the only place we would use it I opted for the sed line instead. If you prefer installing that package and then using config manager more I will adjust.

test/run/ch-fromhost.bats
heasterday: I completely agree. I feel that not much is gained by hardcoding a path versus using `${libpath}` except we get to track whatever package happens to include a search path that the linker checks first. Could you elaborate on why you think this is something we should test?

examples/Dockerfile.centos8
reidpr: No, `sed` is good.

test/run/ch-fromhost.bats
reidpr: I thought it would get `/usr/local/lib`. Maybe that is true now with my tidy? But maybe we should just respect `ch-fromhost --lib-path` (which I think is now `--_lib-path`?).

heasterday:
Thanks for the tidy. 

I feel like removing the CentOS 7 image should be in a separate issue as I would like to get this merged relatively soon so folks can start using the updated images. #695 sounds like a good place to do this.

test/run/ch-fromhost.bats
heasterday: Looks like post "tidy" it is still getting `..../bind9-export`. I will revert to `${libpath}`.

heasterday:
@reidpr Still testing but if all that passes are we good to merge on this one?

reidpr:
I think so. Let me give it another once-over since I feel confused about everything ATM and it will make me feel better.

reidpr:
Note: The GitHub merge note above is an error. I merged manually and the true merge commit is 4fd93dc.

PR #667, opened 2020-02-25
doc: fix parallel build

reidpr:
Addresses #643.

@trophime, @olifre: Does this fix the bug for you?


olifre:
That seems to fix it, even with very high `-j12` for me. Thanks :+1: .

Issue #594, opened 2019-12-20
fix RPM build


reidpr:
As a result of PR #589, RPMs no longer build.

Issue #535, opened 2019-09-05
integrate fedora spec file with our build script


j-ogas:
Our official fedora [spec file](https://src.fedoraproject.org/rpms/charliecloud/blob/master/f/charliecloud.spec) has diverged from our initial `packaging/fedora/charliecloud.spec` spec file.

PR #670, opened 2020-02-25
add autoconf fedora package

j-ogas:
WIP.

Addresses #535, #594.

TODO:
- [x] fix the following rpmlint issues below
- [x] convert to package template style
- [x] unskip package test
- [x] update `build` script
- [x] add centOS 8 (el8)
- [x] make travis happy
- [x] patch ch-test for el7 `%{_pkgdocdir}` (/usr/share/doc/%{name}-%{version})


j-ogas:
This is in a reviewable state. Note that epel7 wasn't removed as we can't completely build charliecloud out on epel8. Epel8 doesn't have a bats or bats-core rpm package thus we can't install `charliecloud-test`.

packaging/fedora/charliecloud.spec
```diff
@@ -46,53 +29,120 @@ a standard Linux filesystem tree.
...
-%make_install PREFIX=%{_prefix}
+%make_install
 
 cat > README.EL7 <<EOF
 For RHEL7 you must increase the number of available user namespaces to a non-
 zero number (note the number below is taken from the default for RHEL8):
 
   echo user.max_user_namespaces=3171 >/etc/sysctl.d/51-userns.conf
-  systemctl -p
+  reboot
```
reidpr: I think this should read `sysctl -p` ([man page](http://man7.org/linux/man-pages/man8/sysctl.8.html)); a reboot is not required. `systemctl -p` was indeed incorrect and means something else.

test/run/build-rpms.bats
```diff
@@ -1,19 +1,30 @@
...
+        skip 'Sphinx is not installed'
     fi
+}
+
+@test 'build/install epel7 RPMs' {
+    prerequisites_ok centos7
     img=${ch_imgdir}/centos7
+    image_ok "$img"
+    if [[ -d ${BATS_TMPDIR}/rpmbuild/SOURCES/charliecloud ]]; then
+        rm -rf "${BATS_TMPDIR}/rpmbuild/SOURCES/charliecloud"
```
reidpr: Add `--one-file-system` for a little extra safety?

test/run/build-rpms.bats
```diff
@@ -1,19 +1,30 @@
...
     # Build and install RPMs into CentOS 7 image.
     (cd .. && packaging/fedora/build --install --image="$img" \
                                      --rpmbuild="$BATS_TMPDIR/rpmbuild" HEAD)
+}
 
+@test 'check epel7 RPM files' {
+    prerequisites_ok centos7
+    img=${ch_imgdir}/centos7
     # Do installed RPMs look sane?
     run ch-run "$img" -- rpm -qa "charliecloud*"
```
reidpr: Check for doc package too?

test/run/build-rpms.bats
```diff
@@ -54,4 +74,63 @@ load ../common
     echo "$output"
     [[ $status -eq 0 ]]
     [[ $output = '' ]]
+    rm -rf "${BATS_TMPDIR}/rpmbuild"
```
reidpr: Is this needed given the `rm` above?

test/run/build-rpms.bats
```diff
@@ -54,4 +74,63 @@ load ../common
...
+    # Build and install RPMs into CentOS 8 image.
+    (cd .. && packaging/fedora/build --install --image="$img" \
+                                     --rpmbuild="$BATS_TMPDIR/rpmbuild" HEAD)
+}
+
+@test 'check epel8 RPM files' {
+    prerequisites_ok centos8
+    img=${ch_imgdir}/centos8
+    # Do installed RPMs look sane?
+    run ch-run "$img" -- rpm -qa "charliecloud*"
```
reidpr: Probably should have a new issue to add `charliecloud-test` once EPEL8 gets its act together on Bats.

test/run/build-rpms.bats
```diff
@@ -54,4 +74,63 @@ load ../common
...
+    if [[ -d ${BATS_TMPDIR}/rpmbuild/SOURCES/charliecloud ]]; then
+        rm -rf "${BATS_TMPDIR}/rpmbuild/SOURCES/charliecloud"
+    fi
+
+    # Build and install RPMs into CentOS 8 image.
+    (cd .. && packaging/fedora/build --install --image="$img" \
+                                     --rpmbuild="$BATS_TMPDIR/rpmbuild" HEAD)
+}
+
+@test 'check epel8 RPM files' {
```
reidpr: Are these two tests close enough that they could be put into functions and defined once?

examples/Dockerfile.centos8
```diff
@@ -18,3 +21,23 @@ RUN    echo "/usr/local/lib" > /etc/ld.so.conf.d/usrlocal.conf \
...
+           python3-sphinx \
+           python3-sphinx_rtd_theme \
+           rpm-build \
+           rpmlint \
+           rsync \
+           wget
+# CentOS8 doesn't have a bats package, thus we install from source
+RUN wget https://github.com/sstephenson/bats/archive/v0.4.0.tar.gz \
+    && tar xf v0.4.0.tar.gz \
+    && bats-0.4.0/install.sh /usr
```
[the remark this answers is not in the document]
j-ogas: Agreed, however this kludge was an attempt to get rpmbuild to think it could satisfy the `/usr/bin/bats` dependency. It didn't work, thus should be removed.

packaging/fedora/build
```diff
@@ -75,11 +92,16 @@ def main():
...
-   cmd("make", "-j2", cwd=git_tmp+"/doc-src")
    print("# building source tarball")
-   cmd("make", "export", cwd=git_tmp)
-   ch_version = open(git_tmp + "/VERSION.full").read()[:-1]
+   cmd(CH_RUN, "-b", "%s" % git_tmp, "-c", "/mnt/0", args.image,
+       "--", "./autogen.sh")
+   cmd(CH_RUN, "-b", "%s" % git_tmp, "-c", "/mnt/0", args.image,
+       "--", "./configure")
+   cmd(CH_RUN, "-b", "%s" % git_tmp, "-c", "/mnt/0", args.image,
+       "--", "make")
```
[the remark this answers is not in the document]
j-ogas: If I don't do make dist without make we get the following:
``` 
IOError: [Errno 2] No such file or directory: '/tmp/ch-test.tmp.jogas/bats.tmp/rpmbuild/SOURCES/charliecloud/lib/version.txt'
```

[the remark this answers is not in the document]
reidpr: Hmm, that's a bug.

test/run/build-rpms.bats
j-ogas: Not really. I tried and ended up with a function that had if statements to avoid checking for the charliecloud-test package on el8.

packaging/fedora/build
```diff
@@ -162,16 +185,22 @@ def main():
...
       print("# installing")
       for p in PACKAGES:
-         cmd(container, "rpm", "--install",
-             "%s/%s-%s.*.rpm" % (rpms, p, rpm_vr))
+         if p == 'charliecloud-doc':
+            file_ = "%s/%s-%s.%s.noarch.rpm" % (rpms_noarch, p, rpm_vr, rpm_dist)
+         else:
+            file_ = "%s/%s-%s.%s.%s.rpm" % (rpms, p, rpm_vr, rpm_dist, ARCH)
+         cmd(container, "rpm", "--install", file_)
       cmd(container, "rpm", "-qa", "charliecloud*")
```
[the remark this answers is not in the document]
j-ogas: Yes. This is because I need to bind `git_tmp` at `/mnt/0` and instantiate the process there. Since `git_tmp` is removed after the configure step, including `-b %s --cd /mnt/0` in the container list would cause a failure here.

I didn't see a clean way to merge both use cases into one variable, so I left `container`  alone and just used CH_RUN above.

packaging/fedora/build
```diff
@@ -19,25 +19,29 @@ import time
...
-# version:            %(version)s
-# image:              %(image)s
-# install:            %(install)s
-# rpmbuild root:      %(rpmbuild)s""" % args.__dict__)
+  # print("""\
+# version:            %(args.version)s
+# image:              %(args.image)s
+# install:            %(args.install)s
+# rpmbuild root:      %(args.rpmbuild)s
+# architecture:       %s(ARCH)""" % locals())
```
reidpr, line 44: Should this still be commented out?

packaging/fedora/build
```diff
@@ -19,25 +19,29 @@ import time
...
+            "charliecloud-test"]
 ARCH = platform.machine()
 
 def main():
 
    # Parse arguments.
    ap = argparse.ArgumentParser()
-   ap.add_argument("version")
-   ap.add_argument("--image", metavar="DIR")
+   ap.add_argument("image", metavar="IMAGE")
```
reidpr, line 32: What is `IMAGE`? If still a directory, we should be clear about that.

test/run/build-rpms.bats
j-ogas: Is that not covered by the glob here?

packaging/fedora/build
j-ogas: It doesn't work with `locals()`. I can revert it back to two less tidy print statements.

reidpr: Yes, better to have uglier print calls than no status output IMO.

test/run/build-rpms.bats
reidpr: Sorry, I was unclear. I was thinking of below, which checks the output of this `rpm -qa`. The doc package is missing from those checks.

Issue #664, opened 2020-02-24
tests: no tests for ch-tar2squash


reidpr:
Somehow, we left `ch-tar2squash` out of the tarball and testing did not notice (#662). We should add some basic tests for the script, or get rid of it.

PR #671, opened 2020-02-26
ch-tar2squash: remove

Brofessional:
`ch-tar2squash` doesn't have much individual value. Remove it, so we don't spend time including it in our test suite. Addresses #664.


reidpr:
Note, the original comment referred to #671, i.e. itself. I edited that out.

Issue #649, opened 2020-02-18
getpwuid fails with no clear error message


reidpr:
Bug report in comment by @trophime in #137:

> @reidpr I've just tried with 0.13 release. I still have issue with the tutorial:
>
> ```
> hello$ uname -a
> Linux calcul19 5.4.0-3-amd64 #1 SMP Debian 5.4.13-1 (2020-01-19) x86_64 GNU/Linux
> hello$ lsb_release -cs
> bullseye
> hello$ sudo sysctl kernel.unprivileged_userns_clone
> kernel.unprivileged_userns_clone = 1
> images/hello$ ch-run /var/tmp/hello -- echo "I'm in a container"
> ch-run[12095]: error (ch_core.c:406)
> ```

reidpr:
This error points to line 406 in `ch_core.c`, which is:

```
   T_ (p = getpwuid(c->container_uid));
```

The error message says that (1) we haven't written any specific error message, (2) `getpwuid(3)` returned NULL, and (3) `errno` was zero.

(Aside, [tips on interpreting error messages](https://hpc.github.io/charliecloud/faq.html#how-do-i-read-the-ch-run-error-messages) from our FAQ.)

Looking at the [man page](http://man7.org/linux/man-pages/man3/getpwnam.3.html), this behavior happens when the UID (here `c->container_uid`) is not found in `/etc/passwd`. @trophime, is that true on your system?

Trophime:
I tried CharlieCloud on a linux box with authentification against ActiveDirectory...
My user is declared in the AD but not in the standard /etc/passwd
I used this tutorial to make the authentification works: https://help.ubuntu.com/community/ActiveDirectoryHowto

reidpr:
Thanks for the follow-up. Looking more closely at the man page, it sounds like it *should* get the information from Active Directory too: "The getpwnam() function returns a pointer to a structure containing the broken-out fields of the record in the password database (e.g., the local password file /etc/passwd, NIS, and LDAP)".

So, I wonder if there is a misconfiguration somehow in your Active Directory setup?

That said, we should probably deal with this case (e.g., by inserting a dummy `/etc/passwd` line), or at least give a coherent error message.

reidpr:
@Trophime, do you have a workaround?

Trophime:
Googling a little a found this [discussion](https://stackoverflow.com/questions/16666587/checking-username-getpwnam-getpwnam-r-no-such-file-or-directory)

[...]

PR #673, opened 2020-02-27
ch-run: don't fail if UID or GID account info unavailable

reidpr:
`ch-run` builds `/etc/passwd` and `/etc/group` files and bind-mounts them into the container, using information in the account database. If no such information is found, it exits with an unhelpful error message (#649); i.e., it doesn't distinguish between `getpwuid(3)` and `getpwgid(3)` failing and returning no results.

This PR adds code to instead use dummy account information if there are no results, and proceed with the user command. With `--verbose`, it prints a log message to say so.

For example:

```
$ ch-run -v -u 8675309 -g 8675310 /var/tmp/images/chtest -- /bin/true
ch-run[7985]: verbosity: 2 (ch-run.c:155)
ch-run[7985]: newroot: /var/tmp/images/chtest (ch-run.c:156)
ch-run[7985]: container uid: 8675309 (ch-run.c:157)
ch-run[7985]: container gid: 8675310 (ch-run.c:158)
ch-run[7985]: join: 0 0 (null) 0 (ch-run.c:159)
ch-run[7985]: private /tmp: 0 (ch-run.c:160)
ch-run[7985]: UID 8675309 not found; using dummy info (ch_core.c:416)
ch-run[7985]: GID 8675310 not found; using dummy info (ch_core.c:441)
$ echo $?
0
```

@trophime, does this help?

### DIFF
--- a/bin/ch_core.c
+++ b/bin/ch_core.c
@@ -403,10 +403,22 @@ void setup_passwd(struct container *c)
       T_ (1 <= dprintf(fd, "root:x:0:0:root:/root:/bin/sh\n"));
    if (c->container_uid != 65534)
       T_ (1 <= dprintf(fd, "nobody:x:65534:65534:nobody:/:/bin/false\n"));
-   T_ (p = getpwuid(c->container_uid));
-   T_ (1 <= dprintf(fd, "%s:x:%u:%u:%s:/home/%s:/bin/sh\n",
-                    p->pw_name, c->container_uid, c->container_gid,
-                    p->pw_gecos, getenv("USER")));
+   errno = 0;
+   p = getpwuid(c->container_uid);
+   if (p) {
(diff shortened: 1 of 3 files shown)


bin/ch_core.c
```diff
@@ -382,13 +382,20 @@ void setup_namespaces(struct container *c)
...
+/* Build /etc/passwd and /etc/group files and bind-mount them into newroot.
+
+   /etc/passwd contains root, nobody, and an entry for the container UID,
+   i.e., three entries, or two if the container UID is 0 or 65534. We copy the
+   host's user data for the container UID, if that exists, and use dummy data
+   otherwise (see issue #649). /etc/group works similarly: root, nogroup, and
+   an entry for the container GID.
+
+   We build new files to capture the relevant host username and group name
+   mappings regardless of where they come from. We used to simply bind-mount
```
j-ogas: Nice, thanks!

Issue #638, opened 2020-02-09
user+mount namespaces check always fails in chroots


wiene:
The user+mount namespaces check as implemented in the [`configure.ac` file](https://github.com/hpc/charliecloud/blob/master/configure.ac#L279-L288) always fails if run in a chroot environment. Such setups are common to ensure a clean build environment. According to the [`unshare` man page](http://man7.org/linux/man-pages/man2/unshare.2.html), `unshare` always throws an EPERM error in this case:

>        EPERM (since Linux 3.9)
>              CLONE_NEWUSER was specified in flags and the caller is in a
>              chroot environment (i.e., the caller's root directory does not
>              match the root directory of the mount namespace in which it
>              resides).

reidpr:
Thanks Peter, good catch.

Off the top of my head, it's not clear what to do about it. The man page quotation above seems to imply that user namespaces simply cannot be created within a chroot, i.e., it would be impossible to fix the check.

That said, the check is informational only; `configure` doesn't make any decisions based on it. So we could simply document the flaw to clarify this issue.

What is your opinion?

reidpr:
Check that: it looks like we could [test if we're in a chroot](https://unix.stackexchange.com/questions/14345) and have the check warn about false negatives or something like that.

wiene:
I like your proposals:
1. Mention this issue in the documentation.
2. Add an extra check which prints a warning if the build runs in a chroot.

reidpr:
Great! We'll do that.

PR #674, opened 2020-02-27
add warning about no user namespaces in a chroot

reidpr:
This hopefully addresses #638, but I don't have good enough chroot-fu to test it properly. @wiene, does this address the problem?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -288,6 +288,12 @@ AC_CHECK_PROG([UMOCI], [umoci], [umoci])
 CH_CHECK_VERSION([UMOCI], [$vmin_umoci], [--version | cut -d' ' -f3])
 
 # User namespaces
+AC_MSG_CHECKING([if in chroot])  # https://unix.stackexchange.com/a/14346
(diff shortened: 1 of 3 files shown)


wiene:
Using this patch, I get (running in a chroot):
```
[...]
checking if in chroot... yes
checking if user+mount namespaces work... no
[...]
  ch-run(1): no
    user+mount namespaces ... no    Warning: configure is running in a chroot, but user namespaces cannot be
    created in a chroot; see the man page unshare(2). Therefore, the above may
    be a false negative.
[...]
```
Running outside of a chroot environment, I see:
```
[...]
checking if in chroot... no
checking if user+mount namespaces work... yes
[...]
  ch-run(1): yes
    user+mount namespaces ... yes
[...]
```
To me this looks fine. Maybe one can clarify the warning a bit more by mentioning that this check is informational only and does not affect the build.

reidpr:
Great, thank you. I will merge once we get 0.14 out the door.

PR #676, opened 2020-02-27
examples/lustre: add missing variable, fix example grouping

j-ogas:
Adds the example grouping header. This should probably be documented in a FAQ or the developers guide.

It appears that `ch_lustre` will also need to be exported, the pendantic_fail trips otherwise (`ch_lustre` is empty at the test.bat file level).

### DIFF
--- a/bin/ch-test
+++ b/bin/ch-test
@@ -692,6 +692,9 @@ fi
 cd "$CHTEST_DIR"
 
 export PATH=$ch_bin:$PATH
+if [[ -n $ch_lustre ]]; then
+    export CH_LUSTRE=$ch_lustre
+fi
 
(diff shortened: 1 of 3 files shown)


reidpr:
OK, this is my fault. I missed a design error in review.

My commit adds an environment variable `CH_TEST_LUSTREDIR` to complement `ch-test --lustre`, much like `CH_TEST_PERMDIRS` and the others. This will make testing by the team easier b/c they don't have to give `--lustre` or be pedantic-failed.

I didn't actually test it on the clusters, so there are likely to be errors that remain.

Brofessional:
Everything appears to be in order

j-ogas:
CTS1 testing passed; looks good.

Issue #637, opened 2020-02-05
change spark dir from home to BATS_TMP?


j-ogas:
Can we move the spark dir away from home?

```
spark_dir=~/ch-spark-test.tmp  # runs before each test, so no mktemp
```

reidpr:
I think maybe it needs to be at a path shared between all the workers? But, there's no fundamental reason it needs to be in the home directory. I also don't recall if it needs shared writing, or if copying it to all the nodes would be OK.

reidpr:
Point being, proceed with putting in a better place if you think that's best.

PR #679, opened 2020-02-28
move spark test artifacts to TMP_

j-ogas:
Addresses #637.


test/common.bash
```diff
@@ -169,7 +169,8 @@ fi
 # User-private temporary directory in case multiple users are running the
 # tests simultaneously.
 # shellcheck disable=SC2154
-btnew=$TMP_/bats.tmp
+ch_tmp=$TMP_
+btnew=$ch_tmp/bats.tmp
```
reidpr: Why a new second variable instead of using `$TMP_` directly or replacing it?

reidpr:
Does this work on the clusters? Do the distributed workers need access to the configuration directory too?

reidpr:
Per offline discussion, the distributed workers do need access to the configuration directory. It needs to be at the same path on each node but doesn't need to be the same actual file (a copy is OK).

Issue #484, opened 2019-07-11
more friendly image management (“Goff Plan”)


shanegoff:
With the new builders and squashfs support we've created a number of new scripts and helpers for managing all this stuff. It's awesome to support so many things but it might be easier for users to have the utility crammed into one place.

We have files to create different output files between tar and squash. Adding the new builders, the new support was added into a single file instead of multiple files. We should follow that idea farther I think.

Example:

`ch-export -b [BUILDER] -t [OUT_TYPE] IMAGE OUTDIR`

This would make it so flattening an image from a builder to a file would be a one stop shop. 

`ch-export -b ch-grow -t dir alpine /tmp`
`ch-export -b docker -t squash lammps /dev/shm`

This would be easier for people when managing images. They wouldn't have to call separate files for different output types.

It would also clear up some confusion regarding that the builders don't actually build something when you call `ch-builder2tar` so it would be more clear that you are creating a flattened image that was already created by the builder, not building it.

reidpr:
In our discussion today, it was pointed out that we now have 17 executables, which is a lot. We decided that the next step is to articulate the workflow better so we can understand how best to reorganized.

j-ogas:
Recommend using something like`-o, --out-type` instead of `-t, --out-type`. `-t` is closely related to `TAG`.

reidpr:
@Brofessional is suggesting that `ch-run` should know how to run tarballs and squashes in addition to directories.

Brofessional:
I also suggest using a single command with sub commands for the future

`ch-run $IMAGE -- $COMMAND` ---> `ch run $IMAGE $COMMAND`
`ch-builder2squash $IMAGE $OUTDIR` ----> `ch export $IMAGE $OUTDiR`

etc
I'll look into creating bash completion scripts for the use case.

reidpr:
This has been done, with two exceptions: (1) `ch-convert` doesn't speak Podman (issue #1360), and (2) `ch-run` doesn't know how to call `ch-fromhost` automatically (issue #1361). Closing.

PR #684, opened 2020-03-05
Make image packing friendly

shanegoff:
With some changes made in #634 it is simple (I think) to make packing an image more friendly for users. First installment to the Goff plan #484


Brofessional:
I am open to names other than `ch-pack-image` on the condition the name change proposal includes you going and renaming it all the places it has to change =]

bin/Makefile.am
```diff
@@ -19,11 +19,11 @@ ch_ssh_SOURCES = ch-ssh.c ch_misc.h ch_misc.c
 
 dist_bin_SCRIPTS = ch-build \
                    ch-build2dir \
-                   ch-builder2squash \
                    ch-builder2tar \
```
j-ogas: Should ch-build2tar be removed here as well?

bin/Makefile.am
```diff
@@ -19,11 +19,11 @@ ch_ssh_SOURCES = ch-ssh.c ch_misc.h ch_misc.c
...
                    ch-dir2squash \
                    ch-fromhost \
                    ch-mount \
+                   ch-pack-image \
```
j-ogas: FWIIW I prefer the simpler `ch-pack`.

bin/ch-pack-image
```diff
@@ -0,0 +1,79 @@
+#!/bin/sh
+
+libexec=$(cd "$(dirname "$0")" && pwd)/../libexec/charliecloud
+. "${libexec}/base.sh"
```
j-ogas: I believe the `$()` format is a bashism. This will need to be replaced with sh syntax if we plan to keep this script POSIX.

```

File: bin/ch-pack-image [hunk 4]
```diff
@@ -0,0 +1,79 @@
...
+
+# Kludgey argument parsing.
+while true; do
+    case $1 in
+        -b|--builder)
+            shift
+            CH_BUILDER=$1
+            shift
+            ;;
+        -b=*|--builder=*)
```
j-ogas: should this be `-b|--builder=*`?

bin/ch-pack-image
```diff
@@ -0,0 +1,79 @@
...
+  $ $(basename "$0") [-b BUILDER] [--fmt CH_PACK_FMT] IMAGE OUTDIR [ARGS ...]
+
+You must have sufficient privilege (via sudo) to run the Docker commands.
+EOF
+)
+
+parse_basic_args "$@"
+
+# Kludgey argument parsing.
+while true; do
```
j-ogas: Can we not use the number of arguments, e.g., `$#`, in POSIX shell?

bin/ch-pack-image
```diff
@@ -0,0 +1,79 @@
...
+        -b|--builder)
+            shift
+            CH_BUILDER=$1
+            shift
+            ;;
+        -b=*|--builder=*)
+            CH_BUILDER=${1#*=}
+            shift
+            ;;
+        --pack-fmt)
```
j-ogas: This argument, `--pack-fmt`, is inconsistent with what usage specifies (`--fmt`).

FWIIW I think `--format` is better.

bin/ch-pack-image
```diff
@@ -0,0 +1,79 @@
...
+            shift
+            CH_PACK_FMT=$1
```
j-ogas: I think explicit variables are more clear, e.g. `CH_PACK_FORMAT`.

bin/ch-pack-image
Brofessional: `--pack-fmt` makes it consistent with `ch-test`

bin/ch-pack-image
Brofessional: I'm not opposed to this, but I also don't like how much work I would have to do to go change it everywhere lol

bin/ch-pack-image
Brofessional: No, this should allow users to `-b=$BUILDER` or `--builder=$BUILDER`

bin/ch-pack-image
reidpr: ShellCheck is really good about flagging such things. If you're unsure, use the better syntax, and let ShellCheck gripe if you're wrong. Often that's easier than trying to look it up.

bin/ch-pack-image
j-ogas: > `--pack-fmt` makes it consistent with `ch-test`

I am pointing out a discrepancy in the usage output and your argument handling. Your usage message tells users to pass `--fmt` but your argument handling is looking for `--pack-fmt`.

bin/ch-pack-image
j-ogas: > No, this should allow users to `-b=$BUILDER` or `--builder=$BUILDER`

Do we do this anywhere else? If not, why start here?

Issue #648, opened 2020-02-16
several files in libexec are erroneously executable


wiene:
Charliecloud installs various files marked as executables (e. g. `base.sh` or `version.*`) although those files do not contain shebang lines (and are not binaries either). This is an unusual combination which triggers warnings when using lint tools. I would suggest to remove the executable bit for those files.

reidpr:
Yes, that's definitely a mistake. The reason is that was how we were able to persuade Autotools to install them. If you or anyone has better ideas, I'm all ears. :) Related issue is #596.

j-ogas:
This will important for the updated Fedora package. However, I can always change their permissions in the spec file.

Issue #683, opened 2020-03-05
don't hard-code path to Python and sh libraries


ana:
Hi,

After the build system rewrite, several ch-* scripts and base.sh have harcoded the path `../libexec/charliecloud`.  This wants to create `/usr/libexec/charliecloud` but `/usr/libexec`  is not a standard directory in the Filesystem Hierarchy Standard.
I have patched this directly for openSUSE, please see: https://build.opensuse.org/package/view_file/network:cluster/charliecloud/0001-Replace-hardcoded-libexec.patch?expand=1
This will be also problematic for Debian.

However, I think a variable in the buildsystem so the path can be chosen would be a better solution for this. There was a previous discussion in this issue https://github.com/hpc/charliecloud/issues/93

reidpr:
Hello Ana, thanks for the report. We'll look into it.

We do know that a lot, but not all, of the stuff currently in `libexec` needs to be moved to `lib`; see issue #596. The current state also causes problems with Autocools; see issue #648. Looking at your patch, this move alone may address your immediate concern.

A couple of constraints that are relevant:

1. We want to keep Charliecloud working when built in the source directory, without install. I'm not sure how to do a build system variable and keep this; have I missed something?
2. Fedora/CentOS/Red Hat *insists* on certain things going in `libexec`; e.g., we got a lot of blowback trying to put the test suite in `doc`.

reidpr:
Looks like `/usr/libexec` is in [FHS 3.0](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#usrlibexec) but not 2.3. Debian policy manual says the project [follows FHS 3.0](https://www.debian.org/doc/debian-policy/ch-opersys.html#file-system-hierarchy), but you know better — what am I missing?

ana:
Hi @reidpr 

Indeed, after checking the Debian policy and also some Debian tools (lintian, debhelper, etc), you can use `/usr/libexec` without problems from a few months before the release of buster.

I'm afraid this is still not the case in openSUSE/SUSE, where it's still FHS 2.3: https://build.opensuse.org/package/show/Base:System/fhs

In any case, paths shouldn't be hard coded and given you're using autotools I've have expected I could have used the parameter `--libexecdir=$PREFIX/lib` to use `usr/lib` instead of `usr/libexec` without more changes. But this didn't work because the ch-* script have hard coded 
 `../libexec/charliecloud`

[...]

Issue #596, opened 2019-12-20
move files from libexec to lib?


reidpr:
I believe everything we install in `libexec` (`version.*` and `base.sh`) might properly belong in `lib`. If this is true, perhaps also move the test suite from `libexec/test` to `libexec`.

Follow-up to PR #589.

PR #685, opened 2020-03-05
move libraries from libexec to lib

reidpr:
This PR addresses #596. That issue also proposes moving the test suite to the root of `libexec`. Even though `test` is the only entry remaining in `libexec`, I did not move the test suite up, because I anticipate growing some new executables in `libexec`.

I believe it also addresses #648. @wiene, can you comment? Should it also close that bug?

It may also address #683. @ana, what is your opinion on that? Would you like that issue to remain open due to the request for a build system variable, or be closed because it eliminates the need for your patch, or something else?

Thanks to @ana for providing a patch that forms part of this PR.


ana:
Thanks @reidpr, this solves the need for a patch in openSUSE. I still think the hard coded path is a bad idea as mentioned in https://github.com/hpc/charliecloud/issues/683#issuecomment-595780211 but you can close the issue if you wish.

.gitignore
```diff
@@ -11,11 +11,11 @@ a.out
...
-/libexec/charliecloud
-/libexec/contributors.bash
-/libexec/version.py
-/libexec/version.sh
-/libexec/version.txt
+/lib/charliecloud
+/lib/contributors.bash
+/lib/version.py
+/lib/version.sh
+/lib/version.txt
```
j-ogas: I'm not sure I like putting these generically named files directly into`/lib`. What are your thoughts about putting these files in a charliecloud sub directory, for example`/lib/charliecloud`, as the default behavior?

test/travis.yml
```diff
@@ -28,10 +28,10 @@ compiler: gcc
 # weird errors, e.g.:
 #
 #   make[2]: Entering directory [...]
-#    /bin/mkdir -p '/var/tmp/libexec/charliecloud'
-#    ../yes base.sh '/var/tmp/libexec/charliecloud'
+#    /bin/mkdir -p '/var/tmp/lib/charliecloud'
+#    ../yes base.sh '/var/tmp/lib/charliecloud'
```
j-ogas: It looks like this path is expecting a charliecloud parent directory, e.g., `/lib/charliecloud`.

wiene:
> I believe it also addresses #648. @wiene, can you comment?

Why do you think it does?

reidpr:
@wiene, the files you flagged as being executable but without a shebang should now be not executable (and still without a shebang).

wiene:
I've just tried out this PR and the mentioned files are actually not executable anymore. Thus I think that you are right that this patch fixes #648.

PR #697, opened 2020-03-11
distribute Autoconf Archive macros we care about with Charliecloud

reidpr:
This PR adds the Autoconf Archive macros we care about to the Charliecloud source code, eliminating the external dependency on Autoconf Archive to run `autogen.sh` (e.g., to build from a Git checkout).

Installing Autoconf Archive is not hard if you control your build box, because you can just install your distro's package. However, we keep running into boxen where we want to build from Git, but Autoconf Archive is not installed and it's hard to install promptly.

### DIFF
--- a/misc/m4/README
+++ b/misc/m4/README
@@ -1,0 +1,38 @@
+This directory contains additional M4 macros for the build system.
+
+Currently, these are all from Autoconf Archive. While many distributions have
+an Autoconf Archive package, which autogen.sh can use if present, it’s a
+little uncommon to have installed, and we keep running into boxen where we
+want to run autogen.sh, but Autoconf Archive is not installed and we can’t
+install it promptly.
+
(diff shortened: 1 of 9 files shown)


misc/m4/README
```diff
@@ -0,0 +1,38 @@
...
+an Autoconf Archive package, which autogen.sh can use if present, it’s a
+little uncommon to have installed, and we keep running into boxen where we
+want to run autogen.sh, but Autoconf Archive is not installed and we can’t
+install it promptly.
+
+There is a licensing exception for these macros that lets us redistribute
+them: “Every single one of those macros can be re-used without imposing any
+restrictions whatsoever on the licensing of the generated configure script. In
+particular, it is possible to use all those macros in configure scripts that
+are meant for non-free software.” [1]
```
j-ogas: Nice.

PR #698, opened 2020-03-17
new script ch-convert

reidpr:
This PR proposes a new script `ch-convert` that would replace `ch-tar2dir` and several others. Relevant to #484.

Currently it's only a man page.

### DIFF
--- a/doc/ch-convert_desc.rst
+++ b/doc/ch-convert_desc.rst
@@ -12,7 +12,9 @@ Copy image :code:`IN` to :code:`OUT` and convert its format along the way if
 needed. Replace :code:`OUT` if it already exists, unless :code:`--no-clobber`
 is specified.
 
-The Charliecloud workflow involves four basic steps:
+:code:`ch-run(1)` requires an image in directory form. Therefore, the goal of
+the Charliecloud workflow is to end up with a directory, but there is more
+than one way to get there. One possible workflow is:
 
   1. Build an image using a Dockerfile or pull it from a repository, placing
      the resulting image in *builder storage*.
@@ -51,11 +53,13 @@ Thus, Charliecloud deals with images stored in four different formats:
 by mounting a SquashFS archive using :code:`ch-mount(1)` — but that is not an
 image *storage* format.)
 
-The purpose of :code:`ch-convert` is to convert images among these four
-formats. However, it can also be used to copy images within the same format.
+The purpose of :code:`ch-convert` is to copy images, either within the same
+format or between different formats.
 
-Some copies require temporary storage equal to the uncompressed size of the
-image.
+Many of the copy algorithms have shortcuts available, which :code:`ch-convert`
+knows about, so you should typically invoke it only once per workflow, asking
+for the final format you actually need. Some copies require temporary storage
+equal to the uncompressed size of the image.
 
 
 Arguments
@@ -100,8 +104,9 @@ Arguments
     :code:`ch-grow(1)`. (Currently only implemented for :code:`ch-grow(1)`.)
 
   :code:`--tmp DIR`
-    Path to temporary directory. Default: :code:`$TMP` if specified; otherwise
-    :code:`/tmp`.
+    Path to temporary directory. Storage needed varies between zero and the
+    uncompressed size of the image, depending on the input and output formats.
+    Default: :code:`$TMP` if specified; otherwise :code:`/tmp`.
 
   :code:`-v`, :code:`--verbose`
     Print extra chatter.
@@ -113,31 +118,47 @@ Format and filename inference
 :code:`ch-convert` tries to save typing by guessing formats and filenames when
 they are reasonably clear.
 
-Format inference is done according for both :code:`IN` and :code:`OUT`. The
-algorithm tries to match the value agains the following globs in this order.
-Paths need not exist in the filesystem.
+Format inference is done for both :code:`IN` and :code:`OUT`. The algorithm
+tries to match the value against the following globs in this order. Paths need
+not exist in the filesystem.
 
   1. :code:`*.sqfs`, :code:`*.squash`, :code:`*.squashfs`: SquashFS.
-  2. :code:`*.tar`, :code:`*.t?z`, :code:`*.tar.??`, :code:`*.tar.??`: tarball.
-  3. :code:`*:*` (i.e., containing a colon): builder storage.
-  4. Otherwise: directory.
+
+  2. :code:`*.tar`, :code:`*.t?z`, :code:`*.tar.??`, :code:`*.tar.??`: Tarball.
+
+  3. :code:`/*`, :code:`./*` (i.e., absolute path or relative path with
+     explicit dot): Directory.
+
+  4. :code:`*:*` (i.e., containing a colon): Builder storage.
+
+  5. Otherwise: No format inference.
 
 A notable consequence of these rules is than a simple image reference such as
-:code:`debian` is inferred to be a directory. The workaround is to add the
-default tag, e.g. :code:`debian:latest`.
+:code:`debian` cannot be inferred. The workaround is to add the default tag,
+e.g. :code:`debian:latest`.
 
-:code:`ch-convert` also infers filenames for :code:`OUT` (but not :code:`IN`)
-to save typing. If :code:`OUT` is path to a directory that exists, and the
-output format is:
+:code:`ch-convert` also infers filenames for :code:`OUT` (but not :code:`IN`).
+If :code:`OUT` is path to a directory that exists, and the output format is:
 
   * directory,
+
     * and :code:`OUT` ends in slash: append inferred filename.
     * otherwise: use :code:`OUT` without modification.
+
   * tarball or squash: append inferred filename.
   * builder storage: use :code:`OUT` without modification.
 
-Inferred filename rules: FIXME
+Filenames are inferred by deriving a base from :code:`IN` and then adding an
+extension appropriate for :code:`--out-fmt`. Base if :code:`IN` is:
 
+  * builder storage: image reference (i.e., :code:`IN`) with slashes changed
+    to percents.
+
+  * tarball: filename with tarball extension (e.g., :code:`.tar.gz`) removed.
+
+  * SquashFS: filename with SquashFS extension (e.g., :code:`.sqfs`) removed.
+
+  * directory: final component of the path.
 
 Supported conversions
 =====================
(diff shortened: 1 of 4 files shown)


doc/ch-convert_desc.rst
```diff
@@ -0,0 +1,259 @@
...
+  2. :code:`*.tar`, :code:`*.t?z`, :code:`*.tar.??`, :code:`*.tar.??`: Tarball.
+
+  3. :code:`/*`, :code:`./*` (i.e., absolute path or relative path with
+     explicit dot): Directory.
+
+  4. :code:`*:*` (i.e., containing a colon): Builder storage.
+
+  5. Otherwise: No format inference.
+
+A notable consequence of these rules is than a simple image reference such as
```
Brofessional: I think "than" was a typo? Did you mean "that"?

doc/ch-convert_desc.rst
```diff
@@ -0,0 +1,259 @@
...
+:code:`ch-convert` tries to save typing by guessing formats and filenames when
+they are reasonably clear.
+
+Format inference is done for both :code:`IN` and :code:`OUT`. The algorithm
+tries to match the value against the following globs in this order. Paths need
+not exist in the filesystem.
+
+  1. :code:`*.sqfs`, :code:`*.squash`, :code:`*.squashfs`: SquashFS.
+
+  2. :code:`*.tar`, :code:`*.t?z`, :code:`*.tar.??`, :code:`*.tar.??`: Tarball.
```
Brofessional: It may be useful to try using the `file` command to infer file types. 

```
(base) $ ls
0a_directory  centos.sqfs  centos.tar.gz  centos7xz.tar.xz
(base) $ file centos.tar.gz
centos.tar.gz: gzip compressed data, last modified: Tue Mar  3 15:12:31 2020, from Unix, original size modulo 2^32 244943360 gzip compressed data, reserved method, has CRC, has comment, from FAT filesystem (MS-DOS, OS/2, NT), original size modulo 2^32 244943360
(base) $ file centos7xz.tar.xz
centos7xz.tar.xz: XZ compressed data
(base) $ file centos.sqfs
centos.sqfs: Squashfs filesystem, little endian, version 4.0, 69720732 bytes, 11087 inodes, blocksize: 131072 bytes, created: Tue Mar  3 15:12:45 2020
(base) $ file a_directory/
a_directory/: directory
```

I am not certain it works for every one of those tar formats but I would expect so. This is mostly helpful with squashfs I think. Other certain container tools using squashfs have the freedom to name it anything. Charliecloud could still work with these images if we used the text returned from `file` instead of the extension.

reidpr: One issue I see above is, how do you tell a tar-gzip from plain gzip?

For SquashFS, we could check the magic number ourselves with `dd`. I don't believe `file(1)` is POSIX, but that said I have a hard time imagining a system without it, but that said is the output format machine-parseable?

doc/ch-convert_desc.rst
reidpr: Yes; thank you.

doc/ch-convert_dest.rst
```diff
@@ -0,0 +1,179 @@
...
+  * tarball or squash: append inferred filename.
+  * builder storage: use :code:`OUT` without modification.
+
+Inferred filename rules: FIXME
+
+
+Supported conversions
+=====================
+
+Not all input/output format pairs are supported.
```
rstyd: This statement that not all input/output format pairs are supported, but the diagram states that every input/output format is either supported or will eventually be supported. Am I misinterpreting this?

doc/ch-convert_desc.rst
Brofessional: You can ask it for all valid extensions for the file type. The bad news is that only works for gzip compressed. It might also be difficult to parse the output of file. I thought there was a flag to only print the file type, but I can't seem to make it work.

doc/ch-convert_desc.rst
```diff
@@ -0,0 +1,259 @@
...
+error because :code:`/var/tmp` already exists but doesn't look like an image.
+
+::
+
+  $ ch-convert foo/bar:latest /var/tmp
+  input:   builder storage  foo/bar:latest
+  output:  directory        /var/tmp
+  ch-convert[1234]: directory exists but doesn't appear to be an image: /var/tmp
+
+Copy an image to a SquashFS file, and provide options :code:`-b 1048576` to
```
rstyd: What does -b 1048576 mean here?

reidpr:
Need to take care with `unsquashfs(1)` options to avoid:

```
Parallel unsquashfs: Using 64 processors
20198 inodes (23645 blocks) to write
 
 
write_xattr: could not write xattr security.selinux for file /var/tmp/ior-openmpi-at/bin because you're not superuser!
 
write_xattr: to avoid this error message, either specify -user-xattrs, -no-xattrs, or run as superuser!
 
Further error messages of this type are suppressed!
[=============================================================================================- ] 23644/23645  99%
 
created 17960 files
created 2059 directories
created 2020 symlinks
created 0 devices
created 0 fifos
```

Thanks for catching this @atorrez and @Brofessional!

shanegoff (commented):
I like the new approach. I'm sure I'll mess up the formats every time trying to do implicit so I'm happy about the explicit options, it looks a little more the way I'd be used to exporting images.

I am totally unreasonably opposed to adding other output extensions for squashfs. No strong reason, it just makes output extensions more complicated.

down with `.squash` and `.squashfs` long live `.sqfs`

doc/command-usage.rst
```diff
@@ -26,6 +26,13 @@ Build a Charliecloud image from Dockerfile and unpack it into a directory.
...
 
+ch-builder2squash
++++++++++++++++++
+
+Flatten a builder image into a Charliecloud SquashFS file.
+
+.. include:: ./ch-builder2squash_desc.rst
+
 ch-builder2tar
 ++++++++++++++
```
j-ogas, line 37: Shouldn't this be shoveled now?

j-ogas (changes requested):
Looks good. Are we still planning to add the conveniece subcommand `ch-image convert`? If so we should also add the documentation draft for that here as well.

doc/command-usage.rst
reidpr: Lots of the existing scripts will get the shovel. I can add deprecation notices to them.

reidpr:
> Are we still planning to add the conveniece subcommand `ch-image convert`?

Yes, though I think it would be three subcommands: `dir`, `tar`, `squash`. I was thinking that would be a separate PR though. What's your argument it should be included here?

j-ogas:
> Yes, though I think it would be three subcommands: dir, tar, squash. I was thinking that would be a separate PR though. What's your argument it should be included here?

Ah, three explicit subcommands sounds like a better approach. In that case, yes, a separate PR seems appropriate.

reidpr:
This now has the key features working is ready for a first review. Miscellaneous things that I still want to fix here:

* [ ] `ch-image` does not add `:latest` by default, i.e. if you say `ch-image build -t foo` it will create an image called `foo` with no tag, rather than `foo:latest` like Docker. This makes format inference less useful because `ch-image` images are not guaranteed to have a colon. We could have builder be fallback format, rather than inference error as currently.
* [ ] Update tests throughout to use `ch-convert` rather than the old scripts.
* [ ] Add `--no-clobber` tests.
* [ ] Conversion tests should be pedantic mode. This is a little tricky because Docker needs `sudo` on Travis while we delete it for the `ch-image` tests.
* [ ] Print `done` or similar at end, along with info about output product? Gathering size of `dir`, `ch-image`, or `docker` format images requires a directory walk.
* Document:
  * [ ] noisyness of conversions to SquashFS
  * [ ] temporary directory is left on failure

Additional imperfections that I want to put into future issues:

* [ ] Several of the conversions do not have a progress meter; I tried to mark all of these in comments. E.g. `ch-image` → `dir` doesn't have one, because `cp` doesn't and there's no simple way to add one; we could use `rsync` (less portable) or `tar`/`pv`/untar (adds extra steps).
* [ ] Several of the conversions have a needless compress/uncompress cycle.
* [ ] Command line argument parsing is rather inflexible; e.g. `ch-convert -v -o foo bar baz` works but neither `ch-convert -vo foo bar baz` (no combined options) nor `ch-convert -v bar -o foo baz` (no mixing options and arguments) work. This is due to the manual parsing. GNU `getopt` would work but is not POSIX.
* [ ] Remove deprecated scripts in next release, including remove FAQ about "no command specified", since I figured out how to get rid of that.
* [ ] Color output might be nice, though see issue #1093.

rstyd:
Could we integrate a `--format` or `--output` parameter into `ch-image` that leverages this so that the image is immediately converted into the desired format after it's built?

shanegoff (approved):
I think the discussions we had around implicit filenames for directory output is really the only big thing on my mind. The code looks good and I think this will help simplify converting between formats.

I look forward to the final form and testing it out.

shanegoff:
> 
> 
> Could we integrate a `--format` or `--output` parameter into `ch-image` that leverages this so that the image is immediately converted into the desired format after it's built?

In the same spirit we could also have ch-convert take a dockerfile as an in format and choose an output format for the built image

doc/ch-convert_desc.rst
```diff
@@ -0,0 +1,167 @@
...
+
+      [1]: https://manpages.debian.org/unstable/golang-github-containers-image/containers-transports.5.en.html
+
+   2. There used to be an [OUT_ARG ...] that would be passed unchanged to the
+      archiver, i.e. tar(1) or mksquashfs(1). However it wasn't clear there
+      were real use cases, and this has lots of opportunities to mess things
+      up. Also, it's not clear when it will be called. For example, if you
+      convert a directory to a tarball, then passing e.g. -J to XZ-compress
+      will work fine, but when converting from Docker, we just compress the
+      tarball we got from Docker, so in that case -J wouldn't work.
```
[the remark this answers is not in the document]
reidpr: My `mksquashfs` doesn't have `-quiet`. Should `-quiet` be the default if available?

doc/ch-convert_desc.rst
```diff
@@ -0,0 +1,167 @@
...
+  input:   docker    foo/bar:latest
+  output:  squashfs  /var/tmp/foobar.sqfs
+  copying ...
+  done
+  $ ch-run /var/tmp/foobar.sqfs -- echo hello
+  hello
+
+Same conversion, but no format inference::
+
+  $ ch-convert -i ch-image -o squash foo/bar:latest /var/tmp/foobar.sqfs
```
shanegoff, line 160: If I tell `ch-convert` to explicitly give me a squash and give it a .tar.gz extension it is totally fine with that. Not sure if maybe we want to raise a warning of a format mismatch in explicit out situations.

shanegoff: Ubuntu 20.04 also has no qualms coloring my ls output making it look like it is really a tar file too.

image

shanegoff (changes requested):
Overall I think everything is working as intended. Only a few small things here and there that I thought were maybe confusing in the documentations.

doc/ch-convert_desc.rst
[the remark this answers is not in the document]
shanegoff: I believe the original grand design included `-quiet` as the default and it didn't work on any OS except maybe ubuntu 18 and 20. The version in centos7 was too old to support it. I think if checking for the flag is easy enough -quiet would help with the large amount of output `mksquashfs` creates.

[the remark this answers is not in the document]
shanegoff: I guess that never occurred to me in my first real PR to add support for squashfs, haha.

reidpr:
More to-do:

* [ ] re-check docs for deprecated scripts post merge of PR #1102

doc/ch-convert_desc.rst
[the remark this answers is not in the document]
reidpr: Issue #1210.

Issue #672, opened 2020-02-26
ch-test: don't delete image directories before knowing if we can proceed


reidpr:
@Brofessional reports offline:

> If you run `ch-test` with no arguments, it tries to run all the phases. Including a phase where it removes and creates your tar dir. This may be the desired outcome on your laptop. But on [our cluster] it `rm`d my full scope build full of images leaving me to [...] retransfer back to the cluster. So for cluster testing this has been a pretty big setback.

Example output:

```
$ ./bin/ch-test
ch-run:       /users/rgoff/charliecloud-0.14~pre+chtest2.81040d1/bin/ch-run
bats:         /users/rgoff/[...]/libexec/bats (0.4.0)
tests:        /users/rgoff/charliecloud-0.14~pre+chtest2.81040d1/test
installed:    no
git workdir:  no
contributor:  yes; rgoff@lanl.gov in README.rst

phase:                build run examples
scope:                standard (default)
builder:              docker (environment)
pedantic mode:        yes (default)
use generic sudo:     no (default)
unpacked images dir:  /users/rgoff/img (environment)
packed images dir:    [...]/rgoff/tarballs (environment)
fs permissions dirs:  skip (environment)

ch-test version: 0.14~pre+chtest2.81040d1
warning: inconsistent source version: 0.14~pre

checking namespaces ...
ok

removing [...]/rgoff/tarballs
creating [...]/rgoff/tarballs

removing /users/rgoff/img
creating /users/rgoff/img

checking builder ...
error: builder: docker: not installed
```

In this specific case, the `build` phase couldn't be done because there is not builder installed, but `ch-test` deleted `$CH_TEST_TARDIR` before figuring that out.

`ch-test` shouldn't do any deleting until it's fairly confident it can get through all the requested tests.

j-ogas:
During an offline meeting we discussed the following options:

1. Remove default `phase=all` behavior and instead require users to specify a phase;
2. sanity check, e.g., see if we can rebuild what we are deleting, or detect if we are on a cluster;
3. use `$TMP_` for creating image artifiacts and then move to `CH_TEST_[TAR|IMG]DIR`;
4. use one or more sub directories; and
5. do nothing.

I prefer option one (1). Users should specify either `build`, `run`, `examples`, or `all`; otherwise print `usage`. This is, IMO, the simplest to implement and most intuitive usage solution.

PR #702, opened 2020-03-20
ch-test: remove `phase=all` default

j-ogas:
Addresses: #672.


bin/ch-test
```diff
@@ -786,8 +786,7 @@ printf 'contributor:  %s\n\n' "$ch_contributor_note"
 
 printf "%-21s %s" 'phase:' "$phase"
 if [[ -z $phase ]]; then
-    phase=all
-    printf 'build run examples'
+    fatal 'phase: no phase specified'
```
reidpr: Should this still print `build run examples` if you give `all`?

doc/ch-test_desc.rst
```diff
@@ -37,6 +37,10 @@ filesystem permissions fixtures must be created manually, in order to
 accommodate configurations where sudo is not available via the same login path
 used for running tests.
 
+The packed and unpacked image directories specified for testing are volatile.
+The contents of these directories are removed before each respective build and
+run phases.
+
```
reidpr: Suggest minor rephrase: "The packed and unpacked image directories are deleted before the build and run phases, respectively.".

bin/ch-test
j-ogas: I think so. I feel `build run examples` is more explicit than `all`, however, I am wiling to change it if you prefer.

reidpr: I agree; I think `build run examples` is clearer. But in that case, looks like that code needs to be added.

j-ogas: We added `all` to the above case statement that assigns `phase=all`. Down further our primary switch has:

```
    all)
        phase=build
        dir_tar_mk
        builder_check
        test_make
        test_build

        phase=run
        dir_img_mk
        dir_tar_check
        test_make
        test_run

        phase=examples
        dir_tar_check
        test_make
        test_examples
        win
```

Am I missing something?

j-ogas: Looks like I need to update the docs for starters (oops).

reidpr:
One other detail from call: Need to update the test that checks whether executables give help on no arguments, which `ch-test` is currently excluded from.

bin/ch-test
reidpr: I think I made a mistake. The current behavior in this PR with phase `all` is that it rolls through `build`, `run`, `examples`, announcing each in turn? Right? If so, then my objection is erroneous because it *does* announce each of the phases, just not where I was looking.

doc/ch-test_desc.rst
```diff
@@ -47,8 +51,8 @@ tested on different systems by copying the necessary artifacts between them,
 e.g. by building images on one system and running them on another. The *scope*
 allows trading off thoroughness versus time.
 
-:code:`PHASE` must be one of the following; the default is to run
-:code:`build`, :code:`run`, and :code:`examples` in that order.
+:code:`PHASE` must be one of the following; :code:`build`, :code:`run`,
+:code:`examples`, or :code:`all`.
```
reidpr: `PHASE` can be anything in the bullet list; the clause after the semicolon can be removed.

doc/ch-test_desc.rst
```diff
@@ -62,6 +66,10 @@ allows trading off thoroughness versus time.
     Example applications. Requires an unpacked images directory produced by a
     successful :code:`run` phase.
 
+  :code:`all`
+    Executes phases :code:`build`, :code:`run`, and :code:`examples`, in that
```
reidpr: Change to imperative mood.

Issue #607, opened 2020-01-04
ch-test: allow execution of individual files


reidpr:
It would be nice if we could use `ch-test` to execute individual `.bats` files ~~or even individual tests~~. Currently the man page advertises this as a not-yet-implemented scope. Possible UI:

```
$ ch-test ./examples/hello/test.bats
```

~~or~~

```
$ ch-test './build.bats:lint shell scripts'
```

~~The latter form maybe should imply full scope, since if the user is going to the trouble of requesting a specific test, they probably want to run it even if the (perhaps unstated) scope argument says otherwise.~~

CalvinDSeamons:
I do like the idea of running `ch-test` and then point to a specific file (straight to the point). The other option I thought was to following the scope command. 
```console
$ ch-test -scope singleton ./example/hello/test/bats
```
This way I feel mitigates the confusion around a user selecting `--scope full` on top of trying to choose just a single test.

reidpr:
I think the UI would be simpler if specifying an individual test just implied `--scope full`.

I don't think specifying a `.bats` file should imply anything about scope, because files can be mixed. Though, we will want to do something intelligent if none of the tests in that file are in scope.

reidpr:
Scope change. This issue is for individual files; #703 is for individual tests.

PR #704, opened 2020-03-20
ch-test: add --file and --file-tag

reidpr:
Addresses #607, allowing execution of individual Bats files.

There is one thing that's a bit weird; in `test.bats` we now refer to `$ch_test_tag`, which is sometimes substituted with `sed` and sometimes a real shell variable. I didn't see a workaround for this other than always copying whatever file is specified, which seemed like it could be tricky since we don't know what kind of file it is (build, run, example).

I have also alphabetized the functions in `ch-test` and split `build.bats` and `build_post.bats` into multiple files.


bin/ch-test
```diff
@@ -560,28 +577,32 @@ test_make_template_print () {
...
-            ;;
-        esac
-        export CH_PACK_FMT=$pack_fmt
-        method='command line'
-    elif [[ -n $CH_PACK_FMT ]]; then
-        method='environment'
-    else
-        pack_fmt_choose
-        method='default'
+test_one_file () {
```
Brofessional: Could we force the scope to full in this function? It feels silly to have to specify full scope to test an example not in standard scope. Could we call `require_unset` on scope and just set it to full?

Brofessional: I guess you could easily just set your scope to full with an environment variable if you wanted to be lazy. It could get confusing for people running full scope tests when you didn't mean to.

reidpr:
> I thought about forcing scope to full, but that seems overkill here.

Sounds like we agree, but let me give a concrete example. The test `test/build/50_dockerfile.bats:ARG and ENV` is standard scope under `--builder=ch-grow` because it takes a few seconds, but full scope under the other builders because it takes over a minute. 🙄 So that would slow testing of the rest of that file a lot.

Other discussions I believe have converged on setting full scope if you give an individual test, once we have that implemented.

reidpr:
> This may be a second issue, or a followup, but I would like to be able to provide `ch-test` a phase and a tag and have ch-test run accordingly. For example, `ch-test all --file-tag lammps -s full`. I imagine this would populate `examples` and/or `images` variable to one artifact.

I agree, that sounds cool but a separate issue. I wonder if that's a good way to get individual tests as well, since that would need to roll through the filters to pull out the test we want.

doc/ch-test_desc.rst
```diff
@@ -100,6 +104,12 @@ Additional arguments:
   :code:`--dry-run`
     Print summary of what would be tested and then exit.
 
+  :code:`--file-tag TAG`
+    If running an individual Bats file with :code:`--file`, and that file is a
+    :code:`test.bats` to accompany a directory in :code:`examples`,
+    :code:`ch-test` is not smart enough to infer the image tag. Thus, you must
+    specify it with this option.
+
```
j-ogas: I had commented earlier about a followup, however, the more I think about it the more I think the `--file-tag` option should be smarter. In our `run` phase we already have some logic that 1) gets a tag from a `CH_TEST_TARDIR` artifact, 2) searches for a corresponding dockerfile, Build script, etc., and 3) modifies the `example` and `image` variables responsible for generating build and run phase tests.

How hard would it be to update this logic to work with a `--file-tag` argument?

I think it would flow much better to specify a tag and phase.

reidpr: The issue I ran into is that a given `test.bats` can have more than one Dockerfile or `Build.*`, so it's not possible to derive a test tag from a `test.bats`. Also, I'm not sure how to map a tag to non-`test.bats` files, e.g. something in `test/run` or `test/build`.

j-ogas: That seems right. My thought is that `--file` and `--file-tag` can be used separately or together. For example, if you want to target a file, e.g., `test.bats`, `build_auto.bats`, `run/ch-run_misc.bats`, .etc you would use `--file` tag (don't think a tag needs to be specified with those?) However, if you want to test an example by tag, you could do so with just `--file-tag`. In this case, both `ch-test full --file-tag mpihello`.

Does that cover your issue? The tag is what matters when multiple files share a `test.bats`, not sure about `Build.*` though.

bin/ch-test
```diff
@@ -784,7 +822,20 @@ else
 fi
 printf 'contributor:  %s\n\n' "$ch_contributor_note"
 
+if [[ $phase = one-file ]]; then
+    if [[ ! -f $one_file ]]; then
+        fatal "not a file: $one_file"
+    fi
+    one_file=$(readlink -f "$one_file")  # make absolute b/c we cd later
+    if [[ $one_file = */test.bats && -z $one_file_tag ]]; then
+        fatal 'must specify --file-tag for --file=*/test.bats'
```
j-ogas: I believe this if condition and error message needs an update. Other than that, looks good to me.

reidpr: Good catch; thank you.

Issue #715, opened 2020-04-15
Dockerfiles could use a "tidy"


heasterday:
It was noticed that some of the Dockerfiles result in more layers than necessary and the source is occasionally removed in a different layer than where it is installed.

An example is Dockerfile.openmpi
```dockerfile
# UCX. There is stuff to build Debian packages, but it seems not too polished.
ENV UCX_VERSION 1.6.1
RUN git clone --branch v${UCX_VERSION} --depth 1 \
              https://github.com/openucx/ucx.git
RUN    cd ucx \
    && ./autogen.sh \
    && ./contrib/configure-release-mt --prefix=/usr/local \
    && make -j$(getconf _NPROCESSORS_ONLN) install
RUN rm -Rf ucx*
```
This likely should just be and ENV and then a single RUN.

PR #726, opened 2020-04-23
Tidy example images

heasterday:
Addresses #715

It looks like the spack image could use a refactor to move some of the testing to its bats file but that can be a separate issue.


examples/Dockerfile.centos7
```diff
@@ -31,8 +20,15 @@ RUN yum -y install \
 # impact. In the interest of time, we demonstrate this by upgrading a single
 # package, though "yum upgrade" seems to pretty reliably tickle the problem.
 # See issue #472.
-RUN          yum -y install fakeroot
-RUN fakeroot yum -y upgrade dbus
-
-# Clean up.
-RUN yum clean all
+ RUN    yum -y install epel-release \
+    && yum -y install \
```
j-ogas, line 24: These two `yum`s appear to be misaligned.

Additionally, I prefer an additional indentation on the packages. For example:

```
RUN    yum install epel-release \
    && yum install bats \
                   gcc \
                   make \
                   python36 \
                   [...]
 
```

@heasterday @reidpr What are your thoughts?

heasterday: You are correct, I went ahead and corrected the misalignment. Good catch!

As for the additional indentation I can see the appeal. I do worry it would look odd in cases like the following though from `Dockerfile.openmpi`
```dockerfile
RUN    dnf install -y --setopt=install_weak_deps=false \
           automake \
           file \
           gcc \
```
As this would become the following if I understand the new style you are suggesting.
```dockerfile
RUN    dnf install -y --setopt=install_weak_deps=false automake \
                                                                                        file \
                                                                                        gcc \
```

examples/lustre/Dockerfile
```diff
@@ -3,18 +3,17 @@
...
-ENV LUSTRE_URL=https://downloads.whamcloud.com/public/lustre/lustre-${LUSTRE_VERSION}/el7/client/RPMS/x86_64/
+ARG LUSTRE_VERSION=2.12.3
+ARG LUSTRE_URL=https://downloads.whamcloud.com/public/lustre/lustre-${LUSTRE_VERSION}/el7/client/RPMS/x86_64/
 
 RUN    wget ${LUSTRE_URL}/lustre-client-${LUSTRE_VERSION}-1.el7.x86_64.rpm \
-    && wget ${LUSTRE_URL}/kmod-lustre-client-${LUSTRE_VERSION}-1.el7.x86_64.rpm
-
-RUN yum localinstall -y *.rpm
+    && wget ${LUSTRE_URL}/kmod-lustre-client-${LUSTRE_VERSION}-1.el7.x86_64.rpm \
+    && yum localinstall -y *.rpm \
```
reidpr, line 18: Why `yum localinstall` instead of `rpm` directly?

examples/mpibench/Dockerfile.mpich
```diff
@@ -1,12 +1,13 @@
 # ch-test-scope: full
 FROM mpich
 
-RUN dnf install -y which
+RUN    dnf install -y which \
```
reidpr, line 4: `dnf` is missing `--setopt=install_weak_deps=false` here?

examples/paraview/Dockerfile
```diff
@@ -21,10 +22,10 @@ RUN pip3 install --no-binary=mpi4py \
 
 # Mesa. We need a version newer than Debian provides. Not sure which bug fix
 # resolved it but versions prior to 19 appear to have issues rendering on ARM.
```
reidpr, line 24: 1. CentOS now, right?
2. Do we need to build it manually still?

examples/spark/Dockerfile
```diff
@@ -19,11 +19,11 @@ RUN touch /usr/bin/ch-ssh
...
-ENV URLPATH https://www.apache.org/dist/spark/spark-2.4.5/
-ENV DIR spark-2.4.5-bin-hadoop2.7
-ENV TAR $DIR.tgz
-RUN wget -nv $URLPATH/$TAR
-RUN tar xf $TAR && mv $DIR /opt/spark && rm $TAR
+ARG URLPATH=https://www.apache.org/dist/spark/spark-2.4.5/
+ARG DIR=spark-2.4.5-bin-hadoop2.7
+ARG TAR=$DIR.tgz
+RUN    wget -nv $URLPATH/$TAR \
+    && tar xf $TAR && mv $DIR /opt/spark && rm $TAR
```
reidpr, line 26: Put the separate commands on multiple lines, for consistency & clarity.

reidpr:
> It looks like the spack image could use a refactor to move some of the testing to its bats file but that can be a separate issue.

I think the tests that are there need to test installing stuff, which seemed natural during build. But, I could be convinced otherwise.

examples/Dockerfile.openmpi
```diff
@@ -61,70 +61,67 @@ FROM centos8
...
+     libibverbs \
+     libibverbs-devel \
+     libibverbs-utils \
+     librdmacm \
+     librdmacm-devel \
+     rdma-core \
+     make \
+     numactl-devel \
+     wget \
+# Note that libpsm2 is x86-64 only so we skip if missing
```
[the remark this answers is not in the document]
heasterday: Good catch, this change was made locally but not pushed.

examples/Dockerfile.centos7
```diff
@@ -31,8 +20,15 @@ RUN yum -y install \
 # impact. In the interest of time, we demonstrate this by upgrading a single
 # package, though "yum upgrade" seems to pretty reliably tickle the problem.
 # See issue #472.
-RUN          yum -y install fakeroot
-RUN fakeroot yum -y upgrade dbus
-
-# Clean up.
-RUN yum clean all
+ RUN    yum -y install epel-release \
```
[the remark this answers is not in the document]
heasterday: Good catch

examples/lustre/Dockerfile
heasterday: It looks like `yum localinstall` will pull dependencies from repos whereas rpm will not. @Brofessional any thoughts?

examples/Dockerfile.centos7
j-ogas: I think for hairy cases like that it would be something like:

```
RUN    dnf install -y --setopt=install_weak_deps=false \
                      automake \
                      file \
                      gcc \
```

examples/mpibench/Dockerfile.mpich
heasterday: Since it's only installing `which` I didn't believe it had additional weak dependencies it was bringing in anyway, I may have missed them though.

examples/paraview/Dockerfile
heasterday: Corrected. We need 19+ which is newer than what is available to work around issues on ARM. I adjusted the comment to be more clear

examples/Dockerfile.mpich
```diff
@@ -8,33 +8,33 @@
...
+       make \
+       wget \
+    && dnf clean all
 
 WORKDIR /usr/local/src
 
 # We currently need our own patched patchelf; see issue #256.
-RUN git clone https://github.com/hpc/patchelf.git
-RUN    cd patchelf \
+RUN    git clone https://github.com/hpc/patchelf.git \
```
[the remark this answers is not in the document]
heasterday: I feel like the downloads are quite stable and if debugging needs to be done a local copy of the dockerfile should be created with each of the operations of concern as their own `RUN`. Additionally, with the download as a separate `RUN` its payload is now in a layer.

heasterday:
> I think the tests that are there need to test installing stuff, which seemed natural during build. But, I could be convinced otherwise.

It looks like most of the tests are actually functionality "smoke tests" as to whether or not the install was successful e.g.
```dockerfile
RUN which spack
RUN spack --version
RUN spack compiler find
RUN spack compiler list
RUN spack compiler list --scope=system
RUN spack compiler list --scope=user
RUN spack compilers
RUN spack spec charliecloud
```
Having each of these be a new layer in the image doesn't seem right when they can be in the bats file. It's also worth noting a couple of these operations them are in both the Dockerfile and the bats test.

These lines seem fitting in the dockerfile
```dockerfile
RUN spack install charliecloud+docs
RUN spack clean --all
```

examples/Dockerfile.openmpi
```diff
@@ -61,70 +61,67 @@ FROM centos8
...
+     libibumad-devel \
+     libibverbs \
+     libibverbs-devel \
+     libibverbs-utils \
+     librdmacm \
+     librdmacm-devel \
+     rdma-core \
+     make \
+     numactl-devel \
+     wget \
```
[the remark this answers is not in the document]
heasterday: I like same column as the options, it looks quite clean immediately following the `install`. @j-ogas thoughts?

examples/Dockerfile.centos7
heasterday: What about we remove the extra whitespace for the `&&` so something like this:
```dockerfile
RUN yum install -y epel-release \
 && yum install -y \
                bats \
                gcc \
                make \
                python36 \
                rpm-build \
                rpmlint \
                wget \
                fakeroot \
 && fakeroot yum -y upgrade dbus \
 && yum clean all
```
@reidpr @j-ogas

reidpr: I would be OK with this, provided we're consistent across the Dockerfiles.

examples/Dockerfile.mpich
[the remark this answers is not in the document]
reidpr: That's fair. Please include a tip about this in the contributors' guide; I think while debugging it is important to keep the downloads separate so we don't risk getting bann3d.

examples/mpibench/Dockerfile.mpich
reidpr: I think it would be good to be consistent about adding the no weak dependencies options for both yum and apt, since it will keep the images smaller and avoid wondering why it's missing in some cases.

heasterday: After thinking about it more I agree, it keeps folks from feeling like they need to lookup if a package has weak dependencies or not and consistency is good.

heasterday:
Merged `master` and worked around some other things that popped up, please re-review.

heasterday:
@reidpr Fixed the issue we discussed and this is now passing Travis.

Issue #500, opened 2019-08-01
update C code to take full advantage of C99


reidpr:
We build (or should be building) the C files with `-std=c11`, but we don't actually take advantage of C11 features. This issue is to:

* audit the code and make a checklist of changes to make it really C11
* lead discussion on what to do
* do it

rstyd:
Here are features of C11 that could apply to charliecloud

    * A reimplementation of fopen called fopen_s
        - More secure than fopen
        - Instead of returning a FILE pointer, fopen_s accepts a FILE pointer as an argument
        - Validates parameters are invalid and errors out instead of providing undefined behavior

    * New implementations of the various formatted output and input functions
        - Ones we include are printf, fprintf, asprintf, dprintf
        - Each of these are appended with _s
        - Much like fopen_s they provide a more secure interface than the original

    * A native C thread library and atomic types
        - Doesn't support semaphores so we would have to rewrite charliecloud.c
        - There isn't any killer feature we'd gain from this

    * Unicode strings
        - Provides native unicode support with u8"¿String?"
        - This would provide charliecloud with better support for other languages

    * A new _Noreturn keyword l
        - The compiler can make additional optimizations when _noreturn is prepended in front of a void function

    * Type-generic macros
        - Gives access to macros that can accept any type
            #define fabs(X) _Generic( (X), \
                                 long double: fabsl, \
                                 default: fabs, \
                                 float: fabsf \
                                ) (X)

    * static_assert
        - Works much like the current test functions we have
            - Prints the line number, file, and function where the error occurs
        - Requires a string literal which won't work for us since we can't use the msg function

     * Anonymous Structs and unions


Of these features, only unicode, static_assert, atomic types, and anonymous structs/unions are included in gcc. 

The annex K functions (fopen_s, printf_s) are pretty controversial. They're not included in gcc, and it appears they may be removed from future C specifications.

reidpr:
Thanks Rusty. I spent a little time poking too. I just looked at the [C99](https://en.wikipedia.org/wiki/C99) and [C11](https://en.wikipedia.org/wiki/C11_(C_standard_revision)) Wikipedia pages, which have nice change lists.

I did find some C99 features that we aren't using consistently but possibly should be:

* More specific [data types](https://en.wikipedia.org/wiki/C_data_types).
* Designated initializers (e.g. `struct foo = { .bar = 1, .baz = 2 }`).
* Compound literals.
* Variadic macros, we are already using.
[...]

PR #727, opened 2020-04-24
constify

reidpr:
Addresses #500, in part.

This PR is a partially complete implementation of adding `const` everywhere it fits. This keyword designates a variable as unmodifiable, though there is a lot of nuance ([1](https://en.wikipedia.org/wiki/Const_(computer_programming)), [2](https://softwareengineering.stackexchange.com/questions/204500/when-and-for-what-purposes-should-the-const-keyword-be-used-in-c-for-variables)) and compiler enforcement is [incomplete for double pointers](https://stackoverflow.com/questions/5055655)

Currently, `ch_misc.{c,h}` and `ch_core.{c.h}` are modified to add `const` to all pointer function arguments. `ch_misc.c` also adds `const` to non-pointer function arguments and local variables. If we like it, I will complete the implementation.

The advantage is that we convert some runtime errors into compile failures, though I did not find any such cases yet. The disadvantage is that it adds a fair amount of noise to the code.

### DIFF
--- a/bin/ch_core.c
+++ b/bin/ch_core.c
@@ -68,26 +68,29 @@ struct {
 
 /** Function prototypes (private) **/
 
-void bind_mount(char *src, char *dst, char *newroot,
-                enum bind_dep dep, unsigned long flags);
-void bind_mounts(struct bind *binds, char *newroot,
+void bind_mount(char const * const src, char const * const dst,
+                char const * const newroot, enum bind_dep dep,
+                unsigned long flags);
+void bind_mounts(struct bind const * const binds, char const * const newroot,
                  enum bind_dep dep, unsigned long flags);
-void enter_udss(struct container *c);
-void join_begin(int join_ct, char *join_tag);
-void join_namespace(pid_t pid, char *ns);
+void enter_udss(struct container * const c);
+void join_begin(int join_ct, char const * const join_tag);
+void join_namespace(pid_t pid, char const * const ns);
 void join_namespaces(pid_t pid);
 void join_end();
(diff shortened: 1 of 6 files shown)


bin/ch_core.c
```diff
@@ -330,7 +333,7 @@ void run_user_command(char *argv[], char *initial_dir)
 
 /* Wait for semaphore sem for up to timeout seconds. If timeout or an error,
    exit unsuccessfully. */
-void sem_timedwait_relative(sem_t *sem, int timeout)
+void sem_timedwait_relative(sem_t * const sem, int const timeout)
```
rstyd, line 336: timeout should not be const. I'm of the opinion that only pointers should be const.

bin/ch_core.c
```diff
@@ -68,26 +68,29 @@ struct {
 
 /** Function prototypes (private) **/
 
-void bind_mount(char *src, char *dst, char *newroot,
-                enum bind_dep dep, unsigned long flags);
-void bind_mounts(struct bind *binds, char *newroot,
+void bind_mount(char const * const src, char const * const dst,
+                char const * const newroot, enum bind_dep dep,
+                unsigned long flags);
+void bind_mounts(struct bind const * const binds, char const * const newroot,
```
rstyd, line 74: In the meeting, we decided that const should only be used to ensure that functions can't change variables that are passed by reference. That would mean removing `const * const` since the second one (?) only affects what's inside the function.

rstyd:
Looks great! I really like the faq section.

Issue #655, opened 2020-02-21
ch-tug/ch-grow: pull from any public repo


reidpr:
Follow-up to PR #616.

Currently, `ch-tug` and the underlying library code is hard-coded to only work with public images from Docker Hub. It seems that the correct procedure is to ask for the resource we want, and then follow the instructions in the [WWW-Authenticate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate) header that comes back if we get HTTP 401. Example:

```console
$ wget -SO- https://registry-1.docker.io:443/v2/charliecloud/whiteout/manifests/2020-01-10 > /dev/null
[...]
  HTTP/1.1 401 Unauthorized
  Content-Type: application/json
  Docker-Distribution-Api-Version: registry/2.0
  Www-Authenticate: Bearer realm="https://auth.docker.io/token",service="registry.docker.io",scope="repository:charliecloud/whiteout:pull"
  Date: Fri, 21 Feb 2020 17:38:09 GMT
  Content-Length: 164
  Strict-Transport-Security: max-age=31536000
```

reidpr:
Note, this talks about "authentication" for anonymous pulling of public images, which is understandably confusing. What is going on is that is you indeed have to authenticate even to access public images anonymously, by requesting a token (without providing any actual authentication information e.g. username/password!) and the `WWW-Authenticate` header tells you where to go for that request.

PR #734, opened 2020-05-11
pull: Authenticate using instructions in WWW-Authenticate

reidpr:
Address #655 and #737.

### DIFF
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -196,14 +196,11 @@ class Image:
       self.unpack_create_ok()
       shutil.copytree(other.unpack_path, self.unpack_path, symlinks=True)
 
-   def download(self, use_cache=True):
+   def download(self, use_cache):
       """Download image manifest and layers according to origin and put them
          in the download cache. By default, any components already in the
          cache are skipped; if use_cache is False, download them anyway,
          overwriting what's in the cache."""
-      def _url(type_, address):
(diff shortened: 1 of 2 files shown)


trandles-lanl:
Works for me from my yellow linux workstation. I tried from fg-fey and it didn't work but that's because HTTP access to git.lanl.gov isn't enabled, only ssh access works. Is it possible to pull an image via ssh?

reidpr:
I thought push/pull was an HTTP(S) only protocol? I could be wrong though.

lib/charliecloud.py
```diff
@@ -597,47 +587,145 @@ def from_tree(self, t):
...
+         fp.write(res.content)
+         fp.close()
+      except OSError as x:
+         FATAL("can't write: %s: %s" % (path, x))
+
+   def get_layer(self, hash_, path):
+      "GET the layer with hash hash_ and save it at path."
+      # /v1/library/hello-world/blobs/<layer-hash>
+      url = self._url_of("blobs", "sha256:" + hash_)
+      accept = "application/vnd.docker.image.rootfs.diff.tar.gzip"
```
j-ogas, line 697: If we don't already, we should include a URL where this media type stems from.

reidpr: Turns out we do, on line 208 in `Image.download()`.

PR #745, opened 2020-05-21
fix example bugs

j-ogas:
Unclear why the netcdf package disappeared from spack. Replaced it with hdf5.


reidpr:
This PR also corrects some paths in the full scope.

Issue #645, opened 2020-02-14
documented dependencies should mention ShellCheck


rstyd:
I noticed the Contributor's Guide doesn't mention ShellCheck. It should list ShellCheck as a development dependency  and state which version we require. Installation instructions are located at https://github.com/koalaman/shellcheck/tree/v0.7.0#installing.

reidpr:
This should go in configure, since it's needed to run the complete test suite, and also the installation instructions so we can provide a link. It's a bit hidden how to find the static binary.

PR #746, opened 2020-05-21
document test suite dependency on ShellCheck

reidpr:
Addresses #645.

I really wanted to put it in the "recommended tests with tarballs" section, but it seemed too much. Opinions?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -306,6 +306,13 @@ EOF
   AC_MSG_RESULT($have_requests)
 ])
 
+# ShellCheck
+vmin_shellcheck=0.6.0
+AC_CHECK_PROG([SHELLCHECK], [shellcheck], [shellcheck])
+# https://stackoverflow.com/questions/6022384
+CH_CHECK_VERSION([SHELLCHECK], [$vmin_shellcheck],
+                 [--version | sed -n '2{s/^version: //;p;q}'])
+
 # SquashFUSE
 vmin_squashfuse=0.1.100  # Ubuntu 16.04 (Xenial). CentOS 7 has 0.1.102.
 AC_CHECK_PROG([SQUASHFUSE], [squashfuse], [squashfuse])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -306,6 +306,13 @@ EOF
   AC_MSG_RESULT($have_requests)
 ])
 
+# ShellCheck
+vmin_shellcheck=0.6.0
+AC_CHECK_PROG([SHELLCHECK], [shellcheck], [shellcheck])
+# https://stackoverflow.com/questions/6022384
+CH_CHECK_VERSION([SHELLCHECK], [$vmin_shellcheck],
+                 [--version | sed -n '2{s/^version: //;p;q}'])
```
rstyd, line 314: Can you explain this sed command?

reidpr: Yes; all the `CH_CHECK_VERSION` calls have something similar, to pull out a stand-alone version number without all the clutter from bare `--version`. If you're asking what the `-n` and `p;q` does, that *should* be explained in the SO link but I can clarify if needed.

configure.ac
```diff
@@ -610,5 +618,6 @@ Test suite
   complete test suite: ${have_tests_all}
     recommended tests with SquashFS ... ${have_tests_squash}
     documentation ... ${have_docs}
+    ShellCheck ≥ $vmin_shellcheck ... ${SHELLCHECK_VERSION_NOTE}
```
j-ogas, line 621: Where is `$SHELLCHECK_VERSION_NOTE` being defined?

configure.ac
rstyd: Ah. Perfect!

configure.ac
reidpr: Sorry, my comment got eaten. It is part of the magic of `CH_CHECK_VERSION`.

Issue #751, opened 2020-06-03
ch-tug: fails on registry.fedoraproject.org


reidpr:
```
$ ch-tug registry.fedoraproject.org/f29/httpd
pulling image:   registry.fedoraproject.org/f29/httpd
destination:     /var/tmp/ch-grow/img/registry.fedoraproject.org%f29%httpd

manifest: downloading
error: HTTP GET failed; expected status 401 but got 200: OK
```

trandles-lanl:
More detailed output:
```
[tcr@case ~]$ ch-tug -v registry.fedoraproject.org/f29/httpd
pulling image:   registry.fedoraproject.org/f29/httpd
destination:     /var/tmp/ch-grow/img/registry.fedoraproject.org%f29%httpd

use cache:       True
download cache:  /var/tmp/ch-grow/dlcache
manifest:        /var/tmp/ch-grow/dlcache/registry.fedoraproject.org%f29%httpd.manifest.json

downloading image: registry.fedoraproject.org:443/f29/httpd:latest
ensuring directory: /var/tmp/ch-grow/dlcache
manifest: downloading
GETting: https://registry.fedoraproject.org:443/v2/f29/httpd/manifests/latest
initializing session
requesting auth parameters
error: HTTP GET failed; expected status 401 but got 200: OK
[tcr@case ~]$
```

PR #752, opened 2020-06-03
pull: deal with registries that don't require a token

reidpr:
Addresses #751.

Apparently the Fedora registries do not require a token for anonymous pull. This PR adds null authentication for such registries.


reidpr:
Note, this diff looks bigger than it is. If you hide whitespace changes, that will make the actual changes more clear.

Issue #603, opened 2019-12-20
build: "make installcheck" passes


reidpr:
During PR #589, we didn't try `make installcheck`, and I'm almost certain it will fail. It should pass, and we should verify it passes in Travis.

j-ogas:
Looks like `make check` and `make installcheck` always succeed. The former will make if you haven't yet and the latter appears to check the source directory instead of the `--prefix` install path.

reidpr:
`make check` [needs to be told](https://www.gnu.org/software/automake/manual/html_node/Tests.html) what the tests are. 

I wasn't able to figure out exactly what `make installcheck` does. I suspect it does the checks after a standard `make install`, i.e. using `--prefix`, but I wasn't able to verify this.

j-ogas:
> `make check` [needs to be told](https://www.gnu.org/software/automake/manual/html_node/Tests.html) what the tests are. There is a lot in there, but I think we should just run `ch-test`.

Wouldn't that introduce bats as a build dependency?

PR #753, opened 2020-06-04
automake: implement make check and make installcheck

j-ogas:
Addresses: #603.

Note that when `make check` executes all the output is piped into a `test-suite.log` file leaving the terminal inactive for several minutes before reporting success.


bin/ch-test
```diff
@@ -840,10 +839,22 @@ if [[ $phase = one-file ]]; then
...
+if [[ -z $phase && -n $CH_TEST_PHASE ]]; then
+    case $CH_TEST_PHASE in
+    all|build|examples|run)
+        phase=$CH_TEST_PHASE
+    ;;
+    *)
+    ;;
+    esac
+fi
+
```
reidpr, line 852: I think we should just copy `$CH_TEST_PHASE` to `$phase` without checking it. That will be sufficient for internal purposes, and at some point in the future we can write code that checks for valid phase just once.

test/build/10_sanity.bats
```diff
@@ -50,10 +50,20 @@ load ../common
             ch-checkns)
                 ;;
             *)
+                # CH_TEST_PHASE is an undocumented variable that allows us to
+                # execute ch-test with the automake without arguments.
+                local phase
+                if [[ -n $CH_TEST_PHASE ]]; then
+                    phase=$CH_TEST_PHASE
+                    unset CH_TEST_PHASE
+                fi
```
reidpr, line 59: It's unclear to me what adding this tests?

.gitignore
```diff
@@ -60,10 +60,13 @@ Makefile.in
 /aclocal.m4
 /autom4te.cache/
 /bin/.deps/
+/bin/ch-test.log
+/bin/ch-test.trs
```
reidpr, line 64: These will need to be added to `autogen.sh` too (see comment).

test/build/10_sanity.bats
j-ogas: This prevents the `executables sane` test from failing and bats from executing `ch-test all`. The documented and expected behavior for `./bin/ch-*` executables is print usage and exit. Without the above, `ch-test` will execute with the phase specified in `CH_TEST_PHASE` when executed with `make check`.

reidpr: Aha, OK. Would it work to simply `unset CH_TEST_PHASE`?

j-ogas:
It seems we are unable to redirect the output generated by `make check` to `stdout` instead of `test-suite.log`. This lack of terminal activity makes our travis tests time out.

Marking as differed and closing.

Issue #740, opened 2020-05-13
push images to container repository


j-ogas:
Now that ch-tug can authenticate with our internal repositories using PAT and cryptocard (#737), it would be nice to be able to push.

Related issue: #583

See docker v2 api push: https://docs.docker.com/registry/spec/api/#pushing-an-image

trandles-lanl:
Rough estimate on release date? A resolution of month/year would be good for my purposes.

PR #756, opened 2020-06-09
new feature: ch-nudge

j-ogas:
Addresses: #740.


doc/ch-nudge_desc.rst
```diff
@@ -0,0 +1,103 @@
...
+underlying code with :code:`ch-grow(1)`.
+
+:code:`ACTION` must be one of the following:
+
+  :code:`delete IMAGE_REF`
+    TODO
+  :code:`list IMAGE_REF`
+    TODO
+  :code:`push IMAGE_REF`
+    TODO
```
reidpr, line 39: I wonder if we should also consider this interface:

```
$ ch-nudge --delete IMAGE_REF
$ ch-nudge --list IMAGE_REF
$ ch-nudge IMAGE_REF [DEST_IMAGE_REF]
```

i.e.:

1. default is push
2. would need to document that list should not have a tag or hash

doc/ch-nudge_desc.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Other arguments:
+
+  :code:`-h`, :code:`--help`
+    Print help and exit.
+
+  :code:`--chunk SIZE`
+    Upload layer in chunks of size :code:`SIZE`. This is typically unnecessary
+    and should be used only by users who are familiar with image uploading and
+    why this may useful.
```
reidpr, line 51: Would we ever need this? If not, we should leave it out.

doc/ch-nudge_desc.rst
```diff
@@ -0,0 +1,103 @@
...
+Examples
+========
+
+Push image "hello-world" to the Charliecloud image repository on the Docker
+registry.
+
+::
+
+  $ ch-nudge push registry-1.docker.io:443/charliecloud/hello-world
+  pushing image: hello-world
```
reidpr, line 79: I'd print the full image ref being used here, like `ch-tug` does.

doc/ch-nudge_desc.rst
```diff
@@ -0,0 +1,103 @@
...
+  deleting image: hello-world
+  [...]
+
+List all available tags in the centos repository on the Docker registry.
+
+::
+
+  $ ch-nudge list registry-1.docker.io:443/charliecloud/tags/list
+  listing tags
+  [...]
```
reidpr, line 97: For this one I think example output would be especially good. I do wonder about more complete output for all the examples, though I think the way to put that in is copy-and-paste later.

doc/ch-nudge_desc.rst
```diff
@@ -0,0 +1,103 @@
...
+
+List all available repositories from the Docker registry.
+
+::
+
+  $ ch-nudge list registry-1.docker.io:443/_catalog
```
reidpr, line 103: The `_catalog` is pretty inside baseball. If we have list capabilities at all, then the syntax here I'd suggest something like:

```
$ ch-nudge list registry-1.docker.io:443
```

That said, who actually knows what `registry-1.docker.io` means?

Is listing by browsing the registry website adequate?

j-ogas:
Closing in favor of #867.

Issue #677, opened 2020-02-27
/test/dev_proc_sys.py fails on kernels built without CONFIG_STRICT_DEVMEM=y


heasterday:
It appears when a REHL7/CentOS7 system with a recent kernel (4.x) is built without `CONFIG_STRICT_DEVMEM=y` then `/dev/mem` is not present which is tested by `dev_proc_sys.py` leading to test failure.

```
 ✗ /dev /proc /sys
   (in test file run/ch-run_uidgid.bats, line 49)
     `ch-run $uid_args $gid_args "$ch_timg" -- /test/dev_proc_sys.py' failed
   no ID arguments
   
   ERROR	/dev/mem: exception: [Errno 2] No such file or directory: '/dev/mem'
   SAFE	/proc/kcore: read not allowed
   SAFE	/sys/kernel/mm/page_idle/bitmap: read not allowed
```

Relevant: http://man7.org/linux/man-pages/man4/mem.4.html

heasterday:
It's worth noting both of the systems presenting the issue are aarch64.

reidpr:
Good find.

Can you figure out an alternate device that unprivileged users shouldn't be able to read? Ideally it would be broadly available and could simply replace `/dev/mem`, but if we have to try more than one, it's OK too.

heasterday:
@reidpr the device `/dev/cpu_dma_latency` looks like a decent candidate to me. This device is part of the [power management quality of service framework](https://www.kernel.org/doc/html/v5.0/admin-guide/pm/cpuidle.html) in the linux kernel, appears to be ubiquitous, and requires privilege to access. 

I think trying `/dev/mem` and if it is not found `/dev/cpu_dma_latency` like what is done with the `/proc` test should resolve this.

reidpr:
Sounds good to me.

PR #762, opened 2020-06-11
dev_proc_sys.py: check /dev/cpu_dma_latency in addition to /dev/mem

heasterday:
Addresses #677 

I just duplicated the existing code checking multiple paths for /sys and used it for /dev, I could rework it to be based around a function to reduce code duplication but that would increase complexity for what is a simple script. If folks prefer the function route let me know and I can rework.

### DIFF
--- a/examples/chtest/dev_proc_sys.py
+++ b/examples/chtest/dev_proc_sys.py
@@ -3,11 +3,12 @@
 import os.path
 import sys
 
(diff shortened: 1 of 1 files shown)


heasterday:
@reidpr I believe the failure will be resolved by the merge of #726, once that is merged I will merge master and re-check the tests.

heasterday:
@reidpr This passed, good to merge.

Issue #792, opened 2020-06-26
ch-grow: document a minimum version of lark-parser


j-ogas:
We currently do not document a minimum version of lark-parser. 0.7.8 does not work.

rstyd:
Agreed: the current minimum working version is `0.8.6`. In addition, we should add a check on the current lark-parser version to `configure.ac`. Currently, we just check if lark-parser exists on the system with `${have_lark}`. This was sufficient before `0.8.6` changed the instruction order.

reidpr:
Last I checked 0.8.5 worked too. A binary search shouldn't be too hard.

reidpr:
Distro packages:

* CentOS 7 has 0.7.1 via EPEL; CentOS 8 has 0.8.1 via EPEL.
* Debian Bullseye (test) has 0.8.6 and Sid (unstable) has 0.8.9; does not appear to be in any stable.
* Ubuntu Focal (20.04 LTS) has 0.8.1 and Groovy (20.10?) has 0.8.8.

rstyd:
In that case, should we just suggest users acquire it via `pip`?

reidpr:
Using `test/Dockerfile.argenv` and current master; `lark-parser` installed via `pip3`:

```
0.8.9  ok
0.8.8  ok
0.8.7  ok
0.8.6  ok
0.8.5  ok
0.8.4  ok
0.8.3  ok
0.8.2  ok
0.8.1  ok
0.8.0  ok
0.7.8  fail: no attribute 'visit_topdown'
```

reidpr:
Same, but replace `visit_topdown()` with `visit()`:

```
0.8.9  fail: out of order
0.8.6  fail: out of order
0.8.5  ok
0.8.1  ok
0.8.0  ok
0.7.8  ok
0.7.7  ok
0.7.5  ok
0.7.4  ok
0.7.3  ok
0.7.2  ok
0.7.1  ok
```

reidpr:
I propose we:

1. Document 0.7.1 as the minimum version.
2. Add a small monkey patch: if `lark.Visitor.visit_topdown()` does not exist, add it as an alias to `visit()`. Or we could just call `visit()` directly.

This would make it work with the `lark-parser` in EPEL7.

rstyd:
I think those are both excellent ideas.

PR #797, opened 2020-07-10
clean up Python module version dependencies

reidpr:
This addresses #792 but also stands alone because it cleans things up and adds a version for requests too.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -180,18 +180,17 @@ AS_IF([test -n "$SPHINX"], [
 
 # "docutils" module
 vmin_docutils=0.14
-docutils=$sphinx_python  # FIXME: output is confusing
 AS_IF([test -n "$SPHINX"], [
   # Sphinx depends on docutils, so we don't need to check if the module exists
   # before checking its version. (CH_CHECK_VERSION isn't smart enough to deal
   # with Python being present but a module not.)
-  CH_CHECK_VERSION([docutils], [$vmin_docutils],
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -301,9 +300,11 @@ except Exception:
    sys.exit(1)
 EOF
   AS_IF([test $? -eq 0],
-    [have_requests=yes],
-    [have_requests=no])
+    [REQUESTS=$PYTHON; have_requests=yes],
+    [REQUEST=;         have_requests=no])
```
j-ogas, line 304: Should these variables be the same?
```suggestion
    [REQUESTS=$PYTHON; have_requests=yes],
    [REQUESTS=;         have_requests=no])
```

configure.ac
```diff
@@ -285,12 +281,15 @@ except Exception:
...
-  AC_MSG_RESULT($have_lark)
+    [LARK=$PYTHON; have_lark=yes],
+    [LARK=;        have_lark=no])
+  AC_MSG_RESULT([$have_lark])
+  CH_CHECK_VERSION([LARK], [$vmin_lark],
+    [-c 'import lark; print(lark.__version__)'])
 ])
 
 # Python module "requests"
+vmin_requests=2.6.0  # CentOS 7; FIXME: haven't actually tested this
```
rstyd, line 292: I checked this on CentOS 7, and it does indeed work.

Issue #798, opened 2020-07-11
configure: fails checking for asprintf(3) under Clang


olifre:
As found in:
https://bugs.gentoo.org/732214
configure tests fail with clang, since clang has `-Wformat` as part of `-Wall`, and the check testing for `asprintf` complains with a format warning, which becomes an error due to `-Werror`:
```
configure:4946: checking for asprintf
configure:4946: x86_64-pc-linux-gnu-clang -o conftest -O2 -pipe -march=native -frecord-gcc-switches -fno-diagnostics-color -std=c99 -Wpedantic -Wall -Werror  -Wl,-O1 -Wl,--as-needed -fuse-ld=lld -Wl,--defsym=__gentoo_check_ldflags__=0 conftest.c  >&5
conftest.c:46:18: error: format string missing [-Werror,-Wformat]
return asprintf ();
       ~~~~~~~~  ^
1 error generated.
[...]
| int
| main ()
| {
| return asprintf ();
|   ;
|   return 0;
| }
configure:4946: result: no
configure:4954: error: asprintf() not found. Please report this bug.
```
I'm not sure how to best fix this (I'm not an autotools expert), probably dropping `-Werror` for the test or using `-Wno-error=format` might be easiest, or using a different test for `printf`-like functions which uses a dummy format string.

junghans:
Over the years I have patched `-Werror` out of some many packages, I personally like to run all the CI with `-Werror`, but don't inject in by default.

reidpr:
Thanks Oliver and Christoph!

The general principle I like is to have diagnostics that are as noisy as practical. That's the reason we do `-Wall -Werror`.

As for the `asprintf(3)` test, IMO that's definitely a broken test. The format string isn't optional. I think that it's indeed our fault; we do not seem to instruct Autoconf to use an argument:

https://github.com/hpc/charliecloud/blob/254fa07a2241ea0349179ea42471b1abdefafeea/configure.ac#L158

I'm not sure exactly how to fix it, as a quick look at the `AC_CHECK_FUNC` documentation doesn't say anything about function arguments.

Let's have the focus of this issue be fixing that test.

As for `-Werror` generally, I would be very open to a PR that added an argument to `configure` that replaced or extended `-Werror`, or otherwise cleaned up how we deal with `CFLAGS`. E.g., in this case it could be `-Wno-error=format`. This should be changeable more easily than a patch.

[...]

PR #799, opened 2020-07-13
configure: allow to disable Werror

junghans:
For #798

PR #800, opened 2020-07-13
add custom check for asprintf(3)

reidpr:
Addresses #798.

@olifre, @junghans : Does this fix the `configure` failure under Clang?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -165,12 +165,14 @@ AC_MSG_CHECKING([for asprintf])
 AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
     #define _GNU_SOURCE
     #include <stdio.h>
+    #include <stdlib.h>
 
     int main(void)
     {
-        char *p;
-        asprintf(&p, "WEIRD AL YANKOVIC\n");
-        return 0;
+       char *p;
+       if (asprintf(&p, "WEIRD AL YANKOVIC\n") >= 0)
+          free(p);
+       return 0;
     }
   ]])],
   [AC_MSG_RESULT([yes])],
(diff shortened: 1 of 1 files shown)


olifre:
Indeed, that fixes the configure test for me with clang! :-).

olifre:
ah, but now `gcc` is unhappy (as travis also sees):
```
configure:4889: gcc -c -g -O2 -std=c99 -Wpedantic -Wall -Werror  conftest.c >&5
conftest.c: In function 'main':
conftest.c:19:9: error: ignoring return value of 'asprintf', declared with attribute warn_unused_result [-Werror=unused-result]
   19 |         asprintf(&p, "WEIRD AL YANKOVIC\n");
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```
Interestingly, `clang` does not care about this.

reidpr:
That's weird; my local GCC doesn't care either.

olifre:
I'm on gcc 9.3.0 here, which seems to default to `-Wunused-result`. Of course, it's right — usually the return value of `asprintf` should not be ignored. 
Changing the test to just `return` the `int` produced by `asprintf` seems to fix things, making both compilers happy ;-).

olifre:
@reidpr Nice, that's fore sure the syntactically more correct and instructive solution. I can confirm this works for me with my gcc 9.3.0 and with Clang here.

reidpr:
Merged. Leaving the original bug open since there is another PR pending too (#799).

reidpr:
Actually, I am going to close the original bug since as written it does fix the issue. I'll tag the other PR.

PR #804, opened 2020-07-16
ch-run: fuse-based squashfs automount integration

chernikovanna:
This is what we have so far. All documentation has been updated with the current usage for squash filesystems.


reidpr:
Per discussion, this excellent work is going to make its way in via other means.

pramodk:
This PR looks very useful esp because, IIUC, we can directly run squashfs images without need of extracting! From our recent experience, untaring images and copying them around is quite inconvenient (as it's slow on parallel fs).

> Per discussion, this excellent work is going to make its way in via other means.

Could you briefly elaborate what is the plan for this? (and what do you mean by other means?)

reidpr:
> Could you briefly elaborate what is the plan for this? (and what do you mean by other means?)

Sure. This PR was a prototype, and libsquashfuse has changed in the time since. SquashFUSE automount is being actively worked in PR #1102, which I hope to merge sometime later this summer.

And yes, you've nailed the use case exactly.

Does that clarify?

pramodk:
Yes. Great! thanks for clarification! 

And, @mphinney1100 - thank you very much in advance for adding this feature! ⭐

PR #805, opened 2020-07-20
ch-pull2dir: tidy use of sed

paulbry:
Just a minor change I wanted to propose to the `ch-pull2dir` script after I ran into this issue this past weekend using it. With image names containing multiple `/` characters (e.g. from GitLab container registries) they hadn't been replaced with the `sed` command. Additionally, these will be observed by the `rm` used to remove the original tarball.


reidpr:
Closing to kick Travis? Will re-open.

paulbry:
Thanks. Yeah I took a look at the failing tests but they didn't seem related.

PR #808, opened 2020-07-21
configure: add --enable-buggy-build to omit -Werror

reidpr:
Extends PR #799 with a different option name and some documentation of the option.

@junghans : Does this meet your needs?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -35,7 +35,7 @@ AC_CONFIG_FILES([Makefile
 
 # Need a C99 compiler. (See https://stackoverflow.com/a/28558338.)
 AC_PROG_CC
-AC_DEFUN([CH_CFLAGS], [-std=c99 -Wpedantic -Wall -Werror])
+AC_DEFUN([CH_CFLAGS], [-std=c99 -Wpedantic -Wall])
 AX_CHECK_COMPILE_FLAG([CH_CFLAGS], [
   CFLAGS+=" CH_CFLAGS"
 ], [
@@ -44,6 +44,9 @@ AX_CHECK_COMPILE_FLAG([CH_CFLAGS], [
 AS_IF([test "$CC" = icc],
       [AC_MSG_ERROR([icc not supported (see PR @%:@481)])])
 
+AC_ARG_ENABLE([Werror],
+  AS_HELP_STRING([--disable-Werror], [do not inject -Werror]),
+  [], [CFLAGS+=" -Werror"])
 
 ## Libraries
 
--- a/configure.ac
+++ b/configure.ac
@@ -35,18 +35,32 @@ AC_CONFIG_FILES([Makefile
 
 # Need a C99 compiler. (See https://stackoverflow.com/a/28558338.)
 AC_PROG_CC
-AC_DEFUN([CH_CFLAGS], [-std=c99 -Wpedantic -Wall])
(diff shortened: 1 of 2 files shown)


trandles-lanl:
After reading all of the various links, PRs, and the original issue I _think_ I'm persuaded by the @Flameeyes blog post. But I also agree with @junghans that this PR fixes the immediate problem and that's probably good enough. I approve.

If I were BDFL (@reidpr) I might be tempted to make disabling ```-Werror``` the default and have a configure flag to enable it. This way, if I'm some naive user who is trying to build Charliecloud from source, or someone packaging it for a distribution, I'm not caught out by the warnings. If I'm a developer on Charliecloud and want the extra strict checking then I would enable ```-Werror``` (and whatever else comes up later) using a flag like ```--enable-extra-checks```.

reidpr:
OK, thanks all for the comments. If we could actually test with a wide variety of compilers, I would set the default differently, but we can't.

One thing I haven't yet seen addressed in the arguments is how new compiler warnings will reliably be reported to us without `-Werror`.

I revised the docs a little in response to the feedback and the new blog post.

@junghans @olifre I would greatly appreciate if you use `--enable-buggy-build` only for the releases that actually need it, not routinely. Please recall that we fixed #798 quickly, and we'll be happy to do so again in the future. It is important to me that Charliecloud build on Gentoo with zero warnings, including the configure tests, which seems to hide warnings even with `-Wall`.

olifre:
@reidpr I think for Gentoo packages I basically have to use the flag, to adhere to the policy. But what I will surely do is test compilation on Gentoo with the latest stable GCC packaged for Gentoo, and latest Clang, on several of my machines (with `-Werror`), so I will definitely extend compiler test coverage :-). 

Sadly, I also do not have an answer to your (true) statement:

> One thing I haven't yet seen addressed in the arguments is how new compiler warnings will reliably be reported to us without `-Werror`.

I don't think distros foresee such a way, unless packagers do it manually. I certainly will for all platforms I can test on.

reidpr:
> But what I will surely do is test compilation on Gentoo with the latest stable GCC packaged for Gentoo, and latest Clang, on several of my machines (with `-Werror`), so I will definitely extend compiler test coverage :-).

Fantastic, thanks so much! I definitely owe you a lot of beers.

junghans:
> @reidpr I think for Gentoo packages I basically have to use the flag, to adhere to the policy. But what I will surely do is test compilation on Gentoo with the latest stable GCC packaged for Gentoo, and latest Clang, on several of my machines (with `-Werror`), so I will definitely extend compiler test coverage :-).
> 
> Sadly, I also do not have an answer to your (true) statement:
> 
> > One thing I haven't yet seen addressed in the arguments is how new compiler warnings will reliably be reported to us without `-Werror`.
> 
> I don't think distros foresee such a way, unless packagers do it manually. I certainly will for all platforms I can test on.

Gentoo has the `custom-flags` use flag for that.

olifre:
> Gentoo has the `custom-flags` use flag for that.

Thanks for the pointer! Is it really meant for this purpose, though (i.e. enabling upstream's preferred cflags)? My understanding of the flag was that it should rather ensure the user's custom cflags are honoured (even if these may be considered "unsafe"). 
Checking though existing uses of this flag, I find it mostly disables strip-flags, and in case of `sys-boot/syslinux` it even removes upstream's cflags to ensure the user's cflags are honoured.

Issue #806, opened 2020-07-20
ch-grow: make --version and --help work if requests isn't installed


trandles-lanl:
I get a traceback when passing a bad argument to ```ch-grow``` Looking at the traceback though, I'm not positive there isn't a different problem with 0.16 on fog.

```
[trandles@fg-fey1 ~]$ ch-run --version
0.16
[trandles@fg-fey1 ~]$ ch-grow --version
Traceback (most recent call last):
  File "/usr/projects/hpcsoft/toss3/common/x86_64/charliecloud/0.16/bin/ch-grow", line 17, in <module>
    import charliecloud as ch
  File "/usr/projects/hpcsoft/toss3/common/x86_64/charliecloud/0.16/bin/../lib/charliecloud/charliecloud.py", line 588, in <module>
    class Repo_Downloader:
  File "/usr/projects/hpcsoft/toss3/common/x86_64/charliecloud/0.16/bin/../lib/charliecloud/charliecloud.py", line 609, in Repo_Downloader
    class Bearer_Auth(requests.auth.AuthBase):
NameError: name 'requests' is not defined
```

reidpr:
Oh hmm; it is not supposed to actually need the dependencies to process `--version`, `--help`, etc.

trandles-lanl:
Looks like a problem with 0.16 on fog.

```
[trandles@fg-fey1 build]$ ls
Dockerfile.clamr
[trandles@fg-fey1 build]$ ch-grow -f Dockerfile.clamr .
Traceback (most recent call last):
  File "/usr/projects/hpcsoft/toss3/common/x86_64/charliecloud/0.16/bin/ch-grow", line 17, in <module>
    import charliecloud as ch
  File "/usr/projects/hpcsoft/toss3/common/x86_64/charliecloud/0.16/bin/../lib/charliecloud/charliecloud.py", line 588, in <module>
    class Repo_Downloader:
  File "/usr/projects/hpcsoft/toss3/common/x86_64/charliecloud/0.16/bin/../lib/charliecloud/charliecloud.py", line 609, in Repo_Downloader
    class Bearer_Auth(requests.auth.AuthBase):
NameError: name 'requests' is not defined
[trandles@fg-fey1 build]$
```

reidpr:
I can reproduce this on current master.

PR #811, opened 2020-07-31
ch-grow: don't crash if requests is not installed

reidpr:
Addresses #806.


reidpr:
Yeah, we do (at least we should) print an error if the user asks for something that actually does need the dependencies.

PR #817, opened 2020-08-10
tests: add ln -T for consistency with examples

reidpr:
As reported in spack/spack#17968, the `test` directory lacks argument `-T` for `ln(1)`.


reidpr:
@g-mathias, does this fix your bug?

g-mathias:
This is what I did in the patch to Makefile.in, which worked. 
I did not use the -T option for the libtest.so but it will not hurt. 
So the problem should be fixed with your changes.
Thanks for the quick reply.

reidpr:
close/reopen to re-run tests

Issue #814, opened 2020-08-07
ch-tug support for manifest schemaVersion 1


trandles-lanl:
I observe the following:

```
[tcr@malkovich ~]$ ch-tug -v quay.io/fenicsproject/stable:latest
pulling image:   quay.io/fenicsproject/stable:latest
destination:     /var/tmp/tcr/ch-grow/img/quay.io%fenicsproject%stable:latest
[latest.zip](https://github.com/hpc/charliecloud/files/5042356/latest.zip)

use cache:       True
download cache:  /var/tmp/tcr/ch-grow/dlcache
manifest:        /var/tmp/tcr/ch-grow/dlcache/quay.io%fenicsproject%stable:latest.manifest.json

downloading image: quay.io:443/fenicsproject/stable:latest
ensuring directory: /var/tmp/tcr/ch-grow/dlcache
manifest: downloading
GETting: https://quay.io:443/v2/fenicsproject/stable/manifests/latest
initializing session
requesting auth parameters
error: can't parse manifest file: /var/tmp/tcr/ch-grow/dlcache/quay.io%fenicsproject%stable:latest.manifest.json
[tcr@malkovich ~]$
```

The manifest retrieved is attached.
[latest.zip](https://github.com/hpc/charliecloud/files/5042360/latest.zip)

PR #818, opened 2020-08-14
ch-grow pull: add schema version one (1) pull capability

j-ogas:
Addresses #814.

This PR adds schema version one (1) image pulling. Note that pushing schema version one (1) images is deprecated and has been for quite some time. There are no plans to deprecate pulling schema version one (1) images (see: https://github.com/moby/moby/issues/39701).


lib/charliecloud.py
```diff
@@ -739,8 +789,17 @@ def get_layer(self, hash_, path):
...
+            '2': "application/vnd.docker.distribution.manifest.v2+json"}
+      # Use undocumented variable to specify schema version for testing.
+      if env_schema_version:
+            DEBUG('requesting schema version %s (environment)'
+                   % env_schema_version)
+            accept = sv[env_schema_version]
+      else:
+         accept = [sv['2'], sv['1']]
+      self.get(url, path, { "Accept": str(accept) })
+
```
reidpr, line 802: 1. Can we add weights to prefer v2 over v1 if we accept both?
2. If you write the logic like the following, I think it will be simpler:
```
sv = {}
if v1 is OK:
  sv['1'] = ...
if v2 is OK:
  sv['2'] = ...
assert (sv is not empty)
```

test/build/50_ch-tug.bats
```diff
@@ -234,3 +234,48 @@ EOF
...
+    scope standard
+    if ( ! ch-tug --dependencies ); then
+
+        [[ $CH_BUILDER != ch-grow ]]
+        skip "ch-tug missing dependencies"
+    fi
+
+    unpack=$BATS_TMPDIR
+    cache=$unpack/dlcache
+    img=charliecloud%symlink
```
reidpr, line 249: Why this image?

test/build/50_ch-tug.bats
```diff
@@ -234,3 +234,48 @@ EOF
...
+
+    export CH_SCHEMA_VER=1
```
reidpr, line 251: I don't think you need to export it; I think you can just set it for the `ch-tug` command.

test/build/50_ch-tug.bats
```diff
@@ -234,3 +234,48 @@ EOF
...
+    img1=$unpack/charliecloud%symlink%v1
+    ch-tug --unpack-dir="$img1" \
+           --dl-cache="$cache" \
+           --no-cache charliecloud/symlink
+    [[ $status -eq 0 ]]
+    cat "${cache}/charliecloud%symlink.manifest.json" | grep -F '"schemaVersion": 1'
```
reidpr, line 257: I'd remove `cat` here since `grep` takes a filename argument.

test/build/50_ch-tug.bats
```diff
@@ -234,3 +234,48 @@ EOF
...
+    [[ $status -eq 0 ]]
+    cat "${cache}/charliecloud%symlink.manifest.json" | grep -F '"schemaVersion": 2'
+
+    # The following horror ensures that the same image unpacked via different
+    # schemas are the same.
+    cd "$img1"
+    find "$img" -type f -exec md5sum {} + | sort > "${img1}/contents.txt"
+    cd "$img2"
+    find "$img" -type f -exec md5sum {} + | sort > "${img2}/contents.txt"
+    diff -u "${img1}/contents.txt" "${img2}/contents.txt"
```
reidpr, line 273: I think you can just compare two trees directly with `diff`.

test/build/50_ch-tug.bats
```diff
@@ -234,3 +234,48 @@ EOF
...
+    [[ $status -eq 0 ]]
+
+    # Ensure we error if we don't get the specified schem
+    run ch-tug --unpack-dir="$unpack" \
+               --dl-cache="$cache" \
+               --no-cache quay.io:443/fenicsproject/stable:latest
+    [[ $status -ne 0 ]]
```
reidpr, line 280: This to me feels like a separate test.

test/build/50_ch-tug.bats
j-ogas: `diff` complains about broken symlinks and errors with some flavor of`file doesn't exist` error. These symbolic links are pointing to locations that are relative to the container root and are resolvable after `pivot_root`. However, when we probe them in storage they point to places that may not exist.

It wasn't quite clear to tell `diff -r` to treat symbolic links as regular files, i.e., don't follow them.

lib/charliecloud.py
```diff
@@ -711,6 +750,16 @@ def authenticate_maybe(self, url):
...
+   def check_response_schema_maybe(self, url, response):
+      "Ensure the schema is what we requested"
+      if 'manifest' in url and env_schema_version:
+         content = response.headers['Content-Type']
+         DEBUG("preferred schema: v%s (environment)" % env_schema_version)
+         DEBUG("response schema content: %s" % content)
+         if not ("v" + env_schema_version) in content:
+            FATAL("requested schema v%s and response returned Content-Type: %s"
+                  % (env_schema_version, content))
+
```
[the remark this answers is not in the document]
j-ogas: The issue is we need the response object to determine whether we received what we asked for. `get_manifest` calls `get_raw` and the response is obtained in `get_raw`.

lib/charliecloud.py
```diff
@@ -301,10 +314,23 @@ def layer_hashes_load(self):
...
          FATAL("can't parse manifest file: %s:%d: %s"
                % (self.manifest_path, x.lineno, x.msg))
-      try:
-         self.layer_hashes = [i["digest"].split(":")[1] for i in doc["layers"]]
-      except (AttributeError, KeyError, IndexError):
-         FATAL("can't parse manifest file: %s" % self.manifest_path)
+
+      self.schema_version = doc['schemaVersion']
+      if self.schema_version == 1:
+         DEBUG('loading layer hashes from schema version one (1) manifest')
```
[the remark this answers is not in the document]
j-ogas: Here we are getting the value from the `schemaVersion` key directly from the manifest. I've updated everywhere else to use `v1` where appropriate.

lib/charliecloud.py
```diff
@@ -185,6 +196,7 @@ class Image:
                 "download_cache",
                 "image_subdir",
                 "layer_hashes",
+                "schema_version",
```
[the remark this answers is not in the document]
j-ogas: Not if we use the environment variable. Removed on my local branch.

test/build/50_ch-tug.bats
j-ogas: No specific reason. Do you prefer another?

lib/charliecloud.py
```diff
@@ -735,8 +748,9 @@ def get_layer(self, hash_, path):
    def get_manifest(self, path):
       "GET the manifest for the image and save it at path."
       url = self._url_of("manifests", self.ref.version)
-      accept = "application/vnd.docker.distribution.manifest.v2+json"
-      self.get(url, path, { "Accept": accept })
+      accept = ["application/vnd.docker.distribution.manifest.v1+json",
+                "application/vnd.docker.distribution.manifest.v2+json"]
+      self.get(url, path, { "Accept": str(accept) })
```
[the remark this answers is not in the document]
j-ogas: The latest commit changes to a v1 opt in policy. We request only v2 unless the user specifies otherwise.

test/build/50_ch-tug.bats
reidpr: Nothing specific in mind, but that image was created for another purpose, so we should justify its use here.

test/build/50_ch-tug.bats
reidpr: Does `--no-dereference` help?

lib/charliecloud.py
```diff
@@ -341,7 +368,17 @@ def layers_read(self):
          except tarfile.TarError as x:
             FATAL("cannot open: %s: %s" % (path, x))
          members = collections.OrderedDict([(m, None) for m in members_list])
-         layers[lh] = TT(fp, members)
+         if (lh in layers and members):
```
reidpr, line 371: I think the precedence here is `(lh in layers) and members`, not `lh in (layers and members)` which it seems this is looking for.

lib/charliecloud.py
```diff
@@ -341,7 +368,17 @@ def layers_read(self):
...
+            FATAL("duplicate non-empty layer %s" % lh[:7])
+         if (members):
```
reidpr, line 373: What is truth/falsity of a dictionary? Maybe be more explicit e.g. `if (len(members) > 0)`?

lib/charliecloud.py
```diff
@@ -341,7 +368,17 @@ def layers_read(self):
...
+            layers[lh] = TT(fp, members)
+         else:
+            empty_cnt += 1
+      if (self.schema_version == 'v1'):
+         DEBUG('reversing layer order for schema version one (1)')
+         layers = collections.OrderedDict(reversed(list(layers.items())))
```
reidpr, line 379: Can leave the `list` out here.

reidpr (changes requested):
Great work! A few small things.

lib/charliecloud.py
```diff
@@ -304,10 +306,25 @@ def layer_hashes_load(self):
       except json.JSONDecodeError as x:
          FATAL("can't parse manifest file: %s:%d: %s"
                % (self.manifest_path, x.lineno, x.msg))
-      try:
-         self.layer_hashes = [i["digest"].split(":")[1] for i in doc["layers"]]
-      except (AttributeError, KeyError, IndexError):
-         FATAL("can't parse manifest file: %s" % self.manifest_path)
+
+      schema_version = doc['schemaVersion']
```
reidpr, line 310: This needs to be protected against `KeyError`.

lib/charliecloud.py
```diff
@@ -304,10 +306,25 @@ def layer_hashes_load(self):
...
+      if (schema_version == 1):
+         DEBUG('loading layer hashes from schema version 1 manifest')
+         try:
+            self.layer_hashes = [i["blobSum"].split(":")[1] for i in doc["fsLayers"]]
+         except (KeyError, AttributeError, IndexError):
+            FATAL("manifest file %s missing expected v1 keys"
```
reidpr, line 316: Could be problems other than missing key; suggest making this error message more generic. Also check whether it still needs to be wrapped to two lines.

lib/charliecloud.py
```diff
@@ -304,10 +306,25 @@ def layer_hashes_load(self):
       except json.JSONDecodeError as x:
          FATAL("can't parse manifest file: %s:%d: %s"
                % (self.manifest_path, x.lineno, x.msg))
-      try:
-         self.layer_hashes = [i["digest"].split(":")[1] for i in doc["layers"]]
-      except (AttributeError, KeyError, IndexError):
-         FATAL("can't parse manifest file: %s" % self.manifest_path)
+
+      schema_version = doc['schemaVersion']
+      if (schema_version == 1):
```
reidpr, line 311: Do we know that `schema_version` is an int?

lib/charliecloud.py
```diff
@@ -332,6 +349,10 @@ def layers_read(self):
       if (self.layer_hashes is None):
          self.layer_hashes_load()
       layers = collections.OrderedDict()
+      # Schema version one (v1) allows one or more empty layers for Dockerfile
+      # entries like CMD (https://github.com/containers/skopeo/issues/393).
+      # Unpacking an empty layer doesn't accomplish much so we ignore them.
```
reidpr, line 354:
```suggestion
      # Unpacking an empty layer doesn't accomplish anything, so ignore them.
```

lib/charliecloud.py
```diff
@@ -341,7 +362,17 @@ def layers_read(self):
          except tarfile.TarError as x:
             FATAL("cannot open: %s: %s" % (path, x))
          members = collections.OrderedDict([(m, None) for m in members_list])
-         layers[lh] = TT(fp, members)
+         if ((lh in layers) and (len(members) > 0)):
```
reidpr, line 365:
```suggestion
         if (lh in layers and len(members) > 0):
```

lib/charliecloud.py
```diff
@@ -341,7 +362,17 @@ def layers_read(self):
...
+            FATAL("duplicate non-empty layer %s" % lh[:7])
+         if (len(members) > 0):
+            layers[lh] = TT(fp, members)
+         else:
+            empty_cnt += 1
+      if (self.schema_version == 'v1'):
```
reidpr, line 371: Above we compare schema version as an int?

lib/charliecloud.py
```diff
@@ -341,7 +362,17 @@ def layers_read(self):
...
+         DEBUG('reversing layer order for schema version one (1)')
```
reidpr, line 372:
```suggestion
         DEBUG('reversing layer order for schema version one (v1)')
```

lib/charliecloud.py
```diff
@@ -341,7 +362,17 @@ def layers_read(self):
...
+         layers = collections.OrderedDict(reversed(layers.items()))
+      if (empty_cnt > 0):
+         INFO("skipped %d empty layers" % empty_cnt)
```
reidpr, line 375: I wonder if this should be `DEBUG` and unconditional; empty layers seems like something end users might not care about and/or be confused by.

test/build/50_pull.bats
```diff
@@ -233,3 +233,22 @@ EOF
     [[ $output = 'regular file' ]]
     [[ $(cat "${img}/test/source") = 'regular' ]]
 }
+
+@test 'pull image with manifest schema v1' {
+    # Verify we handle images with manifest schema version one (1).
```
reidpr, line 238:
```suggestion
    # Verify we handle images with manifest schema version one (v1).
```

test/build/50_pull.bats
```diff
@@ -233,3 +233,22 @@ EOF
...
+    scope standard
+    if ( ! ch-grow --dependencies ); then
+        [[ $CH_BUILDER != ch-grow ]]
+        skip "ch-grow missing dependencies"
+    fi
+
+    unpack=$BATS_TMPDIR
+    cache=$unpack/dlcache
+    img=debian:squeeze
```
reidpr, line 247: Suggest add comment to explain why we selected this image.

reidpr (changes requested):
Good progress! Several nits but nothing of great concern.

lib/charliecloud.py
```diff
@@ -305,9 +307,28 @@ def layer_hashes_load(self):
...
+         except (KeyError, AttributeError, IndexError) as x:
+            FATAL("can't parse manifest file: %s:%d :%s"
+                  % self.manifest_path, x.lineno, x.msg)
+      elif (schema_version == '2'):
+         DEBUG('loading layer hashes from schema version 2 manifest')
+         try:
+            self.layer_hashes = [i["digest"].split(":")[1] for i in doc["layers"]]
+         except (KeyError, AttributeError, IndexError):
+            FATAL("manifest file %s missing expected v2 keys"
+                  % self.manifest_path)
```
reidpr, line 328:
```suggestion
            FATAL("can't parse manifest file: %s:%d :%s"
                  % self.manifest_path, x.lineno, x.msg)
```

reidpr (approved):
Looks great, one tweak I'll commit. Thanks for your hard work.

Issue #820, opened 2020-08-14
autoconf: separate build and source trees


j-ogas:
We currently don't support the distinction GNU build tools makes between source trees and build trees. This results in test suite and build failures when a user runs `./configure` outside the source directory.

reidpr:
IIRC this was somewhere between very difficult and impossible.

Issue #604, opened 2019-12-20
build: "make distcheck" passes


reidpr:
During PR #589, we didn't try `make distcheck`, and I'm almost certain it will fail. This issue is to fix it, and put it in Travis. I suspect it will be fast, so we could probably put it in the quick phase.

reidpr:
I tried `make distcheck` and it does fail.

PR #821, opened 2020-08-14
automake: override distcheck; add installcheck

j-ogas:
Addresses #604 and partially addresses #603. 

A couple notes: 1) unlike #753, this PR avoids `make check` due to Travis timeout issues; and 2) we override `distcheck` to avoid issue #820.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,3 +1,5 @@
+ch_libdir = $(DESTDIR)$(libdir)/charliecloud
+
 SUBDIRS = lib bin doc examples misc packaging test
 
 # The Travis stuff isn't really relevant for the tarballs, but they should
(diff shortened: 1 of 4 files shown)


reidpr:
Some poorly organized general impressions.

According to the [documentation](https://www.gnu.org/software/automake/manual/html_node/Checking-the-Distribution.html), `make distcheck` does the following:

reidpr:
Some un-organized impressions.

According to the docs, [`make installcheck`](https://www.gnu.org/software/automake/manual/html_node/Install-Tests.html) doesn't do anything unless you add an `installcheck-local` with whatever you want in it. So we need to think less about it.

On the other hand, [`make distcheck`](https://www.gnu.org/software/automake/manual/html_node/Checking-the-Distribution.html) makes a tarball and then does the following list of things (numbering added):

> 1. tries to do a VPATH build ... with the `srcdir` and all its content made read-only;
> 2. runs the test suite (with `make check`) on this fresh build;
> 3. installs the package in a temporary directory (with `make install`), and runs the test suite on the resulting installation (with `make installcheck`);
> 4. checks that the package can be correctly uninstalled (by `make uninstall`) and cleaned (by `make distclean`);
> 5. finally, makes another tarball to ensure the distribution is self-contained. 

Of these:

1. VPATH build is unsupported, and we decided we won't support it anytime soon. We don't test building with the input all read-only, though.
2. we do the equivalent with Travis
3. we do the equivalent with Travis
4. we do not do
5. we do not do

The three missing things (1 read-only, 4, 5) seem nice to have, but perhaps are not a high priority.

Really the only problem here is the VPATH build. There appears to be no documented way to disable this. I suspect that undocumented adventures in `distcheck-hook` or `sed`'ing `Makefile` could work disable it, but that makes me a tad queasy.

The overridden `make distcheck` in this PR does only 2 and 3. So it seems a little redundant.

Right now, `make dist` problems like omitting a file from `Makefile.am` tend to break `test/run/build-rpms.bats`, which is a hard place to debug them. I certainly make this mistake often and it might be nice to have it somewhere else.

The current errors from `make distcheck` and possibly `make installcheck` (I didn't look) are unhelpful.

I have no idea if we `make uninstall` or `make distclean` correctly.

It seems perhaps nice to test building with all the input read-only, to make sure we're not changing the source code during build.

Who is the audience for this `make distcheck`? I wonder if it would be better to implement the missing tests in `ch-test` and put friendly error messages in `make distcheck` and `make installcheck`.

j-ogas:
This PR is deferred. Closing for now, may reopen in the future.

Issue #583, opened 2019-11-13
ch-image build: implement build cache


rstyd:
ch-grow currently does not support caching built images.

reidpr:
Note, however, that base images aren't re-downloaded or re-built if already present.

PR #824, opened 2020-08-18
ch-grow cache

rstyd:
Addresses #583.


rstyd:
Ch-grow's layer caching is currently implemented using git, and git is required for layer caching support. The git repo is located $CH_GROW_STORAGE/img along with the base images. We have a helper class `Cache` that is responsible for all communication with `git`. We implemented this class so that we could easily switch git for another version control system at a later date. To help illustrate how the ch-grow layer caching works, here's a walkthrough using a couple of simple example dockerfiles. 

Dockerfile.base:    

    FROM alpine: latest
    RUN touch /foo
    RUN touch /bar

We'll build `Dockerfile.base` with the command `ch-build -t base -f Dockerfile.base`.

When we build an image, we first check if the base image exists in the cache using the image specified in the `FROM` instruction. In this case, it's `alpine: latest`. We append the dockerhub digest with the image name and compute a SHA256 hash which will refer to as `base_hash`. `base_hash = hash(digest + "alpine: latest")`. We look for an existing cache using `cache.exists()`. This function searches the git repo for a branch named `base_hash`. If there does not currently exist a branch named `base_hash`, this will return `False`, which it does in this case. We then invalidate the cache using `cache.invalidate()`. Invalidating the cache means that we assume from then on every further instruction is not in the cache. We then create a new image cache for this image with `cache.get_base()`. This function creates a new `git` branch with the name `base_hash` and adds/commits the base image to that branch. For each git add and commit, we use the `--work-tree` parameter. This parameter enables us to support caching tagged images. We discuss how this works further in a later section. 

Next, we have the `RUN` instruction. We compute the hash for subsequent instructions using the hash of the previous instruction appended to the current one. For this instruction, it is `inst_hash = hash(base_hash + "RUN touch /foo")`. Since the cache is invalid, we add the new layer to the cache using `cache.add_layer()`. `cache.add_layer()` performs another git add and commit. We continue building the image in this fashion, adding each new layer as a new commit. 
Now, we'll examine what happens when we build this dockerfile again with `ch-build -t base -f Dockerfile.base`.

First, the hash for the FROM command is the same as before `base_hash`. We check if there is an existing cache using `cache.exists()`. It returns true, so we load up the cache for this base image using `cache.get_base()`.We compute a hash for the second command the same as before to get `inst_cache`. Since the cache is valid, we check if the instruction is in the cache using `cache.in_cache()`. This function looks for a commit within the cache branch with `inst_cmd` as it's commit message. It will return true, so we use `cache.get_layer()`. The cache class keeps a pointer to the last valid step in the cache called `prev_hash`. `cache.get_layer()` sets this variable to that hash. This approach enables us to perform lazy caching. We only pull from the cache when every instruction is in the cache, or we hit an instruction that is not in the cache.

Finally, we'll look at what happens when build a dockerfile that adds an additional instruction.

Dockerfile.bar:   

    FROM alpine: latest
    RUN touch /foo
    RUN touch /bar
    RUN touch /baz

As before, we compute `base_hash` and find the cache with `cache.exists()`, and we check if the following instructions are in the cache with `cache.in_cache()`. When we hit the instruction `RUN touch /baz`, `cache.in_cache` fails, so we invalidate the cache with `cache.invalidate()` and pull down the last good layer with `cache.pull`. 

One difficult feature that`cache.pull()` accomplishes is that it works across tagged images. In ch-grow, we tag images by naming the directory of the image with that tag name. If we build `ch-grow -t alpine -f Dockerfile.base`, the resulting image will reside in a directory named `alpine`. The base image is an entirely different directory. This approach is problematic for pulling down a layer with git since git generally defines files by including their location on the filesystem. Originally, the ch-grow cache could only cache a single tagged image. We get around this by using `--work-tree`. `--work-tree` defines the working file system tree that git will use for a command. With`--work-tree`, we can force git to always pull down the last valid cached commit into the current tagged directory whatever that directory may be. We essentially continuously switch which directory is _the_ directory. With this one slight change, we completely solve the problem of tagged builds.

 We use `git reset` with the `--hard` option in `cache.pull()` to ensure we clean away any residual data from a previous build. We then build the image, as usual, adding the new `RUN touch /baz` layer with `cache.add_layer()`.

rstyd:
Closing. No longer relevant.

Issue #748, opened 2020-05-26
documentation: 404 when searching


j-ogas:
I seem to reach a 404 file not found error when using the documentation search feature.

rstyd:
Ah. Thanks! I mentioned this to Reid a while ago and forgot to add an issue.

tbugfinder:
A working search would be really beneficial for users.

reidpr:
See issue #822 for the specific trouble that @tbugfinder is having because (in part) they can't search.

PR #834, opened 2020-08-25
docs: fix searching 

heasterday:
Addresses #748 and cleans up the html/_images dir from being left behind.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -452,10 +452,12 @@ Sphinx
 We use Sphinx to build the documentation; the theme is
 `sphinx-rtd-theme <https://sphinx-rtd-theme.readthedocs.io/en/stable/>`_.
 
-Minimum versions are listed above. We currently use Sphinx 1.8 for building
+Minimum versions are listed above. We currently use Sphinx 3.2.1 for building
 what's on the web because there are bugs problematic for us in 2.x, e.g. `bad
 spacing in lists
-<https://github.com/readthedocs/sphinx_rtd_theme/issues/799>`_.
+<https://github.com/readthedocs/sphinx_rtd_theme/issues/799>`_ and versions
+prior to 2.x will `return markdown when searching
+<https://github.com/sphinx-doc/sphinx/issues/1618>`_.
 
 If you're on Debian Stretch or some version of Ubuntu, installing with
 :code:`pip3` will silently install into :code:`~/.local`, leaving the
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -452,12 +452,10 @@ Sphinx
 We use Sphinx to build the documentation; the theme is
 `sphinx-rtd-theme <https://sphinx-rtd-theme.readthedocs.io/en/stable/>`_.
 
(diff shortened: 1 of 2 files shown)


heasterday:
I would like to mention that the `_sources` directory contains a file for every `.rst` input so we could list them out rather than handling them like the contents of the `_static` directory.

heasterday:
Looks like searching is returning raw markdown now, seemingly a byproduct of building with sphinx 1.8.5. Build with a newer version of sphinx (3.2.1) appears to resolve the issue. Further investigation + a documentation update is required.

reidpr:
Re. the list spacing bug, here is what 1.8.5 produces.
Screen Shot 2020-08-26 at 1 40 05 PM

reidpr:
With Sphinx 3.2.1 (the latest available on PyPI), we do get the same. But, I had to `pip3 uninstall` everything Sphinx-related first. Upgrading just Sphinx gave extra space above the 2nd level list (open circle bullets).

reidpr:
Search is *better* but it's not *good*. For example, searching for "ch-grow" yields "Your search did not match any documents".

reidpr:
@tbugfinder, we uploaded docs generated by this branch to the [documentation site](https://hpc.github.io/charliecloud). Do you have time to poke at the search a little and see if it's good enough for your needs? (E.g., would gotchas like "ch-grow" not finding anything cause significant confusion for you?)

reidpr:
I'm inclined to merge this, modulo any tidying, but I don't think it solves the search problem completely and we should continue to try to improve things.

tbugfinder:
@reidpr I've executed few searches. Indeed few characters are not included it's certainly helpful.
LGTM😀

doc/install.rst
```diff
@@ -452,10 +452,12 @@ Sphinx
...
 `sphinx-rtd-theme <https://sphinx-rtd-theme.readthedocs.io/en/stable/>`_.
 
-Minimum versions are listed above. We currently use Sphinx 1.8 for building
+Minimum versions are listed above. We currently use Sphinx 3.2.1 for building
 what's on the web because there are bugs problematic for us in 2.x, e.g. `bad
 spacing in lists
-<https://github.com/readthedocs/sphinx_rtd_theme/issues/799>`_.
+<https://github.com/readthedocs/sphinx_rtd_theme/issues/799>`_ and versions
+prior to 2.x will `return markdown when searching
+<https://github.com/sphinx-doc/sphinx/issues/1618>`_.
```
reidpr, line 460: I wonder if something like:

> Minimum versions are listed above. Note that while anything greater than the minimum should yield readable documentation, we don't test quality with anything other than what we use to build the website, which is usually but not always the most recent version available on PyPI.

would be clearer and more future-proof. E.g., saying we use 3.x because 2.x is buggy seems a bit odd.

heasterday: I really like this wording, I will add it.

heasterday:
@reidpr Added your comment. Let me know if there is anything else you would like on this one.

Issue #840, opened 2020-09-01
Vagrantfile: tests fail because of missing getty process


rstyd:
When testing the Vagrantfile with `ch-test` this error occurs. 

```
default: not ok 80 signal process outside container
default: # (from function `pedantic_fail' in file run/../common.bash, line 102,
default: #  in test file run/ch-run_uidgid.bats, line 141)
default: #   `[[ $(pgrep -c getty) -eq 0 ]] && pedantic_fail 'no getty process found'' failed
default: # no ID arguments
default: #
```

PR #845, opened 2020-09-11
Vagrantfile: enable gettys

rstyd:
Addresses #840

### DIFF
--- a/packaging/vagrant/Vagrantfile
+++ b/packaging/vagrant/Vagrantfile
@@ -294,8 +294,8 @@ EOF2
     cp /lib/systemd/system/getty@.service getty@tty1.service
     sed -ri 's|^ExecStart=.*$|ExecStart=-/sbin/agetty --autologin charlie --noclear %I|' getty@tty1.service
 
-    # We need to enable another TTY for the tests to pass
-    # One test attempts to signal this getty process from within the container
+    # A test tries to signal a getty process from within the container, so
+    # we need at least one enabled.
     sudo systemctl enable getty@tty2.service
 
     # Configure SSH to allow password logins. We would prefer to keep the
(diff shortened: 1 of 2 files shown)


reidpr (changes requested):
Great that it was so short! A few suggestions:

1. Do we need five extra? I recall two, but I think actually it's just one because it was whatever was on TTY1 (which is now the autologin stuff) plus one for the test suite.
2. Document briefly why we are enabling the getty(s) here in the Vagrantfile.
3. I was pretty sure this used to be explained in the documentation somewhere? But, it needs to be documented. Even though it's a pedantic fail, we don't want to go on this chase again. Perhaps a comment at the point the test fails?

rstyd:
1. We do not. We could get away with just TTY2. I was thinking we could just enable them all since that seems to be the normal configuration Linux installations have. 
2. Will do. 
3. I'll add something to the test output.

packaging/vagrant/Vagrantfile
```diff
@@ -294,6 +294,10 @@ EOF2
     cp /lib/systemd/system/getty@.service getty@tty1.service
     sed -ri 's|^ExecStart=.*$|ExecStart=-/sbin/agetty --autologin charlie --noclear %I|' getty@tty1.service
 
+    # We need to enable another TTY for the tests to pass
+    # One test attempts to signal this getty process from within the container
+    sudo systemctl enable getty@tty2.service
```
reidpr, line 299: This section doesn't need sudo. Could also fix the other similar errors.

reidpr (approved):
Looks good pending your approval of my tidy. Thanks!

rstyd:
Looks good to me!

Issue #854, opened 2020-09-22
build fails due to `python --version`


trandles-lanl:
Building Charliecloud from source breaks on Summit at ORNL if you have the python 3.7.0 module loaded. The build works fine if NO python module is loaded even though the `./configure` output seems to indicate that `ch-grow` won't work.

Here I have done `git checkout v0.19` in my repo clone. I have loaded python/3.7.0.
```
[trandles@login2.summit charliecloud]$ git status
# HEAD detached at v0.19
nothing to commit, working directory clean
[trandles@login2.summit charliecloud]$ python --version
Python 3.7.0 (default, Oct 12 2018, 15:17:16)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-28)]
```

The following (chopped-up for brevity) output from `configure` shows that it believes the version of Python is too old. Note the "extra" 4.8.5 on a newline in the `configure` output and see the output from `python --version` above.
```
[trandles@login2.summit charliecloud]$ ./configure --disable-man --disable-tests --disable-html
<CUT>
    ch-grow(1) ... yes

Building images via our wrappers
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  with Buildah: no
    Buildah ≥ 1.11.2 ... not found

  with ch-grow(1): no
    enabled ... yes
    Python shebang line ... /usr/bin/env python3
    Python in shebang ≥ 3.4 ... too old (3.7.0
4.8.5)
    "lark-parser" module ≥ 0.7.1 ...
    "requests" module ≥ 2.6.0 ...
    ch-run(1) ... yes
<CUT>
```

This leads to a broken Makefile.
```
[trandles@login2.summit charliecloud]$ make
Makefile:262: *** missing separator.  Stop.
```
Inspecting the Makefile around line 262 I find:
```
PYTHON_SHEBANG = /usr/bin/env python3
PYTHON_VERSION_NOTE = too old (3.7.0
4.8.5)
REQUESTS_VERSION_NOTE =
```
The "extra" version info from above has made its way into the Makefile.

I nuked my local Charliecloud repo, unloaded the python/3.7.0 module, and started over. This time I'm building with a clean clone of the repo, `git checkout v0.19`, and no python3 loaded. `configure` tells me that no python3 is found and it doesn't look like I should be able to build images with `ch-grow`.
```
[trandles@login2.summit charliecloud]$ ./configure --disable-man --disable-tests --disable-html
<CUT>
Building images via our wrappers
[...]

PR #856, opened 2020-09-23
configure: use only first line of Python version

reidpr:
Addresses #854.


trandles-lanl (approved):
This looks good. `configure` produced a clean `Makefile`, even with the previously problematic python/3.7.0 module loaded.

Good work, approved. :)

PR #865, opened 2020-10-05
ch-grow pull: add multi-platform support

j-ogas:
Addresses #1017. [added by @reidpr]

### DIFF
--- a/bin/ch-builder2tar
+++ b/bin/ch-builder2tar
@@ -35,7 +35,8 @@ parse_basic_args "$@"
 # Kludgey argument parsing.
 while true; do
     case $1 in
-        --arch=*)
+        --arch)
+            shift
             arch=$1
             arch=$(printf '%s' "$arch" | sed 's/--arch=//g')
             arch=$(printf '%s' "$arch" | sed 's/\//%/g')
--- a/bin/ch-builder2tar
+++ b/bin/ch-builder2tar
@@ -60,8 +60,6 @@ if [ "$#" -ne 2 ]; then
     usage
 fi
 
-
-
 image=$1
 outdir=$2
 tar=${outdir}/$(tag_to_path "$image").tar
--- a/bin/ch-builder2tar
+++ b/bin/ch-builder2tar
@@ -37,7 +37,7 @@ while true; do
     case $1 in
         --arch=*)
             arch=$1
-            arch=$(printf '%s' $arch | sed 's/--arch=//g')
+            arch=$(printf '%s' "$arch" | sed 's/--arch=//g')
             arch=$(printf '%s' "$arch" | sed 's/\//%/g')
             shift
             ;;
@@ -142,7 +142,7 @@ ch-grow)
 
     echo "exporting"
     storage=$("${ch_bin}/ch-grow" storage-path)
-    if [ -n $arch ]; then
+    if [ -n "$arch" ]; then
         storage="${storage%/*}/${arch}"
     fi
     ( cd "${storage}/img/${image}" && tar cf - . ) | pv_ > "$tar"
--- a/bin/ch-builder2tar
+++ b/bin/ch-builder2tar
@@ -35,6 +35,12 @@ parse_basic_args "$@"
 # Kludgey argument parsing.
 while true; do
     case $1 in
+        --arch=*)
+            arch=$1
+            arch=$(printf '%s' $arch | sed 's/--arch=//g')
+            arch=$(printf '%s' "$arch" | sed 's/\//%/g')
+            shift
+            ;;
         -b|--builder)
             shift
             CH_BUILDER=$1
@@ -53,6 +59,8 @@ done
 if [ "$#" -ne 2 ]; then
     usage
 fi
+
+
 
 image=$1
 outdir=$2
@@ -134,6 +142,9 @@ ch-grow)
 
     echo "exporting"
     storage=$("${ch_bin}/ch-grow" storage-path)
+    if [ -n $arch ]; then
+        storage="${storage%/*}/${arch}"
+    fi
     ( cd "${storage}/img/${image}" && tar cf - . ) | pv_ > "$tar"
 
     ;;
--- a/bin/ch-test
+++ b/bin/ch-test
@@ -113,9 +113,11 @@ builder_set () {
         builder_choose
         method='default'
     fi
+    architecture_choose
     printf "%-*s %s (%s)\n" "$width" 'builder:' "$CH_BUILDER" "$method"
     if [[ $CH_BUILDER == ch-grow ]]; then
-        vset CH_GROW_STORAGE '' "$CH_GROW_STORAGE" "/var/tmp/$USER/ch-grow" \
+        vset CH_GROW_STORAGE '' "$CH_GROW_STORAGE" \
+                             "/var/tmp/$USER/ch-grow/$CH_GROW_ARCH" \
                              "$width" 'ch-grow storage'
     fi
 }
--- a/doc/ch-grow_desc.rst
+++ b/doc/ch-grow_desc.rst
@@ -150,14 +150,17 @@ Destination argument:
 
 Options:
   :code:`--arch`
-    Check the target image :code:`IMAGE_REF` fat manifest for a reference that
-    targets the specified architecture. If found pull it; otherwise error.
+    Check the image :code:`IMAGE_REF` fat manifest for the architecture
+    :code:`--arch=ARCH[/VARIANT]` image manifest. If found pull it; otherwise
+    error. See :code:`--inspect-arch IMAGE_REF` for list of valid platform
+    architectures.
+
+  :code:`--inspect-arch`
+    Print a list of targitable platforms to use with :code:`--arch`.
 
   :code:`--inspect-manifest`
-    Print the image :code:`IMAGE_REF` index (fat manifest). This is typically
-    used to list the image's supported platform digests, e.g., :code:`arm64`,
-    :code:`ppc64le`, etc. Note: some images do not have fat manifests; print
-    the default manifest if a fat manifest is unavailable.
+    Print the image :code:`IMAGE_REF` index, i.e., fat manifest if available;
+    otherwise print the standard default manifest.
 
   :code:`--parse-only`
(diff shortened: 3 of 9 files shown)


doc/ch-grow_desc.rst
```diff
@@ -402,7 +414,7 @@ Download the Debian Buster image and place it in the storage directory::
 
   $ ch-grow pull debian:buster
   pulling image:   debian:buster
-
+  fat manifest: downloading
```
reidpr, line 417: This should IMO say which architecture was selected.

lib/base.sh
```diff
@@ -10,6 +10,37 @@ lib="${ch_bin}/../lib/charliecloud"
...
+            ;;
+        x86_64)
+            arch='amd64'
+            ;;
+        *)
+            arch="$host_arch"
+            ;;
+    esac
+    export CH_GROW_ARCH=$arch
+}
```
reidpr, line 42: 1. This is a lot of lines to implement a table. Is there anything shorter that we can do?
2. I wonder if we should use dot instead of slash or percent to separate architecture and variant internally?
3. We should separate the concerns of (1) mapping arch specifications to the canonical one and (2) setting what arch to use.

lib/charliecloud.py
```diff
@@ -55,6 +56,15 @@
...
 ## Globals ##
 
+# Map common uname machine type to manifest platform architecture.
+ARCH_MAP = {'aarch32': {'architecture': 'arm',   'variant': 'v7'},
+            'aarch64': {'architecture': 'arm64', 'variant': 'v8'},
+            'armv5l':  {'architecture': 'arm',   'variant': 'v5'},
+            'armv6l':  {'architecture': 'arm',   'variant': 'v6'},
+            'armv7l':  {'architecture': 'arm',   'variant': 'v7'},
+            'armv8l':  {'architecture': 'arm64', 'variant': 'v8'},
+            'x86_64':  {'architecture': 'amd64'}}
```
reidpr, line 66: What about tuples instead of subdictionaries?

Also, if `variant` must be present, then there will be fewer code paths.

Use double quotes.

lib/charliecloud.py
```diff
@@ -210,12 +220,36 @@ def __init__(self, ref, download_cache, unpack_dir, image_subdir=None):
    def __str__(self):
       return str(self.ref)
 
+   @property
+   def default_arch_dlcache(self):
+      return os.path.dirname(os.path.dirname(
+                             self.download_cache)) + '/amd64/dlcache'
```
reidpr, line 226: Should we change all these path properties (including the existing ones) to start with `path_`, in keeping with the most-significant-first naming convention?

lib/charliecloud.py
```diff
@@ -210,12 +220,36 @@ def __init__(self, ref, download_cache, unpack_dir, image_subdir=None):
...
+
+   @property
+   def default_arch_unpack(self):
+      return os.path.dirname(os.path.dirname(
+                             self.download_cache)) + '/amd64/img'
```
reidpr, line 231: Is there a shorter way to do this than two nested `dirname`?

lib/charliecloud.py
```diff
@@ -210,12 +220,36 @@ def __init__(self, ref, download_cache, unpack_dir, image_subdir=None):
...
+
+   @property
+   def inferred_fat_dir(self):
+      return os.path.dirname(os.path.dirname(self.download_cache)) + '/fatman'
```
reidpr, line 235: Why do we need separate directories in the download cache for different architectures?

lib/charliecloud.py
```diff
@@ -210,12 +220,36 @@ def __init__(self, ref, download_cache, unpack_dir, image_subdir=None):
...
+
+   @property
+   def inferred_arch(self):
+      "Preferred image architecture and variant key"
+      return "%s" % os.path.basename(os.path.dirname(self.download_cache))
```
reidpr, line 240: This string operation looks like a no-op?

lib/charliecloud.py
```diff
@@ -258,6 +292,17 @@ def download(self, use_cache):
             dl.get_layer(lh, path)
       dl.close()
 
+   def download_fat_manifest(self, dl=None, use_cache=True):
+      """Download image manifest list and put it in the download cache."""
+      if (dl is None):
+         dl = Repo_Downloader(self.ref)
```
reidpr, line 298: Why can `dl` be `None`?

lib/charliecloud.py
```diff
@@ -258,6 +292,17 @@ def download(self, use_cache):
             dl.get_layer(lh, path)
       dl.close()
 
+   def download_fat_manifest(self, dl=None, use_cache=True):
```
reidpr, line 295: Seems awkward that the fat manifest gets its own download method but the other components don't.

lib/charliecloud.py
```diff
@@ -602,10 +752,10 @@ def path_full(self):
 
    @property
    def version(self):
-      if (self.tag is not None):
-         return self.tag
       if (self.digest is not None):
          return "sha256:" + self.digest
+      if (self.tag is not None):
+         return self.tag
```
reidpr, line 758: Why did this move?

lib/charliecloud.py
```diff
@@ -854,6 +1016,36 @@ def INFO(*args, **kwargs):
 def WARNING(*args, **kwargs):
    log(color="31m", prefix="warning: ", *args, **kwargs)
 
+def arch_arg_fixup(arch):
+    if arch[0] == '/': # leading '/' is optional; remove it
+       arch = arch[1:]
+    return arch.replace('/', '%')
+
+def arch_arg_validate(arch):
+    rx = re.compile('^(/?\w+(/\w+)?)$')
```
reidpr, line 1025: Compiling is rarely necessary; it saves time only in rather unusual cases.

reidpr (commented):
Great progress! Various comments in-line. I hope this helps.

lib/charliecloud.py
j-ogas: To prioritize digest over tag. For example:

```
$ ch-grow pull --arch=arm64/v8 centos:8
```

In the above our image ref will target `/v2/library/centos:8`. To get the `arm64` variant we need to  prioritize the hash given to us from the fat manifest.

lib/base.sh
j-ogas: 1. If we wish to remain explicit we could condense it to:
```bash
    host_arch=$(uname -m)
    case $host_arch in
        aarch32) arch='arm/v7'     ;;
        aarch64) arch='arm64/v8'   ;;
         armv5l) arch='arm/v5'     ;;
         armv6l) arch='arm/v6'     ;;
         armv7l) arch='arm/v7'     ;;
         armv8l) arch='arm64/v8'   ;;
         x86_64) arch='amd64'      ;;
              *) arch="$host_arch" ;;
    esac
```
2. OK

3. we could always store at `$CH_GROW_STORAGE/$(uname -m) and let ch-image worry about converting and inferring canonical name.

doc/ch-grow_desc.rst
j-ogas: Before we select the architecture we need to fat manifest. Here we are telling the user it is being downloaded. We could say something like "image platform index: downloading" or something of the like to differentiate it from the term 'manifest'. But fat manifest is technically correct.

bin/ch-test
```diff
@@ -113,9 +113,11 @@ builder_set () {
         builder_choose
         method='default'
     fi
+    architecture_choose
     printf "%-*s %s (%s)\n" "$width" 'builder:' "$CH_BUILDER" "$method"
     if [[ $CH_BUILDER == ch-grow ]]; then
-        vset CH_GROW_STORAGE '' "$CH_GROW_STORAGE" "/var/tmp/$USER/ch-grow" \
+        vset CH_GROW_STORAGE '' "$CH_GROW_STORAGE" \
+                             "/var/tmp/$USER/ch-grow/$CH_GROW_ARCH" \
```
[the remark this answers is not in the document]
j-ogas: Here `CH_GROW_STORAGE` is being set to contain the architecture directories. Line 120 is setting the default.

lib/charliecloud.py
j-ogas: I can fiddle with tuples instead of dictionaries here.

Variant is optional.

bin/ch-builder2tar
```diff
@@ -134,6 +142,9 @@ ch-grow)
 
     echo "exporting"
     storage=$("${ch_bin}/ch-grow" storage-path)
+    if [ -n $arch ]; then
+        storage="${storage%/*}/${arch}"
+    fi
```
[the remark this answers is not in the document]
j-ogas: We do always specify an arch (default is `$(uname -m)`). In the code block above, if `--arch` isn't specified, we use what `$ ch-grow storage path` returns, which relies on `$CH_GROW_STORAGE` which includes the architecture. So on `x86_64` the storage path will be `/var/tmp/$USER/ch-grow/amd64`.

The argument, `--arch`, in the `ch-build*` context as above, is used to override an architecture. For example, if I am on `amd64`, and I want to pull an arm image and then pack it into a tarball, we would need to specify the different architecture via `ch-builder2tar --arch=arm64/v8 /tmp` to target the `arm64/v8` in storage.

Under the hood this is what is happening:
```
+ storage=/var/tmp/jogas/ch-grow/amd64
+ '[' -n arm64%v8 ']'
+ storage=/var/tmp/jogas/ch-grow/arm64%v8
+ cd /var/tmp/jogas/ch-grow/arm64%v8/img/centos:8
```

lib/charliecloud.py
j-ogas: Two different functions use it: `print_arch_list` and `set_digest_for_arch`. In both functions there is a use case in which the fat manifest for a specified image hasn't been downloaded yet.

lib/charliecloud.py
j-ogas: To avoid clashes with manifest files that have the same name but different content (different architectures). Alternatively we can just change our manifest storage convention to specify architecture, e.g., `centos:8.arm64%v8.manifest.json`.

lib/charliecloud.py
reidpr: Aha, let's a comment to that effect then.

PR #867, opened 2020-10-08
ch-image push

j-ogas:
Addresses #740.


reidpr:
@j-ogas, very fine work. I'm done with my refactoring; back at you for further improvements & discussion before we tag it ready for review. Hopefully CI passes.

Tasks remaining, and follow-up tasks:

* Refactor all path handling to use `charliecloud.Path` objects instead of strings.
* Progress bars throughout `ch-image`. `tqdm` looks fantastic, but if we support that, I feel like we need a fallback in case that module is not available, so then we need to test both cases, right?
* Add separate local and remote image references to `ch-image pull`.
* Validate that image path is an image using Megan's new function when merged (PR #908).

What have I missed?

lib/charliecloud.py
```diff
@@ -966,12 +1136,63 @@ def dependencies_check():
    if (len(depfails) > 0):
       sys.exit(1)
 
+def done_notify():
+   if (os.environ.get("USER", None) == "jogas"):
+      INFO("!!! KOBE !!!")
```
j-ogas, line 1141: +1

lib/push.py
```diff
@@ -0,0 +1,146 @@
...
+      config = self.config_new()
+      manifest = self.manifest_new()
+      # Prepare layers.
+      for (i, tar_uc) in enumerate(tars_uc, start=1):
+         ch.INFO("layer %d/%d: preparing" % (i, len(tars_uc)))
+         path_uc = ch.storage.upload_cache // tar_uc
+         hash_uc = ch.file_hash(path_uc)
+         config["rootfs"]["diff_ids"].append("sha256:" + hash_uc)
+         #size_uc = ch.file_size(path_uc)
+         path_c = ch.file_gzip(path_uc, ["-9", "--no-name"])
```
j-ogas, line 110: Is creation time relevant here as well? One issue I had with compression was the metadata bits; compressing the same uncompressed tarball resulted in different hashes. Is `--no-name` sufficient?

lib/charliecloud.py
```diff
@@ -273,137 +229,108 @@ def fixup(self):
...
+   def tarballs_write(self, tarball_dir):
+      """Write one uncompressed tarball per layer to tarball_dir. Return a
+         sequence of tarball basenames, with the lowest layer first."""
+      # FIXME: Yes, there is only one layer for now and we'll need to update
+      # it when (if) we have multiple layers. But, I wanted the interface to
+      # support multiple layers.
+      base = "%s.tar" % self.ref.for_path
+      path = tarball_dir // base
+      try:
+         INFO("layer 1/1: gathering")
```
j-ogas, line 304: In other places I see you use a for loop with an `enumerate`. Should we do that here to provide us with the `1/1` instead of this hardcoded print?

test/build/50_localregistry.bats
```diff
@@ -0,0 +1,90 @@
...
+    echo "$output"
+    [[ $status -eq 0 ]]
+    [[ $output = *'pushing image:   localhost:5000/00_tiny'* ]]
+    [[ $output = *"image path:      ${CH_IMAGE_STORAGE}/img/localhost:5000%00_tiny"* ]]
+
+    rm "$CH_IMAGE_STORAGE"/img/localhost:5000%00_tiny
+}
+
+@test "${tag}: with destination reference" {
+    run ch-image -v --tls-no-verify push 00_tiny localhost:5000/00_tiny
```
j-ogas, line 42: Should we also be testing destination with a paths? What about non-latest tags? For example:

```
00_tiny localhost:5000/tiny_images/00_tiny:tiniest
```

j-ogas (commented):
Like what you've done with the place.

doc/ch-image_desc.rst
```diff
@@ -242,6 +238,37 @@ Options:
     Parse :code:`IMAGE_REF`, print a parse report, and exit successfully
     without talking to the internet or touching the storage directory.
 
+:code:`push`
+------------
+
+Push the image described by the image reference :code:`IMAGE_REF` from the
```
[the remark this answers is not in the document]
reidpr: The reasoning for `IMAGE_REF` over `SOURCE_REF` was that sometimes it's the source reference *and* the destination reference (if `DEST_REF` is omitted), and sometimes it's only the source (if `DEST_REF` is given), and sometimes it's only the destination (if `--image` is given). So `SOURCE_REF` seemed a misnomer.

I see on line 9 it's still `SOURCE_REF` which is an oversight; I meant to change it to `IMAGE_REF`. What's your opinion on the terminology here?

examples/spark/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# ch-test-scope: standard
+# ch-test-scope: full
 # Use Buster because Stretch JRE install fails with:
```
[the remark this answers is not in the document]
reidpr: I made the changes in both places; Git should notice that and be happy.

lib/charliecloud.py
reidpr: I think things will probably change enough that any loop would need to be rewritten anyway when it's time.

lib/charliecloud.py
```diff
@@ -773,63 +787,131 @@ def authorize(self, res):
       else:
          FATAL("unknown auth type: %s" % auth_h)
 
+   def blob_exists_p(self, digest):
```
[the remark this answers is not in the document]
reidpr: It's [from LISP](https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/node69.html) and lets us put the most-significant parts of the name first (here, "blob" and "exists").

lib/charliecloud.py
```diff
@@ -773,63 +787,131 @@ def authorize(self, res):
...
+         a string to prepend to the log messages; default empty string."""
+      INFO("%s%s: checking if already in repository" % (note, digest[:7]))
+      # 1. Check if blob already exists. If so, stop.
+      if (self.blob_exists_p(digest)):
+         INFO("%s%s: already present" % (note, digest[:7]))
+         return
+      INFO("%s%s: not present, uploading" % (note, digest[:7]))
+      # 2. Get upload URL for blob.
+      url = self._url_of("blobs", "uploads/")
+      res = self.request("POST", url, {202})
```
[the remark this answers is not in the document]
reidpr: I was unsure of the significance of the slash. If its presence avoids a redirect, we should have it present, to save a round-trip. 

`requests.head()` seems to be special w.r.t. redirects. Specifically, both `requests.request("HEAD", ...)` nor `requests.post()` follow redirects by default.

lib/charliecloud.py
```diff
@@ -773,63 +787,131 @@ def authorize(self, res):
...
+         username = input("\nUsername: ")
+         password = getpass.getpass("Password: ")
       return (username, password)
 
-   def get(self, url, path, headers=dict(), statuses=[200]):
-      """GET url, passing headers, and write the body of the response to path.
-         Use current session if there is one, or start a new one if not.
-         Authenticate if needed."""
-      DEBUG("GETting: %s" % url)
+   def layer_from_file(self, digest, path, note=""):
```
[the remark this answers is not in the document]
reidpr: Well, this one is in line with `layer_to_file`. The intent was to be clear about the source of the data. FWIW, `manifest_upload` takes a sequence of bytes now.

lib/charliecloud.py
```diff
@@ -773,63 +787,131 @@ def authorize(self, res):
...
+      "Upload gzipped tarball layer at path, which must have hash digest."
+      # NOTE: We don't verify the digest b/c that means reading the whole file.
+      DEBUG("layer tarball: %s" % path)
+      fp = open_(path, "rb")  # open file avoids reading it all into memory
+      self.blob_upload(digest, fp, note)
+      ossafe(fp.close, "can't close: %s" % path)
+
+   def layer_to_file(self, digest, path):
+      "GET the layer with hash digest and save it at path."
+      # /v1/library/hello-world/blobs/<layer-hash>
```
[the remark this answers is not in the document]
reidpr: `v2` I'd think. I wonder if the comment is from early experiments that used v1.

lib/push.py
reidpr: Seems to do the trick:

```
$ echo foo > foo.txt
$ echo foo > bar.txt
$ ls -l *.txt
-rw-r----- 1 reidpr reidpr 4 Dec 17 16:58 bar.txt
-rw-r----- 1 reidpr reidpr 4 Dec 17 16:57 foo.txt
$ gzip -9 --no-name *.txt
$ sha256sum *.txt.gz
9a904ea72399535277c74ccf752b4c815b20100e16f9032078b73db7b86bda79  bar.txt.gz
9a904ea72399535277c74ccf752b4c815b20100e16f9032078b73db7b86bda79  foo.txt.gz
```

test/build/50_localregistry.bats
reidpr: Yeah, probably.

reidpr:
yolo

PR #880, opened 2020-10-21
Move charliecloud man page to section 7

wiene:
At present the charliecloud man page is installed in [section 1](https://manpages.debian.org/unstable/manpages/man-pages.7.en.html#Sections_of_the_manual_pages). This PR moves it to section 7 which is more suitable in my opinion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,7 @@ a.out
 /doc/_deps.rst
 /doc/doctrees/
 /doc/html/
-/doc/man/*.1
+/doc/man/*.[17]
 /lib/charliecloud
 /lib/contributors.bash
 /lib/version.py
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -57,7 +57,7 @@ tutorial.rst
 
 if ENABLE_MAN
 man1_MANS = \
-man/charliecloud.1 \
+man/charliecloud.7 \
 man/ch-build.1 \
 man/ch-build2dir.1 \
 man/ch-builder2squash.1 \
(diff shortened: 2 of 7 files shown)


reidpr (commented):
I agree, that seems like a better place.

Looks like the tests fail because `docs-sane.py.in:92` expects everything to be §1:

```
# conf.py: bad section: charliecloud: 7 != 1
```

This would not be hard to update to check that `charliecloud` is in §7 and everything else in §1. Would you like to do that, or should we?

wiene:
I tried to fix the issues in `docs-sane.py` with my second commit (thanks for your pointer to this file) but somehow the CI tests now take surprisingly long. If the tests still fail (finally), feel free to apply any additional patch to make it work.

reidpr:
Thank you! Changes look good. I think the Travis slowdown is unrelated.

reidpr (approved):

wiene:
It seems to me that the last test failure was caused by a glitch of the CI system:
```
[...]
ERROR: Failed to create usr/libexec/gcc/x86_64-alpine-linux-musl/8.3.0/cc1: I/O error
ERROR: gcc-8.3.0-r0: IO ERROR
[...]
make: gcc: Command not found
make: *** [<builtin>: chroot-escape] Error 127
[...]
```

reidpr:
I think you're right. Travis has been even glitchier than usual past few days. I restarted the job.

wiene:
From my point of view this PR looks fine now. I hope that I finally caught all places where modifications were needed.

reidpr:
Thanks; that's quite a few loose ends to tie up. Much appreciated!

Issue #879, opened 2020-10-20
examples/spark Java env variables on host can cause failures


heasterday:
On some of our systems JAVA_ROOT, JAVA_HOME, and JAVA_BINDIR are set on login, this confuses the spark tests as to where Java is installed causing failures. Explore `--unset-env=JAVA*` in the tests to shield them from this.

PR #881, opened 2020-10-21
examples/spark unset Java env vars prior to running tests

heasterday:
Addresses #879 

I saw three ways this could be resolved:
1. Put the `unset` calls in the setup function
2. Use `--unset-env=JAVA*` on each `ch-run`
3. Add a function to unset these vars

I went with option 3 as 1 would run the `unset` commands for every test even if they don't need it and 2 made the `ch-run` commands quite verbose. This is similar to what we do in the [ch-run_join.bats](https://github.com/hpc/charliecloud/blob/5b2b3cf2518ee39c45bee7f36e8ee22a853d0a31/test/run/ch-run_join.bats#L73) tests.

Tested locally and on the system where the issue was identified.

### DIFF
--- a/examples/spark/test.bats
+++ b/examples/spark/test.bats
@@ -46,6 +46,13 @@ setup () {
     master_log="${spark_log}/*master.Master*.out"
 }
 
+unset_vars() {
+    # Unset these Java variables so the container doesn't use host paths.
+    unset JAVA_BINDIR
+    unset JAVA_HOME
+    unset JAVA_ROOT
+}
+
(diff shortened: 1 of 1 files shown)


reidpr (changes requested):
Option 1 seems better to me, because as written currently, some of the tests have a significantly different environment than others, which seems like a trap for the future, and unsetting three variables is cheap. You can even put them all on the same `unset` line if you'd like to save line count.

heasterday:
Made changes per feedback and retested.

PR #891, opened 2020-10-30
ch-grow build: new option --bind to bind-mount directories during build

reidpr:
This PR adds `ch-grow build --bind` to bind-mount directories at build time.

There's some extra mess so I could get the tests passing on my new platform.

### DIFF
--- a/bin/ch-grow.py.in
+++ b/bin/ch-grow.py.in
@@ -105,8 +105,11 @@ def main():
                        formatter_class=ch.HelpFormatter)
    sp.set_defaults(func=build.main)
    add_opts(sp, common_opts, True)
-   sp.add_argument("--build-arg", action="append", default=None,
-                   metavar="ARG[=VAL]",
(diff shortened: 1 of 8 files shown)


trandles-lanl (approved):
This looks like it works for `RUN` in the Dockerfile but not for `COPY`. I could get it to work by doing `RUN cp /mnt/0/file /usr/local/lib/file`. Docker's `COPY` is convoluted enough so I don't know if this particular oddity is worth fussing over. I would make it _very clear_ in the Charliecloud documentation that the bind mount is only present during `RUN` commands in the Dockerfile and provide an example snippet of its correct use.

I approve the PR with the above caveat.

samuelkgutierrez (approved):
This capability appears to work for my use case. A+ Would Buy Again! Thank you.

heasterday (approved):

trandles-lanl (approved):
I like.

Issue #894, opened 2020-11-05
ch-grow build: don't fakeroot by default


reidpr:
After discussion, fakeroot by default is too magic. We will:

1. Change to no fakeroot by default; i.e., `--no-fakeroot` goes away, to be replaced with `--fakeroot` (but see below for naming).
2. But, always run the detection code (do we have config for the distro in question, and are there any commands that would seem to match).
3. Only run the fakeroot install code if fakeroot doesn't seem to be installed already.
4. If a build fails and the detection code thought fakeroot would help, suggest the user try it.
5. (possibly) Rename `--fakeroot` to avoid scaring people.

PR #911, opened 2020-11-24
refactor fakeroot

reidpr:
Addresses #894.

Code review should run full test suite IMO.


doc/ch-grow_desc.rst
```diff
@@ -209,16 +180,22 @@ This one is (ironically) :code:`apt-get` failing to drop privileges:
...
+     contains, which determines the specific workarounds.
 
-  1. After :code:`FROM`, install a :code:`fakeroot(1)` implementation. This
-     sometimes also needs extra steps like turning off the :code:`apt` sandbox
-     (for Debian Buster) or enabling EPEL (for CentOS/RHEL).
+  2. Before the user command in the first :code:`RUN` instruction where the
+     injection seems needed, install :code:`fakeroot(1)` in the image, if one
+     is not already installed, as well as any other necessary initialization
+     commands. For example, we turn off the :code:`apt` sandbox (for Debian
+     Buster) and enable EPEL (for CentOS/RHEL).
```
trandles-lanl, line 196: Enabling EPEL can have undesirable results, so be careful. If you _only_ need EPEL to install `fakeroot` then I'd suggest:

1. install EPEL repo RPM
1. disable EPEL repository using `yum-config-manager --disable epel`
1. EXPLICITY enable EPEL like `yum --enablerepo=epel`

It's common that there are upstream versions of RPMs in EPEL that will break other non-EPEL distro software. Two weeks ago I had a broken CentOS OpenStack installation. The problem was I had EPEL enabled and it installed some python RPMs that caused the CentOS OpenStack software to fail.

trandles-lanl: You might want to check to see if EPEL is already installed and enabled. That would imply the user knows the EPEL risks and the above is less of a concern.

`yum repolist enabled` will give you the output you want to parse

lib/fakeroot.py
```diff
@@ -48,19 +146,45 @@
    #
    # 1. CentOS seems to have only fakeroot, which is in EPEL, not the standard
    #    repos.
+   #
+   # 2. Enabling EPEL can have undesirable side effects, e.g. different
+   #    version of things in the base repo that breaks other things. Thus,
+   #    when we install EPEL, we don't enable it. Existing EPEL installations
+   #    are left alone.
```
j-ogas, line 153: Thought I saw a new `--enable` line in the epel configuration of the centos7 image above. Is that for testing?

j-ogas (approved):
Assuming the github action tests pass, looks good to me.

lib/fakeroot.py
trandles-lanl: @j-ogas see below around line 169. When epel-release is installed automatically via `--force` is it disabling the epel repo.

trandles-lanl (approved):
This looks good to me and addresses my earlier concerns with the `--force` behavior.

heasterday (approved):

PR #917, opened 2020-12-03
ch-fromhost: clarify docs

reidpr:
Addresses #877.


doc/ch-fromhost_desc.rst
```diff
@@ -11,8 +11,13 @@ Description
...
 
-   This command is experimental. Features may be incomplete and/or buggy.
-   Please report any issues you find, so we can fix them!
+   This command does a lot of heuristic magic; it is not a general file
+   copy-to-image tool. For that, the best practice is :code:`COPY` instruction
+   within your Dockerfile. :code:`cp(1)` into an unpacked image will also work
+   but is discouraged.
+
+   It is known to be somewhat buggy. Please report any issues you find, so we
+   can fix them!
```
heasterday, line 20: What are your thoughts on an example of when you would use `ch-fromhost` vs `COPY`? I think it would really help here.

reidpr: Good point. I revised things; how does it look now?

heasterday (approved):
Looks great

Issue #909, opened 2020-11-24
checking if " /usr/bin/python3 -s" starts with slash... no


hroncok:
Hello. In Fedora, we use the upstream pathfix.py tool https://github.com/python/cpython/blob/master/Tools/scripts/pathfix.py to fix the shebang line of `/usr/bin/sphinx-build`. It looks like this:

```
#! /usr/bin/python3 -s
```

The configure script here fails parsing that shebang line:

```
checking whether sphinx-build executable path has been provided... /usr/bin/sphinx-build
checking if /usr/bin/sphinx-build is executable... ok
checking if /usr/bin/sphinx-build version >= 1.2.3... ok (3.3.1)
checking for sphinx-build Python...  /usr/bin/python3 -s
checking if " /usr/bin/python3 -s" starts with slash... no
configure: error: --with-sphinx-python: must start with slash
```

I would offer a fix, but my autotools skills are low, sorry about that.

reidpr:
Thank you Miro! Definitely a bug and appreciate the report. Fortunately I think a fix will be relatively straightforward. We are happy to look into it, but if you'd like to try, the `sed` command on line 213 of `configure.ac` is where the problem lies. Currently we just strip the first two characters of the shebang line. If this regex were just a little more knowledgeable about the syntax of shebang lines, I think it would fix this bug.

hroncok:
https://github.com/hpc/charliecloud/pull/910

reidpr:
Assigning b/c that's how we track credit.

loveshack:
This doesn't work for me.  I just tried to update the rpm to get the current version into EPEL, and it fails with this (in EL7)
```
checking if /usr/bin/python3 -s is executable... no
```
I changed the second clause in the sed expression to ` -e 's/\s.*$//'` to get it to build.

[I'll raise an issue against the Fedora package, but in case that's not being maintained now, I could update it if I got access. I'll need to build it in open build service now anyway.]

hroncok:
Note that there is a sphinx bugzilla open in Fedora to drop the -s flag anyway. I'm on mobile, so I don't have the link.

reidpr:
Hi @loveshack, thanks for the report. Can you file a new issue saying what you would like us to do, e.g. change the sed expression again as you describe?

hroncok:
> Note that there is a sphinx bugzilla open in Fedora to drop the -s flag anyway. I'm on mobile, so I don't have the link.

[...]

PR #921, opened 2020-12-08
configure: fix --with-sphinx-python

j-ogas:
See #909.


reidpr:
This is a follow-up to #909.

Issue #863, opened 2020-10-01
Image layers that contain massive whiteouts take a long time to ch-grow pull


qwofford:
To reproduce in charliecloud v0.19:

`ch-grow pull qwofford/bsp_prototype:carc-wheeler`

the "resolving whiteouts" step takes a long time if there are a lot of whtieouts between image layers:
src ref: https://github.com/hpc/charliecloud/blob/fd61c292f9f5d879f2903ac277afe2249fda020b/lib/charliecloud.py#L399-L444

One solution would be to add a message along with "resolving whiteouts" that warns the use about this. Ultimately the user would probably not want to add and remove a lot of files between contianer layers. Another solution would be to make the whiteouts resolve faster some other way. Resolving whiteouts does not seem to take as long in other container runtimes. Not clear if this is a consequence of rootless operation.

PR #922, opened 2020-12-08
ch-grow pull: update resolving whiteouts

mphinney1100:
Addresses #863


lib/charliecloud.py
```diff
@@ -479,7 +478,7 @@ def whiteouts_resolve(self, layers):
          if (wo_ct > 0):
             DEBUG("layer %d/%d: %s: processed %d whiteouts; %d members ignored"
                   % (i, len(layers), lh[:7], wo_ct, ig_ct))
-
+
```
j-ogas, line 481:
```suggestion

```

j-ogas: Removes spurious white space added.

lib/charliecloud.py
```diff
@@ -446,7 +445,7 @@ def whiteout_rm_prefix(self, layers, max_i, prefix):
          if (i > max_i): break
          members2 = list(members.keys())  # copy b/c we'll alter members
          for m in members2:
-            if (os.path.commonpath([prefix, m.name]) == prefix):
+             if (prefix == m.name[:len(prefix)]):
```
j-ogas, line 448:
```suggestion
            if (prefix == m.name[:len(prefix)]):  
```

j-ogas: Remove white space added.

j-ogas (changes requested):

Issue #913, opened 2020-12-02
configure: reports “documentation ... yes” with --disable-html --disable-man


reidpr:
Some confusing output with `./configure --disable-html --disable-man`:

```
  documentation: yes
    sphinx-build(1) ≥ 1.2.3 ... ok (3.3.1)
    sphinx-build(1) Python ... /usr/bin/python3
    "docutils" module ≥ 0.14 ... ok (0.16)
    "sphinx-rtd-theme" module ≥ 0.2.4 ... ok (0.5.0)

  will build and install:
    HTML documentation ... no
    man pages ... no
    test suite ... yes
    ch-grow(1) ... no
```

```
  complete test suite: yes
    recommended tests with SquashFS ... yes
    documentation ... yes
    ShellCheck ≥ 0.6.0 ... ok (0.7.0)
    generic sudo ... assumed yes
```

In both cases seems like it should be reporting "no" for the documentation.

Another question: What should the report say with just *one* of `--disable-html` and `--disable-man`?

PR #927, opened 2020-12-14
configure:  reports for documentation

mphinney1100:
Addresses #913

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -413,8 +413,7 @@ CH_CHECK_VERSION([WGET], [$vmin_wget], [--version | head -1 | cut -d' ' -f3])
 AS_IF([   test -n "$SPHINX" \
        && test -n "$DOCUTILS" \
        && test -n "$RTD" \
-       && (test "$enable_man" = yes \ 
-       || test "$enable_html" = yes)],
+       && (test "$enable_man" = yes || test "$enable_html" = yes)],
        [have_docs=yes],
        [have_docs=no])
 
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -412,7 +412,9 @@ CH_CHECK_VERSION([WGET], [$vmin_wget], [--version | head -1 | cut -d' ' -f3])
 
 AS_IF([   test -n "$SPHINX" \
        && test -n "$DOCUTILS" \
-       && test -n "$RTD"],
+       && test -n "$RTD" \
+       && (test "$enable_man" = yes \ 
+       || test "$enable_html" = yes)],
```
reidpr, line 417:
```suggestion
       && (test "$enable_man" = yes || test "$enable_html" = yes)],
```

reidpr (changes requested):
Nice work! One suggestion in-line for formatting and two other items.

1. I think we should maybe move this block:

```
  will build and install:
    HTML documentation ... no
    man pages ... no
    test suite ... yes
    ch-image(1) ... no
```

to be first in the "Building Charliecloud" section, b/c otherwise we say `documentation: no` but everything in that block seems favorable.

2. This should say `complete test suite: no` because of `documentation: no`. (It's not a mistake you made; you've exposed an existing bug.)

```
  complete test suite: yes
    recommended tests with SquashFS ... yes
    documentation ... no
    ShellCheck ≥ 0.6.0 ... ok (0.7.0)
    generic sudo ... assumed yes
```

PR #936, opened 2020-12-22
script to count lines of code

reidpr:
New script `misc/loc`. It counts the lines of code in the project.

Note, I did add it to CI. The advantage of this is that it won't get stale and need to be fixed every time we want to look up the numbers, and it makes the numbers easily available. The disadvantage is that it's another place to update a list of files.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,6 +56,12 @@ jobs:
           free -m
           df -h
 
+      - name: lines of code
+        if: ${{ matrix.builder == 'none' }}
+        run: |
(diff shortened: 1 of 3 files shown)


shanegoff (commented):
I get some small errors on Ubuntu 18.04
image

image

The errors are non fatal but might mess with the lines counted

shanegoff:
Cloc is version 1.74 on Ubuntu 18.04

j-ogas (approved):

shanegoff (approved):

Issue #760, opened 2020-06-10
configure: pack images report inconsistent


reidpr:
This should be `no`:

```
  pack images from builder storage to tarball: yes
    at least one builder ... no
    tar(1) ... tar
```

PR #937, opened 2020-12-22
configure:  fix pack image builder to tar 

mphinney1100:
Addresses #760

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -479,8 +479,8 @@ AS_IF([   test -n "$TAR"],
       [have_pack_tar=yes],
       [have_pack_tar=no])
 
-AS_IF([   test $have_pack_tar = yes \
-       && test $have_any_builder = yes],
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -479,6 +479,11 @@ AS_IF([   test -n "$TAR"],
       [have_pack_tar=yes],
       [have_pack_tar=no])
 
+AS_IF([   test $have_pack_tar = yes \
+       && test $have_any_builder = yes],
```
reidpr, line 483: Switch the order so it matches the report:

```suggestion
AS_IF([   test $have_any_builder = yes \
       && test $have_pack_tar = yes],
```

reidpr (changes requested):
Great work! One style adjustment, otherwise looks ready to merge. Tag it `ready to merge?` when you agree.

Issue #703, opened 2020-03-20
ch-test: allow execution of individual tests


reidpr:
Split from #607.

It would be nice if we could use `ch-test` to execute individual tests, e.g.:

```
$ ch-test './build.bats:lint shell scripts'
```

This maybe should imply full scope, since if the user is going to the trouble of requesting a specific test, they probably want to run it even if the (perhaps unstated) scope argument says otherwise.

PR #938, opened 2020-12-23
ch-test: add individual tests to --file

reidpr:
Addresses #703.

Caveats/gotchas:

1. The exclusion happens in `common.bash:scope()`, so any test that does not call `scope` cannot be excluded. I don't recall if tests can do so and get a default scope.
2. It's implemented by skipping, so you still have to pay the skip overhead. But, this was a much simpler way to do it than editing the files on the fly somehow.

### DIFF
--- a/bin/ch-test
+++ b/bin/ch-test
@@ -43,7 +43,7 @@ Additional options:
 
   -b, --builder BUILDER  image builder to use
   --dry-run              print summary of what would be tested and exit
-  --file FILE            run tests in FILE only
+  --file FILE[:TEST]     run tests in FILE only, or even just TEST
   -h, --help             print this help and exit
   --img-dir DIR          unpacked images directory; same as \$CH_TEST_IMGDIR
   --pack-dir DIR         packed imaged directory; same as \$CH_TEST_TARDIR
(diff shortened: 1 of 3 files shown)


bin/ch-test
```diff
@@ -869,7 +874,7 @@ fi
 
 printf "%-21s %s" 'phase:' "$phase"
 if [[ $phase = one-file ]]; then
-    printf ': %s' "$one_file"
+    printf ': %s (%s)' "$one_file" "$ch_one_test"
```
j-ogas, line 877: We are OK with empty parens if the one test variable is empty?

test/common.bash
```diff
@@ -134,6 +134,14 @@ run () {
 }
 
 scope () {
+    if [[ -n $ch_one_test ]]; then
+        # Ignore scope if a single test is given.
+        if [[ $ch_one_test != "$BATS_TEST_DESCRIPTION" ]]; then
```
j-ogas, line 139: If ch_one_test is gobbledegook, i.e., doesn't match anything, the test suite will pass?

Should it fail if the argument doesn't match at least one test description?

bin/ch-test
reidpr: I guessed that was OK since the audience is debugging and we have a caveat to that effect in the docs.

test/common.bash
reidpr: Yes, it will skip everything and pass.

I could add a counter somewhere and make sure exactly one test is matched. But, like above, I guessed it was an OK trade-off given the audience.

I could also add a sentence in the docs explaining this behavior. Opinions?

j-ogas: I don't have strong feelings in any direction. Just a thought.

Issue #935, opened 2020-12-22
remove packaging/debian?


reidpr:
We have some Debian packaging material in the Charliecloud source code (directory `packaging/debian`, but it's unmaintained and AFAICT not used by anyone. The [actual Debian package](https://packages.debian.org/source/charliecloud) is ably maintained [elsewhere](https://salsa.debian.org/hpc-team/charliecloud). Thus, I'm inclined to remove `packaging/debian`.

Related: #113, #390.

@wiene @lnussbaum, any comments?

wiene:
> Thus, I'm inclined to remove `packaging/debian`.

This is fine with me.

lnussbaum:
Fine with me as well

Issue #390, opened 2019-03-18
refactor debian package with new test suite and example install path


j-ogas:
The change to the charliecloud test suite and example installation location introduced in #338 broke our debian package. Thus, the building and testing the of the debian package is currently being skipped.

reidpr:
After #935, this is moot.

Issue #113, opened 2018-02-14
make clean should delete leftover packaging stuff


reidpr:
I have a bunch of files under `packaging/debian` that seem to be build products. I haven't checked the corresponding Red Hat stuff.

It would be nice if `make clean` deleted this stuff, since I keep bumping into it while searching for unrelated things.

@olifre, @wiene, @lnussbaum — is this a trivial change for any of you?

reidpr:
I looked into this briefly and I don't understand what's going on well enough to reliably implement it myself.

Also I see that the naïve `debuild` command from `travis.sh` puts an `html` directory at the root, which seems strange. Thus, I'm leaving this alone for now.

People who run into this need can be guided with the content of `.gitignore` for now.

reidpr:
After closing #935, this is moot.

PR #941, opened 2020-12-23
remove packages/debian & everything related

reidpr:
Addresses #935. Would also make #113 and #390 moot.


packaging/Makefile.am
```diff
@@ -1,18 +1,3 @@
...
-debian/charliecloud.install \
-debian/compat \
-debian/control \
-debian/copyright \
-debian/docs \
-debian/format \
-debian/rules \
-debian/travis.sh \
-debian/watch \
-fedora/build \
```
j-ogas, line 14: We don't want to remove the fedora packaging bits do we?

j-ogas (changes requested):
See fedora package comment in the makefile.

packaging/Makefile.am
reidpr: foo

j-ogas (approved):

Issue #897, opened 2020-11-06
ch-image: can't pull public images from nVidia NGC registry


reidpr:
`ch-grow pull` does not work on [nVidia NGC](https://ngc.nvidia.com) images. For example:

```console
$ bin/ch-grow pull nvcr.io/nvidia/cuda:11.1-devel-ubuntu16.04
pulling image:   nvcr.io/nvidia/cuda:11.1-devel-ubuntu16.04
manifest: downloading
error: WWW-Authenticate missing key: service
```

Note, this may be hard to test because all the images I found were humongous.

ajyounge:
NGC requires an API key, which needs to be procured ahead of time. Maybe just need a mechanism to point to a shared key file? I wish this NGC requirement would just go away.

https://docs.nvidia.com/dgx/ngc-registry-for-dgx-user-guide/

cr458:
Hi, is there any way around this currently?

reidpr:
Well, `ch-grow` makes assumptions that aren't correct for NGC. But, if you can build with Docker or Buildah, they probably are able to pull from NGC.

Also if you want to contact me behind the firewall I would be more than happy to discuss your specific needs a bit and try to come up with a path forward.

reidpr:
This image might be good for testing; it's only 40MiB: https://ngc.nvidia.com/catalog/containers/hpc:foldingathome:fah-gpu

reidpr:
It seems NGC does allow anonymous access now?

PR #945, opened 2020-12-24
ch-image pull: don't require WWW-Authenticate key "service"

reidpr:
Addresses #897.

Note: I only tried anonymous access.


j-ogas (approved):

reidpr:
Removed `ready to merge?` pending feedback from @cr458 and colleagues.

Also: @ajyounge, does this work for you?

Issue #886, opened 2020-10-22
`ch-run --set-env`: expand environment variables


DavidBrayford:
Would it be possible to have a prepend PATH/LD_LIBRARY option in ch-run?

The example I have is a user wants to use the original Docker environment, but want to use executables and/or libraries from the host. For example, A user want to use the performance tools (perf), debugger, profiler that reside on the host system on a containerized application running on a HPC system

The following commands work,
```
ch-run --set-env=/image/ch/environment -b /host/dir/.:/host/dir image -- bash
export PATH=/host/dir/bin:$PATH
export LD_LIBRARY_PATH=/host/dir/lib:$LD_LIBRARY_PATH
 likwid-perfctr -m -g L2CACHE -C 0 julia mandelbrot.jl
```
However, passing in the parameters to ch-run be preferable when executing via a batch system such as Slurm.
For example,

`mpiexec -n 16 ch-run --set-env=/image/ch/environment --prepend-path=/host/dir/bin --prepend-ld_library_path=/host/dir/lib -b /host/dir/.:/host/dir image --   likwid-perfctr -m -g L2CACHE -C 0 julia mandelbrot.jl`

David

reidpr:
Hi David, thanks for the feature request. Let me think a bit about how best to accomplish this.

Couple other workarounds that may work now:

* The container will read your dotfiles, so the exports could go there.
* `ch-run` mostly does not change environment variables, so they can be exported before `ch-run`.
* Can the bind-mount go to a directory that's already in the paths (e.g., `/usr/local/bin` and `/usr/local/lib`)?

PR #948, opened 2021-01-05
ch-run --set-env: updates PATH

mphinney1100:
Addresses #886


reidpr (changes requested):
Per discussion offline.

bin/ch-run.c
```diff
@@ -51,6 +51,7 @@ const struct argp_option options[] = {
    { "join-pid",     -5, "PID",  0, "join a namespace using a PID" },
    { "join-ct",      -3, "N",    0, "number of ch-run peers (implies --join)" },
    { "join-tag",     -4, "TAG",  0, "label for peer group (implies --join)" },
+   { "no-expand",   -10, 0,      0, "don't expand $ in --set-env input"},
```
reidpr, line 54: I'd call this `--set-env-no-expand`, or something else that ties it to `--set-env`.

bin/ch-run.c
```diff
@@ -112,6 +114,7 @@ int main(int argc, char *argv[])
    args = (struct args){ .c = (struct container){ .ch_ssh = false,
                                                   .container_gid = getegid(),
                                                   .container_uid = geteuid(),
+                                                  .expand = true,
```
reidpr, line 117: Same here — need a little more context on what is being expanded.

bin/ch-run.c
```diff
@@ -184,6 +187,36 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
    (*ds)[i].arg = arg;
 }
 
+/* Parse and set input from --set-env */
```
reidpr, line 190: This function I think needs more commenting explaining the algorithm and its uses.

bin/ch-run.c
```diff
@@ -184,6 +187,36 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
...
+void env_parse(char *name, char *new_value, bool expand)
+{
+   char *token;
+   char *new_env = "";
```
reidpr, line 194: I'd initialized this below, just before the loop.

bin/ch-run.c
```diff
@@ -184,6 +187,36 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
...
+   const char s[1] = ":";
+   if (   strlen(new_value) >= 2
+       && (new_value)[0] == '\''
+       && (new_value)[strlen(new_value) - 1] == '\'') {
+      (new_value)[strlen(new_value) - 1] = 0;  // strip trailing quote
+      (new_value)++;                           // strip leading
+   }
```
reidpr, line 201: Comment this block to say it's stripping leading and trailing single quotes.

bin/ch-run.c
```diff
@@ -184,6 +187,36 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
    (*ds)[i].arg = arg;
 }
 
+/* Parse and set input from --set-env */
+void env_parse(char *name, char *new_value, bool expand)
+{
+   char *token;
+   char *new_env = "";
+   const char s[1] = ":";
```
reidpr, line 195: This variable is used twice and gives you room to change the separator, but I think that amount of indirection outweighs the additional cognitive load it adds. I'd just use literals below with `strtok()`.

Also, it's an unterminated string b/c the length is 1.

bin/ch-run.c
```diff
@@ -184,6 +187,36 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
...
+       && (new_value)[strlen(new_value) - 1] == '\'') {
+      (new_value)[strlen(new_value) - 1] = 0;  // strip trailing quote
+      (new_value)++;                           // strip leading
+   }
+   token = strtok(new_value, s);
+   while( token != NULL ) {
+      if (token[0] == '$' && expand) {
+         token ++;
+         if(getenv(token) != NULL)
+            T_ (1 <= asprintf(&new_env, "%s:%s", new_env, getenv(token)));
```
reidpr, line 207: Can this use our `cat()`?

bin/ch-run.c
```diff
@@ -184,6 +187,36 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
...
+      }
+      else {
+         T_ (1 <= asprintf(&new_env, "%s:%s", new_env, token));
```
reidpr, line 210: Does this work with an empty item at the start or end of the string (i.e. leading/trailing colon)?

bin/ch-run.c
```diff
@@ -184,6 +187,36 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
...
+      }
+      token = strtok(NULL, s);
+   }
+   if (new_env[0] == ':')
+      new_env++;
```
reidpr, line 215: This doesn't quite match the docs. We can change either but I think you are probably better off omitting the two lines and dealing with it elsewhere.

bin/ch-run.c
```diff
@@ -184,6 +187,36 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
...
+   INFO("environment: %s=%s", name, new_env);
+   Z_ (setenv(name, new_env, 1));
```
reidpr, line 217: I recommend you make this function about expansion and put setting elsewhere.

bin/ch-run.c
```diff
@@ -215,8 +248,15 @@ void fix_environment(struct args *args)
    for (int i = 0; args->env_deltas[i].action != END; i++) {
       char *arg = args->env_deltas[i].arg;
       if (args->env_deltas[i].action == SET_FILE) {
```
reidpr, line 250: `SET_FILE` might be misnamed now?

bin/ch-run.c
```diff
@@ -215,8 +248,15 @@ void fix_environment(struct args *args)
...
-         FILE *fp;
-         Tf (fp = fopen(arg, "r"), "--set-env: can't open: %s", arg);
+         FILE *fp = NULL;
+         if (strchr(arg, '=') == NULL) {
+            Tf (fp = fopen(arg, "r"), "--set-env: can't open: %s", arg);
+         }
+         else {  
+            split(&name, &new_value, arg, '=');
+            env_parse(name, new_value, args->c.expand);
```
reidpr, line 257: I think you should structure this a bit differently. Read the file and create a list of lines; *or* create a single-item list of lines. Then both lists can go through the same loop.

bin/ch-run.c
```diff
@@ -374,6 +407,10 @@ static error_t parse_opt(int key, char *arg, struct argp_state *state)
    int i;
 
    switch (key) {
+   case -10: // --no-expand
+      args->c.expand = false;
+      INFO("expand: false");
```
reidpr, line 412: Can probably rm this `INFO`?

doc/ch-run_desc.rst
```diff
@@ -37,6 +37,9 @@ unpacked image directory located at :code:`NEWROOT`.
     label for :code:`ch-run` peer group (implies :code:`--join`; default: see
     below)
 
+  :code:`--no-expand`
+    do not expand variables when using :code:`--set-env`
```
reidpr, line 41: Style: "don't".

reidpr (commented):
Great work as usual! Various suggestions in-line.

I didn't build the code, but make sure you aren't getting any compiler warnings.

I didn't review the tests this time around.

bin/ch-run.c
```diff
@@ -184,6 +188,63 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
...
+{
+   int i;
+   char *name, *new_value;
+   for(i = 0; i < argv; i++) {
+      if (strlen(env_set[i]) == 0 || env_set[i][0] == '\n')
+         continue; 
+      split(&name, &new_value, env_set[i], '=');
+      env_expand_error(name, filename, i+1);
+
+      // strip leading and trailing single quotes
```
j-ogas, line 204: Can there be double quotes, e.g., `"`?

bin/ch-run.c
```diff
@@ -184,6 +188,63 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
...
+      if (   strlen(new_value) >= 2
+          && (new_value)[0] == '\''
+          && (new_value)[strlen(new_value) - 1] == '\'') {
+         (new_value)[strlen(new_value) - 1] = 0;  
+         (new_value)++;                           
+      }
+   
+      char *token, *r;
+      char *new_env = "";
+      r = strdup(new_value);
```
j-ogas, line 214: What is the significance of `r` as a variable name?

bin/ch-run.c
```diff
@@ -212,11 +273,22 @@ void fix_environment(struct args *args)
    }
 
    // --set-env and --unset-env.
+   char* env_set[50];
```
j-ogas, line 276: What is the significance of 50? Should this be a macro, e.g., `#define ENV_MAX`?

j-ogas (changes requested):
Looking good Megan 👍 . Some comments and questions below.

bin/ch-run.c
mphinney1100: That was what the copy of the value being tokenized was called in the documentation for strsep. Looking back I actually don't need this variable so its getting deleted.

bin/ch-run.c
mphinney1100: Double quotes don't get stripped.

bin/ch-run.c
```diff
@@ -65,7 +66,7 @@ const struct argp_option options[] = {
 
 /* One possible future here is that fix_environment() ends up in ch_base.c and
    we add other actions such as SET, APPEND_PATH, etc. */
-enum env_action { END = 0, SET_FILE, UNSET_GLOB };
+enum env_action { END = 0, SET_ENV, UNSET_GLOB };
```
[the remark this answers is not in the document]
mphinney1100: This PR does effect this comment. I think the comment should be removed now the --set-env does all path operation. Now it doesn't seem likely that new actions will be added to env_action.

bin/ch-run.c
```diff
@@ -273,7 +271,7 @@ void fix_environment(struct args *args)
    }
 
    // --set-env and --unset-env.
-   char* env_set[50];
+   char* env_set[ENV_MAX];
```
j-ogas, line 274:
```suggestion
   char *env_set[ENV_MAX];
```

bin/ch-run.c
```diff
@@ -188,9 +187,10 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
    (*ds)[i].arg = arg;
 }
 
-/* Parse each line from --set-env at ':'. Each value is appended to the new
-   environment variable. '$' indicated that the variable will get expanded
-   except if --env-no-expand option is included. */
+/* Read input variable value pairs from --set-env. Split at '=' and tokenize
+   value on ':'. Each tokenize value append to new environment variable value.
+   '$' indicates that the variable will get expanded except if --env-no-expand
+   option is included. */
```
j-ogas, line 193:
```suggestion
/* Read input variable value pairs from --set-env and split on '='. Tokenize
   value on ':' and append each token to new environment variable value.
   Variables beginning with '$' are expanded unless the expand boolean is
   false. */
```

j-ogas: Re. comment above, this would be my suggestion. However, I wouldn't get too hung up on this as @reidpr may (will) have even better suggestions.

bin/ch-run.c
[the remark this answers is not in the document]
j-ogas: We should defer to @reidpr.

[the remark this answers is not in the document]
reidpr: I'm happy to remove the comment. It may be good to replace it with a more generic comment saying what the enum is for.

bin/ch-run.c
reidpr: I would avoid verbs like "tokenize" because they're quite jargony. What does it mean to "tokenize on :"? I don't know.

That said, you may not need to say much about what this does since the behavior is pretty clear in the docs. It's legit to refer to those in comments.

bin/ch-run.c
```diff
@@ -4,6 +4,7 @@
    are modest and the program is short-lived. */
 
 #define _GNU_SOURCE
+#define ENV_MAX 50
```
reidpr, line 7: Put this in "constants and macros" section.

bin/ch-run.c
```diff
@@ -227,22 +297,12 @@ void fix_environment(struct args *args)
...
-            if (   strlen(new_value) >= 2
-                && new_value[0] == '\''
-                && new_value[strlen(new_value) - 1] == '\'') {
-               new_value[strlen(new_value) - 1] = 0;  // strip trailing quote
-               new_value++;                           // strip leading
-            }
-            INFO("environment: %s=%s", name, new_value);
-            Z_ (setenv(name, new_value, 1));
+            env_set[env_setv] = line;
+            env_setv++;
```
reidpr, line 303: Need to make sure to check for overflow of `env_set`.

bin/ch-run.c
```diff
@@ -212,40 +274,38 @@ void fix_environment(struct args *args)
...
+               if (-1 == getline(&line, &line_len, fp)) {
+                  if (errno == 0)  // EOF
+                     break;        // note: line_ct not incremented
+                  else
+                     Tf (0, "--set-env: can't read: %s", arg);
+               }
+               if (line[strlen(line) - 1] == '\n')  // rm newline if present
+                  line[strlen(line) - 1] = 0;
+               T_ (lines = realloc(lines, (line_ct + 1) * sizeof(char *)));
+               lines[line_ct] = line;
```
reidpr, line 302: I changed this from a fixed-length array on the stack to a dynamically allocated/expanded array on the heap. See also #1059.

Note that if `MAX_ENV` had stayed, it would need to be documented, and a value of at least ~1000 would probably be needed.

bin/ch-run.c
```diff
@@ -176,14 +178,74 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
...
+      while (item != NULL) {                       // loop executes ≥ once
+         if (   expand                             // expansion requested
+             && item[0] == '$' && item[1] != 0) {  // ≥1 char in variable name
+            item = getenv(++item);                 // NULL if unset
+            if (item != NULL && item[0] == 0)
+               item = NULL;                        // convert empty to unset
+         }
+         if (item != NULL) {                       // NULL -> omit from output
+            if (first_written)
+               value_new = cat(value_new, ":");
```
reidpr, line 236: I reversed the sense here, to prepend colons if not the first item, rather than append if not the last item. When we last discussed this, the two approaches seemed roughly equivalent. But, I realized that we can reliably tell if we've written the first item, but not if an item is the last one (assuming that unset variables expand to nothing rather than the empty string).

bin/ch-run.c
```diff
@@ -176,14 +178,74 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
...
+      }
+
+      // Walk through value fragments separated by colon and expand variables
+      // per documentation.
+      value_new = "";
+      first_written = false;
+      item = strsep(&value_old, ":");              // cannot be NULL
+      while (item != NULL) {                       // loop executes ≥ once
+         if (   expand                             // expansion requested
+             && item[0] == '$' && item[1] != 0) {  // ≥1 char in variable name
```
reidpr, line 229: Note the short-circuit evaluation here, which is very important. If we said `item[1]` before we knew the string was at least one character long, that's undefined behavior which is very problematic in C because it might work sometimes and might crash sometimes and might break something elsewhere in the program (C is *not* a “try it and see” language like Python mostly is). But, because the boolean is short-circuited, we only read `item[1]` after we know `item[0] != 0` (because it's `$`), which tells us `item[1]` must exist.

bin/ch-run.c
```diff
@@ -212,40 +274,38 @@ void fix_environment(struct args *args)
...
-            errno = 0;
-            if (-1 == getline(&line, &len, fp)) {
-               if (errno == 0)  // EOF
-                  break;
-               else             // error
-                  Tf (0, "--set-env: error reading: %s", arg);
+      if (args->env_deltas[i].action == ENV_SET) {  // --set-env
+         if (strchr(arg, '=') != NULL) {
+            // argument is variable name & value
+            envs_set((char *[]){ arg }, 1, NULL, args->c.env_expand);
```
reidpr, line 282: Note the array literal here (`(char *[]){ arg }`), which is a C99 thing and has a quite hairy syntax.

bin/ch-run.c
```diff
@@ -212,40 +274,38 @@ void fix_environment(struct args *args)
...
+         } else {
```
reidpr, line 283: Note the non-file branch of the test goes first. This is to keep the `else` close to the condition.

bin/ch-run.c
```diff
@@ -212,40 +274,38 @@ void fix_environment(struct args *args)
...
+            int line_ct = 0;
+            FILE *fp;
+            Tf (fp = fopen(arg, "r"), "--set-env: can't open: %s", arg);
+            for (line_ct = 0; true; line_ct++) {
+               char *line;
+               size_t line_len = 0;  // don't care but getline(3) must write
+               errno = 0;
+               if (-1 == getline(&line, &line_len, fp)) {
+                  if (errno == 0)  // EOF
+                     break;        // note: line_ct not incremented
```
reidpr, line 295: I couldn't remember whether the third component of the for loop was executed after `break` (here `line_ct++`), and it was surprisingly hard to look up. I had to write a test program. Hence the comment.

bin/ch-run.c
```diff
@@ -176,14 +178,74 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
...
+         T_ (1 <= asprintf(&lineno_str, ":%d", i+1));
+      Te (name != NULL, "--set-env: no delimiter: %s%s", filename, lineno_str);
+      Te (name[0] != 0, "--set-env: empty name: %s%s", filename, lineno_str);
+
+      // Strip leading and trailing single quotes from value, if both present.
+      if (   strlen(value_old) >= 2
+          && value_old[0] == '\''
+          && value_old[strlen(value_old) - 1] == '\'') {
+         value_old[strlen(value_old) - 1] = 0;
+         value_old++;
```
reidpr, line 219: Re. our wild memory leaking policy: This trick to delete the leading single quote by simply incrementing the pointer past it means we couldn't call `free(value_old)` even if we'd received `value_old` from `malloc(3)` (which I'm not sure is true). I think we could do `free(value_old - 1)`, though this type of arithmetic is one place where pointer arithmetic differs on whether `value_old` is an array or pointer, except in this case where the array elements have length 1.

bin/ch-run.c
```diff
@@ -212,40 +274,38 @@ void fix_environment(struct args *args)
...
+            // argument is filename
+            char **lines = NULL;
+            int line_ct = 0;
+            FILE *fp;
+            Tf (fp = fopen(arg, "r"), "--set-env: can't open: %s", arg);
+            for (line_ct = 0; true; line_ct++) {
+               char *line;
+               size_t line_len = 0;  // don't care but getline(3) must write
+               errno = 0;
+               if (-1 == getline(&line, &line_len, fp)) {
```
reidpr, line 293: Note the incredibly awkward interface for `getline(3)` error checking. This type of thing is one motivation for exceptions, but even in C this could have been done better. E.g. why not return zero on EOF? The only time I can think of that one could read zero bytes is the last line of the file if it doesn't have a newline, but in that case the previous line would be the last line. I think there must be some subtlety I'm missing.

bin/ch-run.c
```diff
@@ -176,14 +178,74 @@ void env_delta_append(struct env_delta **ds, enum env_action act, char *arg)
...
+          && value_old[0] == '\''
+          && value_old[strlen(value_old) - 1] == '\'') {
+         value_old[strlen(value_old) - 1] = 0;
+         value_old++;
+      }
+
+      // Walk through value fragments separated by colon and expand variables
+      // per documentation.
+      value_new = "";
+      first_written = false;
```
reidpr, line 225: I could have avoided this variable by changing `cat()` to accept a null pointer and treat it as the empty string (and initialize `value_new` to `NULL`). (We can't test whether `value_new` is the empty string because the first item written might be the empty string, and in that case we still need a delimiter.) Would that have been better? 🤷‍♂️ I'm open to argument.

reidpr:
The test failure is because Sphinx 4.0 came out; [bloody hell](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-man_make_section_directory).

bin/ch-run.c
mphinney1100: I think changing cat() to treat NULL as an empty string would be cleaner then having the extra bool variable.

reidpr: Feel free to make that change then! :)

reidpr (approved):

Issue #947, opened 2021-01-05
build fails, cannot find base.sh


agilly:
Hello,

I followed the tarball install instructions:

```
wget https://github.com/hpc/charliecloud/releases/download/v0.21/charliecloud-0.21.tar.gz
tar -xvzf charlie*gz
cd charliecloud-0.21
./configure
make
sudo make install
```

Then I created the example docker file with a centos base from the install docs. When I try to build, I get:
```
$> ch-build -t hello .
/usr/local/bin/ch-build: line 4: /usr/local/bin/../lib/charliecloud/base.sh: No such file or directory
```

There seems to be something wrong with paths, links and permissions in general, as an example I cannot run the example since `cd /usr/local/share/doc/charliecloud/examples/hello` throws a permission error.

My OS is OpenSuse 15 and my shell is bash. The full install logs can be found at [https://pastebin.pl/view/6d63e8d3](https://pastebin.pl/view/6d63e8d3).

I also tried pulling directly from the repo and using autoconf, it fails with the same result.

reidpr:
Hi Arthur, I'm sorry for the trouble and we really appreciate the bug report.

Your logs link times out for me — can you try a different pastebin? I believe GitHub Gists have worked for us in the past.

That said, something is definitely wrong with permissions in `sudo make install`:

```
$ rm -Rf /tmp/foo
$ ./configure --prefix=/tmp/foo
$ make
$ sudo make install
$ /tmp/foo/bin/ch-build --help
-bash: /tmp/foo/bin/ch-build: Permission denied
$ ls -ld /tmp/foo
drwxr-x--- 6 root root 4096 Jan  5 10:57 /tmp/foo
```

This happens in 0.20 as well.

As a workaround, you can probably patch up the permissions manually, maybe with something like `chmod o+rX`?

Again I really appreciate your pointing this out! We will get it fixed ASAP.

reidpr:
Appears to be umask related, so better workaround: set your umask to something permissive, e.g. 0022, then `sudo make install`.

agilly:
Hi @reidpr, thanks for being so reactive! Unfortunately neither checking out the `umask_947` branch, neither setting `umask 0022` helps, I still get the same message.

Please find the logs below, hopefully it will work this time: [https://pastebin.ubuntu.com/p/SdbpR6mpx4/](https://pastebin.ubuntu.com/p/SdbpR6mpx4/)

reidpr:
[...]

Issue #598, opened 2019-12-20
build: docs get built twice in "make install"


reidpr:
But not `make`.

Follow-up to PR #589.

reidpr:
I think now we might also be getting multiple builds in just `make`. Certainly, two `(cd doc && make)` in a row builds the docs twice; it should only build them once, on the first invocation.

PR #949, opened 2021-01-05
fix "sudo make install" with non-trivial umask

reidpr:
Also addresses #598. Related: #947, but that's multiple problems.

Previously, `make install` respected umask for directories, but not files. Thus, if you `sudo make install` with a restrictive umask such as 0027 or 0077, the install will be readable only by root and thus largely useless (issue #947). This is the default Autotools behavior and appears to be at least a somewhat common complaint. This PR changes `make install` to create all directories with permissions 755 regardless of umask, by installing directories with `install -d` rather than `mkdir -p`.

The old behavior is still available with `./configure MKDIR_P='mkdir -p'` in case you need it for [some reason](https://lists.gnu.org/archive/html/automake/2004-01/msg00013.html).

This PR also updates CI to `sudo install` with umask 0077; formerly it was 0022, which is why we didn't catch this bug earlier.

As part of this update, I figured out why the documentation is built too many times (issue #598) and fixed it. So, build is faster now too.

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -3,6 +3,21 @@
 # Automake-ification, I stripped out most of the boilderplate and left only
 # the targets that we use.
 
+# We turn off parallel build in doc:
+#
+#   1. Sphinx handles building the whole documentation internally already, as
+#      a unit, so we shouldn't call sphinx-build more than once for different
+#      output files at all, let alone in parallel.
+#
+#   2. Serial build is plenty fast.
+#
+#   3. There is a race condition in Sphinx < 1.6.6 that's triggered when two
+#      instances (e.g., for html and man targets) try to "mkdir doctrees"
+#      simultaneously. See issue #115.
+#
+# This special target was introduced in GNU Make 3.79, in April 2000.
+.NOTPARALLEL:
+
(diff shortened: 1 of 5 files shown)


reidpr:
I think the CI failure is spurious.

trandles-lanl (approved):
This all looks sane and good to me.

doc/Makefile.am
```diff
@@ -137,37 +152,36 @@ BUILDDIR      = .
...
+#
+# [1]: https://www.gnu.org/software/make/manual/html_node/Multiple-Targets.html
+$(nobase_html_DATA) $(man_MANS): ../lib/version.txt _deps.rst $(EXTRA_DIST)
+if ENABLE_HTML
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
-# avoid GitHub messing things up with Jekyll
+#       Avoid GitHub messing things up with Jekyll.
 	touch html/.nojekyll
-# remove unused files that Sphinx made
+#       Some output files are copies with same timestamp as source; fix. Note
```
j-ogas, line 174: FIXME?

j-ogas (approved):

Issue #959, opened 2021-01-15
increase minimum Python to 3.6, and test it


reidpr:
Currently we advertise Python 3.4 compatibility, but it doesn't actually work.

reidpr:
Python takes less than a minute to build without optimizations (and their lengthy profiling phase), which we probably don't need.

reidpr:
Availability:

Distro | Version | Date
-- | -- | --
*Upstream* | 3.4 | 2014-03
*Upstream* | 3.5 | 2015-09
*Upstream* | 3.6 | 2016-12
*Upstream* | 3.7 | 2018-06
*Upstream* | 3.8 | 2019-10
*Upstream* | 3.9 | 2020-10

reidpr:
Notable new features [in 3.5](https://docs.python.org/3/whatsnew/3.5.html):

* `subprocess.run()`
* type hints (requires separate linting)
* automatic re-try of system calls on `EINTR`
* `glob.glob()` supports recursive matching with `**`
* new `math.inf` and `math.nan` constants
* new `os.path.commonpath()` function
* `pathlib.Path.mkdir()` now takes `exist_ok`
* Windows XP no longer supported

reidpr:
On 3.4, `ch-image` fails immediately with a syntax error:

```
$ bin/ch-image --version
  File "bin/ch-image", line 100
    kwargs = { **kwargs,
                ^
SyntaxError: invalid syntax
```

reidpr:
3.5 currently does not pass tests (I think this is the error that came up in the workshop):

```
 ✗ ch-build 00_tiny
[...]
     File "/usr/local/stow/python-3.5.10/lib/python3.5/genericpath.py", line 42, in isdir
       st = os.stat(s)
   TypeError: argument should be string, bytes or integer, not Path
```

reidpr:
See also #992.

PR #969, opened 2021-01-22
test with minimum Python

reidpr:
Addresses #959.


reidpr:
Note, this raises minimum Python to 3.6. If we agree that's right, I'll add it to the documentation.

j-ogas (approved):

Issue #953, opened 2021-01-08
bundle lark-parser with Charliecloud


shanegoff:
Charliecloud currently needs users to manually install lark-parser in order to use `ch-image` which could be a barrier for people. If we shipped a version of lark-parser with Charliecloud we could default immediately to using unprivileged image building with `ch-image` instead of defaulting to docker. A lot of our builder scripts default to docker if you don't specify something else. With some changes users could not have to install docker or another builder.

Options

1. Install lark-parser using `setup.py` from either a git submodule or tar release
2. Include lark-parser as a submodule and just change charliecloud module paths to look for it internally instead of in the system python
3. Make `ch-image` a python module and let pip handle the requirements of requests and lark-parser
4. ??

There could be distribution options I have not considered or learned of we could explore.
lark-parser has an MIT license so we should be able to ship it with Charliecloud.
This could help clear up confusion for users between the lark module and lark-parser module.
This could simplify the install process for users.

reidpr:
Newer Lark does have some nice features, so if it had a predictable version, we could take advantage of those.

shanegoff:
So far there are several options for bundling lark-parser. We could use a git submodule but that puts extra steps on users to do a recursive clone. We could ship a release tar, unpack it, build lark, then manually put it in the right spot in the build phase. We can also just use the installer to install it only for the user, but that might get hairy if users are running as root.

Just running `python3 setup.py buildd` then copying the output under charliecloud/lib works out of the box with very little effort.

PR #973, opened 2021-01-27
bundle Lark

shanegoff:


shanegoff:
Addresses #953

reidpr (commented):
I'm on board with this. From our discussion I think we wanted both Lark and Requests.

Some questions & comments, in no particular order:

1. Will it work to pre-install and put everything needed in the `.tar.gz` file, i.e., the user-facing build system would not do any `python setup.py`. The prototype looks like it will only work from a Git checkout. This will require updating the `Makefile.am` files.

2. Can we install the libraries with `.egg` or `.whl` files? That would make the above easier to deal with.

3. Are the libraries pure Python? If so they are going to be portable; if not, they aren't.

4. Current PR contains some empty tarballs that should be removed.

5. I think it would be good to check that we're using the embedded packages. It wouldn't be hard to validate versions; if it's not too hard to check if it's the embedded copy that would be nice too IMO.

6. I think we need to figure out a plan for storing the embedded libraries. Tarballs (how downloaded)? Submodules? `pip install` to some path?

7. Requests has dependencies; how do we deal with that? (It does seem our problems are greater with Lark than Requests.)

8. Are the Fedora people going to be annoyed?

9. Do we have half-measures available, e.g. a canned virtualenv install to a well-defined path?

Probably next steps are to discuss these and other points that come up.

verybrofessional:
- [ ] Check we are using embedded python modules
- [ ] Check versions of embedded modules
- [ ] Fix CI to stop checking for "ch-image: no" on Docker runs

j-ogas:
#1012 updates our spec file the Fedora spec file to include the lark and request packages.

For this PR we will also need to update the `packaging/fedora/charliecloud/spec` to exclude our bundled request and lark-parser bits.

verybrofessional:
@j-ogas 

We are no longer bundling requests. We have a license conflict with one of the dependencies. Requests is available in most package managers so it isn't a huge hit. Sorry for not letting you know sooner.

shanegoff:
> Will it work to pre-install and put everything needed in the .tar.gz file, i.e., the user-facing build system would not do any python setup.py. The prototype looks like it will only work from a Git checkout. This will require updating the Makefile.am files.

There are a variety of options to be considered here. I think this is still up for discussion depending on the final solution.

> Can we install the libraries with .egg or .whl files? That would make the above easier to deal with.

We can use wheel files with or without pip. We could use egg files, but we would have to build everything from source.

> Are the libraries pure Python? If so they are going to be portable; if not, they aren't.

Lark is pure python. To my knowledge requests is as well. The dependencies of requests that might not be true. Wheel files are designed to be totally portable even if the package is not pure python, so that would be a big advantage.

shanegoff:
Ok the final proposal.

We pip install our dependencies as wheels to the lib folder.
We force the use of the embedded lark-parser.
We make the embedded requests options and check the version at run time.

reidpr:
> We pip install our dependencies as wheels to the lib folder.

I'm open to argument here but a bit queasy; aren't wheels not supposed to be importable?

> We force the use of the embedded lark-parser.

+1

> We make the embedded requests options and check the version at run time.
> We would have to append every requests dependency to the import path though.

–1 

Let's defer Requests and focus on Lark in this PR.

doc/dev.rst
```diff
@@ -1246,3 +1246,23 @@ Additional sources
 * https://github.com/opencontainers/runtime-spec/blob/master/runtime.md
 
 ..  LocalWords:  milestoned gh nv cht Chacon's scottchacon
+
+Updating bundled Lark Parser
+============================
+
+In order to change the version of the blunded lark parser you must first find
```
j-ogas, line 1253:
```suggestion
In order to change the version of the bundled lark parser you must first find
```

doc/dev.rst
```diff
@@ -1246,3 +1246,23 @@ Additional sources
...
+the commit you would like to change the lark version to. Go to
+https://github.com/lark-parser/lark/tags and choose a tagged release.
```
j-ogas, line 1255: This is a little unclear. It says I need a specific commit then says to choose a tagged release. I suspect it can be either; consider revising.

lib/charliecloud.py
```diff
@@ -33,7 +36,11 @@
    # Lark is additionally messy because there are two packages on PyPI that
    # provide a "lark" module.
    import lark   # ImportError if no such module
-   lark.Visitor  # AttributeError if wrong module
+   lark.Visitor # AttributeError if wrong module
+   lark_path = os.path.abspath(lark.__file__)
+   desired_path = os.path.dirname(os.path.abspath(__file__)) + "/lark_parser.egg/lark/__init__.py"
+   if lark_path != desired_path:
+       sys.exit("Lark found was not embedded")
```
j-ogas, line 43:
```suggestion
       sys.exit("lark-parser found was not embedded")
```

j-ogas: We should maybe refer to it as 'lark-parser' to be consistent with other warnings and errors.

lib/charliecloud.py
```diff
@@ -49,14 +56,14 @@
    import requests
    import requests.auth
    import requests.exceptions
-except ImportError:
+except ImportError as e:
+   print(e)
```
j-ogas, line 60: Is this for debugging or intentional?

packaging/fedora/charliecloud.spec
```diff
@@ -123,6 +123,7 @@ ln -s "${sphinxdir}/js"    %{buildroot}%{_pkgdocdir}/html/_static/js
 %{_libdir}/%{name}/*.sh
 %{_libdir}/%{name}/contributors.bash
 %{_libdir}/%{name}/version.txt
+%{_libdir}/%{name}/lark_parser.egg
```
j-ogas, line 126: Is this egg file needed at all times? Or can we get rid of it after the `configure` and `make` steps of our build process?

j-ogas (changes requested):

doc/dev.rst
shanegoff: Tagged releases are specific commits. But I'll see if I can be more clear without too much text.

packaging/fedora/charliecloud.spec
shanegoff: We are forcing users to use the embedded lark-parser so the egg file must be present in order for ch-image to work.

lib/charliecloud.py
shanegoff: That was for debugging good catch.

lib/charliecloud.py
```diff
@@ -33,7 +36,11 @@
    # Lark is additionally messy because there are two packages on PyPI that
    # provide a "lark" module.
    import lark   # ImportError if no such module
-   lark.Visitor  # AttributeError if wrong module
+   lark.Visitor # AttributeError if wrong module
+   lark_path = os.path.abspath(lark.__file__)
+   desired_path = os.path.dirname(os.path.abspath(__file__)) + "/lark_parser.egg/lark/__init__.py"
+   if lark_path != desired_path:
+       sys.exit("lark-parser found was not embedded")
```
reidpr, line 43: Possibly:

```suggestion
   if (lark_path != desired_path):
      depfails.append("wrong version", 'Python module "lark-parser" not embedded')
```

lib/charliecloud.py
```diff
@@ -49,14 +56,14 @@
...
-except ImportError:
+except ImportError as e:
+   print(e)
    depfails.append(("missing", 'Python module "requests"'))
    # Mock up a requests.auth module so the rest of the file parses.
    requests = types.ModuleType("requests")
    requests.auth = types.ModuleType("requests.auth")
    requests.auth.AuthBase = object
 
-
```
reidpr, line 59: Typo?

.github/workflows/main.yml
```diff
@@ -165,12 +165,8 @@ jobs:
       - name: late setup & validation, ch-image
         if: ${{ matrix.builder == 'ch-image' }}
         run: |
-          [[ $(bin/ch-image python-path) = /usr/local/bin/python3 ]]
-
-      - name: late setup & validation, all
-        run: |
-          bin/ch-test --is-pedantic all
-          bin/ch-test --is-sudo all
```
reidpr, line 173: Why remove these lines?

.github/workflows/main.yml
```diff
@@ -165,12 +165,8 @@ jobs:
...
+          bin/ch-test -b ${{ matrix.builder }} --is-pedantic all
```
reidpr, line 168: `-b` should be covered by environment variable.

.github/workflows/main.yml
```diff
@@ -269,9 +265,7 @@ jobs:
           set -x
           fgrep 'HTML documentation ... no' config.log
           fgrep 'man pages ... no' config.log
-          fgrep 'ch-image(1) ... no' config.log
           fgrep 'with Docker: yes' config.log
-          fgrep 'with ch-image(1): no' config.log
```
reidpr, line 274: Why were these lines removed?

reidpr:
Sorry, clicked too soon. This looks real nice. I think this will work; we just need to figure out how to “install” Lark.

configure.ac
```diff
@@ -41,7 +41,7 @@ AS_CASE([$host_os],
 # [3]: https://lists.gnu.org/archive/html/automake/2004-01/msg00013.html
 MKDIR_P=${MKDIR_P:-install -d -m 0755}
 
-AM_INIT_AUTOMAKE([1.13 -Wall -Werror foreign subdir-objects])
+AM_INIT_AUTOMAKE([1.13 -Wall -Werror tar-ustar foreign subdir-objects])
```
reidpr, line 44: IMO it would be good if we could stay with the default tar format. If we install into say `lib/embed` instead of e.g. `lib/lib/python3.7/site-packages` will that work?

shanegoff: I tried using a slightly shorter path name but it breaks even if you use `lib/modules` by a long shot on some files. I can double check if that would work with the no compile flag, as that may shorten some paths.

.github/workflows/main.yml
shanegoff: These were removed early on in the dream of bundling requests and lark. With no dependencies to install it would cause CI to fail because `ch-image` would always be "yes". I think I can roll them back though.

.github/workflows/main.yml
shanegoff: I am not sure if I removed these for some reason, but it looks like I removed lines I didn't intend to.

loveshack:
Please don't insist on using bundled dependencies. That will cause grief for packagers having to undo it (at least for Fedora but, I think, Debian too). Packages depend on the system version, so all's well at runtime.
It's OK to bundle stuff, of course, as long as it's straightforward to avoid it.

reidpr:
@loveshack, a key goal here is that EPEL has a quite old version of lark-parser (0.7.1). Current (0.11.2) has quite a few new features and bug fixes we want to take advantage of. We already have workarounds for bugs in 0.7.1.

I'm disinclined to hold Charliecloud back for a distro that's not up to date on a key library. If EPEL updates its lark-parser to something current, I'm happy to consider that dependency.

reidpr (commented):
After my commit, `ch-image --dependencies` works in Git checkout and after `make install` from both Git checkout and tarball. I didn't do any further testing.

lib/charliecloud.py
```diff
@@ -22,29 +22,25 @@
 
 ## Imports not in standard library ##
 
-# These are messy because we need --version and --help even if a dependency is
-# missing. Among other things, nothing can depend on non-standard modules at
-# parse time.
+sys.path.insert(0, (  os.path.dirname(os.path.abspath(__file__))))
```
reidpr, line 25: I think maybe the executables already do this?

shanegoff: This was actually already here before I changed it to something else then back to that.

shanegoff: I will double check my facts on the above, but I think it needs to be here.

shanegoff:
This is ready for final review. Everything appears to be in working order.

reidpr:
Also addresses #1055.

reidpr (approved):
OK, I think this is ready to go. Thanks for your persistence. LMK if you agree and I'll merge.

shanegoff:
Ready to go

Issue #730, opened 2020-04-30
ch-run: make --join the default in certain scenarios


heasterday:
Within the team we have been discussing making `--join` the default behavior when certain hints in the environment are discovered. This is driven by the belief that the majority of HPC workloads would be run this way and having this done for the user is convenient.

I envision this having a switch to disable this behavior and an accompanying message saying what is happening and why, with an option to suppress.

Some initial things that need investigation:
- Are variables in the environment sufficient hints to engage this behavior? We might look for env variables differently when in a slurm allocation vs direct launched via mpiexec
- Can we check if ch-run was direct launched with a parallel launcher?
- Should we be checking the parent process for ch-run?
- What would happen when more processed tried to join a namespace that had been previously instantiated

Please chime in if you feel this would be a positive/negative change, the more insights the better.

reidpr:
One specific scenario we discussed is, can we tell the difference between launched by `srun` (`--join` is probably wanted) or just directly in a Slurm allocation (probably not wanted).

heasterday:
In the meeting today we decided to move forward creating a list of env variables for various launchers to check e.g. PMI_SIZE > 1 and take that approach.

samuelkgutierrez:
@heasterday and I spoke offline today. He suggested that I document some of our findings.

* `salloc` SLURM environment variables change based on how allocations are requested, so we cannot solely rely on presence or absence of `SLURM_PROCID` and `SLURM_NPROCS` to inform us about automatic `--join` behavior.
*  Reliance on certain `PMI*` variables can suffer from a similar problem.
* `OMPI_*` variables are reliable, but require a particular MPI implementation's parallel process launcher. So the use of these variables as hints may be necessary, but are not sufficient.

The more I understand Hunter's proposed solution to this, the more I like it. It provides additional information that allows `ch-run` to make better guesses regarding `--join`. The only complication then is to nicely handle unknown, valid parent processes necessitating the use of `--join` at a user level. Providing a well-known environment variable to influence this behavior is one option.

reidpr:
Let me summarize the situation as I see it.

We want to support both the parallel job (e.g., `srun ch-run`) and "just running `ch-run`" use cases in a Slurm allocation, even when running on a compute node.

**Current behavior**

* `ch-run` (default) does not join.
* `ch-run --join` tries to join. It needs a join tag and a join count (number of peers to join). If running under Slurm, it gets these from environment variables; if not, they must be provided with `--join-tag` and `--join-ct`.

[...]

PR #974, opened 2021-01-27
bin/ch-run make --join the default in certain situations

heasterday:
Addresses #730 

Please note this is a very rough draft intended to codify this approach (checking parent process names), it needs a major `tidy`.


reidpr (commented):
It's certainly ... interesting, per our discussion. I'm not sure whether it's brilliant or ludicrous. 😉 Why don't we discuss it next meeting?

heasterday:
We decided to take the approach of checking env variables instead.

Issue #978, opened 2021-02-08
examples/spark: incorrect worker count and hang on xc40


j-ogas:
The spark example is having issues on xc40. The worker count test fails reporting an incorrect count and the subsequent pi test hangs indefinitely.

```
 ✗ spark/worker count
   (in test file /tmp/ch-test.tmp.jogas/spark_example.bats, line 107)
     `[[ $worker_ct -eq "$SLURM_NNODES" ]]' failed
   node count: 2; worker count: 0
 ^Cspark/pi
 ```

PR #986, opened 2021-02-11
examples/spark: resolve failures on Cray XC40

heasterday:
Addresses #978

We can get fancier for setting JAVA_HOME but this location appears to be consistent on Debian.

Marking as draft pending more tests.

### DIFF
--- a/examples/spark/test.bats
+++ b/examples/spark/test.bats
@@ -31,13 +31,7 @@ setup () {
     spark_config=$spark_dir
     spark_log=/tmp/sparklog
     if [[ $ch_multinode ]]; then
-        # Use the last non-loopback IP address. This is a barely educated
-        # guess and shouldn't be relied on for real code, but hopefully it
-        # works for testing.
-        master_ip=$(  ip -o -f inet addr show \
-                    | grep -F 'scope global' \
-                    | tail -1 \
-                    | sed -r 's/^.+inet ([0-9.]+).+/\1/')
+        master_host="$(hostname)"
         # Start Spark workers using pdsh. We would really prefer to do this
         # using srun, but that doesn't work; see issue #230.
         command -v pdsh >/dev/null 2>&1 || pedantic_fail "pdsh not in path"
@@ -46,7 +40,7 @@ setup () {
         master_ip=127.0.0.1
         pernode=
     fi
-    master_url="spark://${master_ip}:7077"
+    master_url="spark://${master_host}:7077"
     master_log="${spark_log}/*master.Master*.out"
 }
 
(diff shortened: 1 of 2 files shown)


heasterday:
Passes tests on local dev machine, xc40, and CTS-1.

examples/spark/test.bats
```diff
@@ -31,22 +31,16 @@ setup () {
...
     spark_log=/tmp/sparklog
     if [[ $ch_multinode ]]; then
-        # Use the last non-loopback IP address. This is a barely educated
-        # guess and shouldn't be relied on for real code, but hopefully it
-        # works for testing.
-        master_ip=$(  ip -o -f inet addr show \
-                    | grep -F 'scope global' \
-                    | tail -1 \
-                    | sed -r 's/^.+inet ([0-9.]+).+/\1/')
+        master_host="$(hostname)"
```
reidpr, line 34: Seems like a comment about what network we are selecting here, and why it's good enough for testing but probably not production, could be useful?

examples/spark/test.bats
```diff
@@ -63,7 +57,8 @@ SPARK_LOCAL_DIRS=/tmp/spark
 SPARK_LOG_DIR=$spark_log
 SPARK_WORKER_DIR=/tmp/spark
 SPARK_LOCAL_IP=127.0.0.1
-SPARK_MASTER_HOST=${master_ip}
+SPARK_MASTER_HOST=${master_host}
+JAVA_HOME=/usr/lib/jvm/default-java/
```
reidpr, line 61: Does it work to unset `JAVA_HOME`? I thought Java was smart about that if the variable wasn't already set. I.e., is the problem bogus `JAVA_HOME` leaking in from somewhere?

reidpr (commented):
Could questions. Thanks for dealing with this!

examples/spark/test.bats
heasterday: I agree, I will add this.

examples/spark/test.bats
heasterday: I initially thought `JAVA_HOME` was getting incorrectly set somewhere and bypassing our `unset` as we have had issues with this in the past but that doesn't appear to be the case here. There were a few examples online of adding this `JAVA_HOME` line to the spark env file for situations where it isn't getting picked up so I took this approach.

reidpr: Sounds good. Maybe add a comment to that effect, to head off ignorant commenters like me?

heasterday:
Incorporated feedback, please re-review.

j-ogas (approved):

PR #989, opened 2021-02-17
update tests for new Lark

reidpr:
Lark 0.11.2 broke test suite. To fix it, I removed one of the output checks in the offending test. I could have made that check more fancy, but it seemed like the prior check probably covered it sufficiently.


reidpr:
Optimistically marking `ready to merge?`.

heasterday (approved):

PR #996, opened 2021-03-04
ch-image push: fix Docker Hub

reidpr:
Container registry API can return either HTTP 401 or 404 to say something is not found. We were only accepting 404.

Also addresses #985.


reidpr:
This is ready for review. Please test using the internal HOWTO.

j-ogas (approved):

PR #1012, opened 2021-03-22
packing/fedora: update spec file to 0.24

j-ogas:
Add the lark and request sauce along with various improvements.


reidpr (commented):
Doesn't this need to wait until the Lark stuff is finalized?

loveshack:
If you want to include the spec file, why can't it just be the current Fedora one?  (I'll suggest some surgery on that, though, which might be worth looking at before packaging 0.23.)

j-ogas:
Here are my proposed updates to the charliecloud spec file. Note that the `charliecloud.spec` here is more up-to-date than the most current build on Fedora (represented by `upstream.spec`). Thus, whatever changes we agree on here will need to pushed to https://src.fedoraproject.org/rpms/charliecloud. FWIW, I have submit scratch-builds for this update and they pass on f35, epel8, and epel7.

Some of the more notable changes:

1.  new `-builder` sub-package that encapsulates all of the `ch-image` dependencies (python, python-modules, etc.)
2. provides and obsoletes line for `charliecloud-runtime` and `charliecloud-common`*
3. new primary packages dependencies: `squashfuse`, `squashfs-tools`, and `bash`*
4. `%files` section explicitly lists what each package is providing

Re. `charliecloud-runtime` and `charliecloud-common`: @loveshack and I originally broke the package up into the sub-packages: -runtime, -common, and -builders. This change, `0.24-4`, was staged and made it to stable on f35. After some further consideration I decided to unpush and propose the changes here. The `Provides` and `Obsoletes` here attempt to ensure any such sub-packages on f35/34 get updated.

Re. `bash`: some of the generic helper scripts are bash, e.g., `ch-builder2dir`. We could move these bash specific helper scripts to `-build` and remove the `bash` dependency.

packaging/fedora/build
```diff
@@ -20,6 +20,7 @@ import time
 CH_BASE = os.path.abspath(os.path.dirname(__file__) + "/../..")
 CH_RUN  = [CH_BASE + "/bin/ch-run"]
 PACKAGES = ["charliecloud",
+            "charliecloud-builder", # depends on -doc
```
reidpr, line 23: Why does it depend on `-doc`? If it's some Python thing I wonder if there is a better way to do it.

packaging/fedora/charliecloud.spec
```diff
@@ -16,11 +16,14 @@ Summary:       Lightweight user-defined software stacks for high-performance com
...
-Patch0:        lib64.patch
-%if 0%{?el7}
+BuildRequires: gcc rsync bash
+Requires:      squashfuse squashfs-tools
 Patch1:        el7-pkgdir.patch
-%endif
+# Suggests:    name-builder docker buildah
+Provides:      %{name}-runtime
+Obsoletes:     %{name}-runtime < 0.24-9
+Provides:      %{name}-common
```
reidpr, line 25: Can we have just obsoletes? The `charliecloud-common` package only existed for such a brief while; seems a shame to carry around this forever. I'm also skeptical it's something we should "provide" — people should not be depending on it.

packaging/fedora/charliecloud.spec
```diff
@@ -31,37 +34,48 @@ the performance and functionality already on offer.
...
+BuildArch:     noarch
+BuildRequires: python3-devel
+BuildRequires: python%{python3_pkgversion}-lark-parser
+BuildRequires: python%{python3_pkgversion}-requests
+Requires:      %{name}
+Requires:      python3
+Requires:      python%{python3_pkgversion}-lark-parser
+Requires:      python%{python3_pkgversion}-requests
+Provides:      %{name}-builders
+Obsoletes:     %{name}-builders
```
reidpr, line 50: Same here; we shouldn't be providing a typo forever.

packaging/fedora/charliecloud.spec
```diff
@@ -31,37 +34,48 @@ the performance and functionality already on offer.
...
+
+%description builder
+This package provides tools to build Charliecloud container images.
```
reidpr, line 53: Advertise our unprivileged builder a bit more strongly IMO.

packaging/fedora/charliecloud.spec
```diff
@@ -115,36 +130,64 @@ ln -s "${sphinxdir}/js"    %{buildroot}%{_pkgdocdir}/html/_static/js
...
+%{_bindir}/ch-fromhost
+%{_bindir}/ch-mount
+%{_bindir}/ch-pull2dir
+%{_bindir}/ch-pull2tar
+%{_bindir}/ch-run
+%{_bindir}/ch-run-oci
+%{_bindir}/ch-ssh
+%{_bindir}/ch-umount
+%{_bindir}/ch-tar2dir
+%{_mandir}/man1/ch-build.1*
```
reidpr, line 148: Why a wildcard?

test/run/build-rpms.bats
```diff
@@ -95,22 +104,30 @@ setup () {
...
-    [[ $output = *'/usr/share/doc/charliecloud/examples/lammps/Dockerfile'* ]]
+    [[ $output = *'/usr/lib/charliecloud/base.sh'* ]]
     [[ $output = *'/usr/share/man/man7/charliecloud.7.gz'* ]]
+    run ch-run "$img" -- rpm -ql "charliecloud-builder"
+    echo "$output"
+    [[ $status -eq 0 ]]
+    [[ $output = *'/usr/bin/ch-image'* ]]
+    [[ $output = *'/usr/lib/charliecloud/charliecloud.py'* ]]
+    [[ $output = *'/usr/lib/charliecloud/pull.py'* ]]
+    [[ $output = *'/usr/lib/charliecloud/push.py'* ]]
```
reidpr, line 122: Could probably get away with just one of these `.py` files.

reidpr (commented):
A few comments in-line. Overall seems good but I'm not the right person to catch problems with RPMs, so I'll comment instead of approve or request changes.

packaging/fedora/build
j-ogas: It doesn't, this is left over crud. Good catch.

packaging/fedora/charliecloud.spec
j-ogas: Good questions and comments.

Re. obsoletes: we _could_ just have obsoletes but then we'd need to add a new `rpmlint` ignore entry since rpmbuild will complain that we aren't specifying versions. FWIW I'd like for these obsoletes lines to disappear completely in 0.26 since, as you pointed out, it only existed for a day or two on f35/34. So I didn't mind adding the versions with their imminent removal in the near future in mind. Thoughts?

reidpr (approved):
LGTM though again I'm don't know RPM packaging well at all.

j-ogas:
Builds pass on F35, F34, F33, epel7, and epel8. Merge at will.

Issue #1017, opened 2021-03-29
ch-image: add architecture support


reidpr:
`ch-image` needs to be smart enough to deal with architectures, on pull (and build) definitely, ~possibly also push~ [This issue is build/pull only.]

PR #1024, opened 2021-04-08
ch-image: add arch support

j-ogas:
Addresses #1017.


j-ogas:
This first commit is `ch-image` and `ch-builder2tar` documentation for an implementation loosely based around the design of the original PR #865.
 
 The assumptions are as follows:
- storage has some form of top-level architecture layout, e.g., `/var/tmp/ch-image/${USER}/${ARCH}`, or `/var/tmp/ch-image/${USER}/img/${ARCH}` (latter assumes that dlcache contents can be shared)
- image names/tags are not unique, i.e. multiple images named `centos:7` can exist in storage with different architectures
- architecture guesses use`uname -m` 
  - `amd64` implies `x86_64`
  - `arm64v8` implies `aarch64`
  - `armv7l` implies `aarch32`
  - `armv5` implies `armv5l`
  - `armv6` implies `armv6l`
  - `386` implies `386`
  - `mips64le` implies `mips64le`
  - `ppc64le` implies `ppc64le`
  - `s390x` implies `s390x`
  
Updating `ch-image` is straight forward enough, however, I do see potential issues with other scripts, e.g., `ch-builder2*`, and `ch-test`. I think the current documentation update for `ch-builder2tar` in this PR highlights this:
 
 ```
 If multiple images of same tag :code:`IMAGE` exist in storage with different
architectures, prefer the one that most closely resembles the host. If the
host's architecture doesn't match any image in storage then use the following
defaults in order of precedence: 1) image tagged :code:`IMAGE` with architecture
:code:`arch-unspecified`; 2) first image matching image tagged :code:`IMAGE`
(architectures are searched in alphabetical order).
```

Obviously we have options that can simplify this. For example we could make `ch-builder2*` require an `--arch` argument if multiple images with the same name exist in storage. However, it _feels_ off.

I think it would be a lot simpler if we made image names/tags unique. This would remove the need for a top-level architecture directory in storage and as a result require much less guessing/magic for the other helper scripts.

**Comment by j-ogas** [comment]
@qwofford re. emoji: our workflow is to open draft PRs starting with documentation that propose changes. The "addresses #NUMBER" line is used to establish a relationship between PR and target issue; it isn't to be taken literally until the PR is merged.

**Comment by j-ogas** [comment]
This second commit changes a key assumption: image references are now unique. This removes the need to alter the storage the directory and our helper scripts, e.g., `ch-builder2*`, `ch-test`, etc.

File: doc/ch-image_desc.rst [hunk 1]
```diff
@@ -8,6 +8,7 @@ Synopsis
    $ ch-image [...] list
    $ ch-image [...] pull [...] IMAGE_REF [IMAGE_DIR]
    $ ch-image [...] push [--image DIR] IMAGE_REF [DEST_REF]
+   $ ch-image [...] rename IMAGE_REF NEW_IMAGE_REF
```
reidpr, line 11: Shouldn't `rename` be a separate PR?

doc/ch-image_desc.rst
```diff
@@ -47,7 +48,6 @@ Common options placed before the sub-command:
   :code:`-v`, :code:`--verbose`
     Print extra chatter; can be repeated.
 
-
```
reidpr, line 50: Good catch.

doc/ch-image_desc.rst
```diff
@@ -231,6 +231,15 @@ Destination:
 
 Options:
 
+  :code:`--arch`
+    Specify architecture :code:`--arch=ARCH[/VARIANT]` for image described by
+    :code:`IMAGE_REF`. If found pull it; otherwise error. Use
+    :code:`--list-arch IMAGE_REF` for list of valid platform architectures.
```
reidpr, line 237: I think it would be good, if we can, to avoid separating the architectures into variants, and keep it at a single level, i.e., some architectures happen to include a slash.

```suggestion
  :code:`--arch`
    Specify architecture :code:`--arch=ARCH` for image described by
    :code:`IMAGE_REF`. If image exists but not as architecture
    :code:`ARCH`, fail with error.
```

doc/ch-image_desc.rst
```diff
@@ -231,6 +231,15 @@ Destination:
...
+
+  :code:`--list-arch`
+    Print a list of valid :code:`--arch` targets for image tagged
+    :code:`IMAGE_REF`.
```
reidpr, line 241:
```suggestion
  :code:`--list-arch`
    Print list of available architectures for image
    :code:`IMAGE_REF`.
```

doc/ch-image_desc.rst
```diff
@@ -239,6 +248,13 @@ Options:
     Parse :code:`IMAGE_REF`, print a parse report, and exit successfully
     without talking to the internet or touching the storage directory.
 
+This script tries to target the appropriate architecture when pulling. For
+example, when pulling the image tagged :code:`centos:7`, first check for the
+existence of a fat manifest. If found, check for the architecture matching the
```
reidpr, line 253: Can we avoid mentioning manifests and fat manifests?

doc/ch-image_desc.rst
```diff
@@ -239,6 +248,13 @@ Options:
...
+host. If a match is found, pull it; otherwise proceed as if the fat manifest
+does not exist. If a fat manifest does not exist, pull the image without trying
+to guess the appropriate architecture.
```
reidpr, line 256: E.g. I don't know what this means. What architecture do I get? This should explain what the results are, not what's happening under the hood.

doc/ch-image_desc.rst
```diff
@@ -534,6 +555,50 @@ Same, except place the image in :code:`/tmp/buster`::
...
+   amd64
+   arm/v5
+   arm/v7
+   arm64/v8
+   386
+   mips64le
+   ppc64le
+   s390x
+
+Specify the aarch64 version of Debian Stretch using :code:`--arch`.::
```
reidpr, line 572:
```suggestion
Specify the aarch64 version of Debian Stretch using :code:`--arch`::
```

doc/ch-image_desc.rst
```diff
@@ -534,6 +555,50 @@ Same, except place the image in :code:`/tmp/buster`::
...
+
+   $ ch-grow pull --arch=arm64/v8 debian:buster
+   pulling image:   debian:buster
+   fat manifest: using existing
+   architecture: arm64v8 (aarch64)
```
reidpr, line 577: What does it mean to have two architectures listed?

doc/ch-image_desc.rst
```diff
@@ -534,6 +555,50 @@ Same, except place the image in :code:`/tmp/buster`::
...
+   manifest: downloading
+   layer 1/1: 04aacb1: downloading
+   layer 1/1: 04aacb1: listing
+   validating tarball members
+   resolving whiteouts
+   flattening image
+   layer 1/1: 04aacb1: extracting
+   done
+
+Specify the aarch32 version of Debian Stretch using long name.::
```
reidpr, line 587: Would anyone want to do this?

doc/ch-image_desc.rst
```diff
@@ -598,5 +663,18 @@ in the remote registry, so we don't upload it again.)
    cleaning up
    done
 
+:code:`list`
```
reidpr, line 666: Put in alphabetical order.

doc/ch-image_desc.rst
```diff
@@ -598,5 +663,18 @@ in the remote registry, so we don't upload it again.)
...
+List images in storage.
+
+::
+
+  $ ch-image list
+  IMAGE REFERENCE            NAME      TAG        ARCH
+  arm64/v8/debian:stretch    debian    stretch    arm64v8 (aarch64)
+  debian:stretch             debian    stretch    amd64 (x86_64)
+  centos:8                   centos    8          ppcle64
+  centos:7                   centos    7          amd64 (x86_64)
```
reidpr, line 678: This is going to be hairy to compute the column widths, and because we don't currently distinguish between name and tag. Possibly:

```suggestion
  $ ch-image list
  arm64/v8/debian:stretch (aarch64)
  debian:stretch (x86_64)
  centos:8 (ppcle64)
  centos:7 (x86_64)
```

Also, why do we have one `debian:stretch` with an architecture path and one without?

doc/faq.rst
```diff
@@ -650,6 +650,30 @@ see image references like:
...
+  fat manifest: downloading
+  architecture: ppcle64
+  manifest: downloading
+  [...]
+
+Note that not all images have a fat manifest. In this case there is only one
+version of the image described by :code:`IMAGE_REF`, and charliecloud will
+download that regardless of what architecture it is. To see a list of supported
+architectures an image has, see :code:`ch-image list`.
+
```
reidpr, line 676: I think probably this FAQ can be eliminated with tweaks to the other docs.

reidpr (commented):
I like it. Several comments in-line. Also:

1. Would be good to have `--arch=default` that is the current behavior.
2. I see three stages for this work. Suggest limit this PR to the first stage.
   1. Request arch with `--arch`. Default is `--arch=default`. Only one image with a given reference at a time; i.e., if you do `pull --arch=foo` and then `pull --arch=bar`, the image with arch `foo` is overwritten with a new image of arch `bar`.
   2. Guess current architecture. Default is `--arch=FOO` where `FOO` is selected intelligently somehow to match the current machine.
   3. Multiple images with the same reference and different architectures simultaneously. (Do we really want to do this?)
3. Download cache sharing: Blob are not problem since the hashes will differ, but currently we name manifests after the image reference. We don't want to pick up a cached manifest with the wrong architecture.
4. `--arch` seems like a common option since `build` can also pull?
5. Maybe `--list-arch` should become subcommand `list-arch`? As written we have a `pull --list-arch` that doesn't pull, it lists. It would show both what was have in storage as well as available architectures. Alternately, the existing `list` command could grow an argument `IMAGE_REF` with this behavior.

doc/ch-image_desc.rst
j-ogas: Ah, good question. This shows us both what Docker calls the architecture and what something like `uname -m` reports. Since they can differ, e.g., above and `amd64 == x86_64`, I thought it may be helpful if both were included in the output.

doc/ch-image_desc.rst
j-ogas: Do what exactly? Pull a aarch32 image or pull something different than the host they're on?

doc/ch-image_desc.rst
reidpr: Aha. I think usability wise, we ought to stay in Docker’s vocabulary for these. Later when we’re guessing architectures we can show the translation, but here I think showing two different archs is confusing, it means we have to make the map available more broadly in the code, and going both directions will add complexity if it turns out it’s not 1-to-1.

doc/ch-image_desc.rst
j-ogas: The image reference is unique. In these examples the image with the arch path was pulled via long-name, e.g., `ch-image pull arm64/v8/debian:stretch` thus exists in storage as `arm64%v8d%ebian:stretch`. 

This also illustrates a key assumption: that architecture guessing is only for short-names, e.g., `debian:stretch`. If you specify a long-name you know what you're doing and what you're asking for. For example, it doesn't make much sense for me to do `ch-image pull ppc64le/centos:8` and get a `x86_64` version.

doc/ch-image_desc.rst
j-ogas: The other examples in this file have a period. Should I remove them as well?

doc/ch-image_desc.rst
reidpr: Aha. I don't follow the distinction between short and long names. Seems too magic to me.

reidpr: E.g. what about `foo/bar/debian:stretch`? It's a valid image name but how do we know `foo/bar` isn't an architecture?

j-ogas:
The most recent commits implement a functional prototype.

Notes:
- guess the host architecture by default; disable this behavior by specifying `--arch=default`
- multiple images with the same reference and different architectures cannot exist in storage simultaneously
- images are overwritten each time a new architecture variant is downloaded
- `--arch` is common option
- `list-arch` is a new common function to list the available architectures of an image

reidpr (commented):
This looks nice! Various suggestions in-line.

FWIW, I realized that single-image-per-reference means that the shell scripts can be architecture-agnostic.

lib/charliecloud.py
```diff
@@ -1187,6 +1248,9 @@ def root_env():
    def manifest_for_download(self, image_ref):
       return self.download_cache // ("%s.manifest.json" % image_ref.for_path)
```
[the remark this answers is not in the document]
j-ogas: I don't think so. The only time we get a different manifest is when we use the fat manifest to pull the target image's manifest by digest. We compare the hash of this manifest with that in storage, and if they don't match we overwrite the storage contents for that particular image.

j-ogas:
This next commit addresses feedback.

We now have `--arch=none` (default), `--arch=DOCKER_REF`, and `--arch=host` with behavior that matches the quote below.

>    If --arch=none: No arch checks, no warnings.
>   Otherwise:
>        If fat manifest and desired arch is not present, error w/o further pulling.
>        If arch in config does not match desired arch, error w/o further pulling.
>        If no arch in config, warn that architecture is unknown.

Notes:
- removed the paragraph under the pull section that described guessing. Now that this behavior isn't default I wonder if it should just be in FAQ, e.g., `how do I pull an image that matches my host architecture?`.
- architecture handling is now done as part of the cli parsing (simpler)
- cleaned removed some redundant functions in favor for existing ones
- moved architecture listing behavior to `list [IMAGE_REF]`

doc/ch-image_desc.rst
```diff
@@ -509,22 +518,55 @@ installing into a layer::
...
-  layer 1/1: d6ff36c: downloading
-  layer 1/1: d6ff36c: listing
-  validating tarball members
-  resolving whiteouts
-  flattening image
-  layer 1/1: d6ff36c: extracting
-  done
+   $ ch-image pull debian:buster
+   pulling image:   debian:buster
+   architecture:    amd64
```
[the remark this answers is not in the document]
reidpr: Looking at the current code, I think it qualifies as hard, since we'd have to carry around the source to a lot of placed. IMO we should omit it.

lib/charliecloud.py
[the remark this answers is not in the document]
reidpr: Aha. In that case we should save manifests with their hash in the filename, like blobs. This will let us know if the file in cache is what we want without reading its content.

trandles-lanl:
This may be of help @j-ogas  https://github.com/archspec/archspec

reidpr (approved):
I pushed a commit containing a straw man simplification of the approach. This:

1. twiddles the documentation to be more precise
2. moves special `--arch=host` even further into command line processing, hopefully simplifying the situation

I did not test this proposal and it's not complete.

Notes:

1. `--arch=yolo` and `--arch=amd64` are still special; see the proposed docs.
2. `ch-image build` should probably warn if the base image's architecture seems to not match the host.
3. Pull should warn if the metadata arch (in config) doesn't match the arch requested.

reidpr (commented):
Sorry, mis-clicked "approve" rather than "comment".

reidpr (commented):
Great work! I did some refactoring/tidying.

Misc comments on my changes:

* Output and logging chatter are now different colors; I changed `ch.INFO` to yellow. Is this a good idea? Did I mess anything up?
* Detailed architecture information like Haswell vs. Broadwell likely needs the archspec library, and we're on our own for how to deal with it in registries. They had a nice paper in CANOPIE 2020.

Work remaining:

* Update examples to match actual output.
* We ignore the `mediaType` of skinny manifests listed in the fat manifest. Is this bad?
* Local registry push currently fails. I didn't debug this but it's logged in the latest CI fun.
* Pass manual registry tests.
* Add tests. I can think of:
  * `ch-image list IMAGE_REF`
	* image in builder storage
	* not in builder storage
	* exists remotely, fat manifest
	* exists remotely, fat manifest, requested arch unavailable
	* exists remotely, no fat manifest
	* exists remotely, no Linux architectures
	* does not exist remotely
	* `IMAGE_REF` syntax error
  * `ch-image pull`
	* base image has fat manifest
      * `--arch=yolo`
	  * `--arch=host`
	  * `--arch=amd64` (arch with no variant)
	  * `--arch=arm64/v8` (arch has variant)
	* base image does not have fat manifest
	  * `--arch=yolo` (win)
	  * `--arch=amd64` (win)
	  * `--arch=host` (win on amd64 host, error otherwise)
	  * `--arch=arm64/v8` (error)
	* requested arch does not match image (not sure how to trigger this)

j-ogas:
New batch of commits.

Re. info being yellow: I like it. What happens if someone is using a terminal with a white background?

Re. detailed architecture info: followup issue created here: #1091.

Re. ignoring skinny `mediaType`: I think this is fine.

Re. manual registry tests: they pass, however, I did need to update the output regex for the docker tests.

Re. tests: all of the test cases you listed above have been added except for when `IMAGE_REF` has a syntax error. That is covered already elsewhere.

reidpr:
> Re. info being yellow: I like it. What happens if someone is using a terminal with a white background?

Issue #1093.

reidpr (approved):
Looks great. Pushed a tidy commit, including notably: Use registry architecture vocabulary for the default architecture when we push.

If you're happy with this and it passes CI, let's merge on Monday.

test/build/50_ch-image.bats
```diff
@@ -188,83 +188,52 @@ EOF
...
-    [[ $output = *'in local storage:    no'* ]]
-    [[ $output = *'remote arch-aware:   yes'* ]]
-    [[ $output = *'archs available:'* ]]
-    [[ $output = *'386 amd64 arm/v5 arm/v7 arm64/v8 mips64le ppc64le s390x'* ]]
-
-    # exists remotely, no fat manifest
-    run ch-image list charliecloud/metadata:2021-01-15 --no-cache
+    [[ $output = *'available remotely:'*'yes'* ]]
+    [[ $output = *'in local storage:'*'no'* ]]
+    [[ $output = *'remote arch-aware:'*'yes'* ]]
```
j-ogas, line 217: The reason I preserved the whitespace here is to avoid false positive results. When the ch-image list output is something like below:

```
available remotely:  no
in local storage:    yes
remote arch-aware:   no
```

Both the test cases `[[ $output = *'available remotely:'*'yes'* ]]` and `[[ $output = *'in local storage:'*'no'* ]]` will pass even though they shouldn't.

reidpr: Oh, good catch. I was thinking to avoid false negatives if the number of spaces changes. Options include: (1) revert, and deal with false negatives if/when they arise, (2) change to regular expression with variable number of spaces. Option 2 is rather messy because it's hard to write Bash regexes as a literal (see examples elsewhere; we universally set a variable `$re`. We could also go with Option 1 and then change it to Option 2 if it becomes a problem.

j-ogas: I prefer going with option one (1) for now. If (when) it becomes an issue we can explore option 2 thoroughly.

PR #1025, opened 2021-04-13
COPY: support list form

mphinney1100:
Addresses #784


lib/build.py
```diff
@@ -360,7 +360,14 @@ def __init__(self, *args):
          self.dst = paths[-1]
       else:
          assert (ch.tree_child(self.tree, "copy_list") is not None)
-         self.unsupported_yet_fatal("list form", 784)
+         paths = [variables_sub(i, env.env_build)
+                  for i in ch.tree_child_terminals(self.tree, "copy_list",
+                                                   "STRING_QUOTED")]
+         self.srcs = paths[:-1]
+         for i in range(len(self.srcs)):
+            self.srcs[i] = self.srcs[i][1:-1]
```
reidpr, line 368: 1. Comment what you are doing here.
2. It seems like removing the quotation marks should be the grammar's job. That's out of scope for this PR, but can you create an issue for it?

lib/build.py
```diff
@@ -360,7 +360,14 @@ def __init__(self, *args):
          self.dst = paths[-1]
       else:
          assert (ch.tree_child(self.tree, "copy_list") is not None)
-         self.unsupported_yet_fatal("list form", 784)
+         paths = [variables_sub(i, env.env_build)
```
reidpr, line 363: It seems this is *almost* the same as the shell form, except (1) different terminal name and (2) stripping quotation marks. That said, the code isn't very long so maybe the current approach is easiest to understand. Consider whether the existing organization is right, vs. finer-grained conditionals to deal with the two differences. The latter could be accomplished with subclasses: attribute with terminal name, function to strip quotes that's a no-op for shell form. Add a brief comment justifying your decision.

test/build/50_dockerfile.bats
```diff
@@ -629,6 +620,29 @@ EOF
...
+    [[ $CH_BUILDER = buildah* ]] && skip 'Buildah untested'
+
+    run ch-image build -t foo -f - . <<'EOF'
+FROM 00_tiny
+COPY ["fixtures/empty-file", "."]
+EOF
+    echo "$output"
+    [[ $status -eq 0 ]]
+    [[ $output = *"COPY ['fixtures/empty-file'] -> '.'"* ]]
+    [[ $output = *'grown in 2 instructions: foo'* ]]
```
reidpr, line 635: Also test the file made it in OK!

test/build/50_dockerfile.bats
```diff
@@ -629,6 +620,29 @@ EOF
...
+
+    # multiple sources
+    run ch-image build -t foo -f - . <<'EOF'
+FROM 00_tiny
+COPY ["Build.missing", "common.bash", "/etc/fstab/"]
```
reidpr, line 640: I would make this a successful test.

test/build/50_dockerfile.bats
```diff
@@ -868,3 +882,4 @@ EOF
             ;;
     esac
 }
+
```
reidpr, line 885: Stray blank line added to end.

reidpr (changes requested):
Good work! Some feedback in-line.

lib/build.py
```diff
@@ -356,11 +356,18 @@ def __init__(self, *args):
          paths = [variables_sub(i, env.env_build)
                   for i in ch.tree_child_terminals(self.tree, "copy_shell",
                                                    "WORD")]
-         self.srcs = paths[:-1]
-         self.dst = paths[-1]
-      else:
-         assert (ch.tree_child(self.tree, "copy_list") is not None)
-         self.unsupported_yet_fatal("list form", 784)
+      elif (ch.tree_child(self.tree, "copy_list") is not None):
```
reidpr, line 359: Because this conditional is selecting between two options, a plain `elif` isn't really appropriate — what happens if neither is true? There are a couple ways to deal with this that I can think of: (1) the prior `else`/`assert` structure, or `if`/`elif`/`else` with the `else` asserting about unreachable code reached.

lib/build.py
```diff
@@ -356,11 +356,18 @@ def __init__(self, *args):
...
+         paths = [variables_sub(i, env.env_build)
+                  for i in ch.tree_child_terminals(self.tree, "copy_list",
+                                                   "STRING_QUOTED")]
```
reidpr, line 362: Can the quote stripping go here? You're already in an appropriate conditional.

test/build/50_dockerfile.bats
```diff
@@ -740,6 +731,36 @@ EOF
...
+    # single source
+    run ch-image build -t foo -f - . <<'EOF'
+FROM 00_tiny
+COPY ["fixtures/empty-file", "."]
+EOF
+    echo "$output"
+    [[ $status -eq 0 ]]
+    [[ $output = *"COPY ['fixtures/empty-file'] -> '.'"* ]]
+
```
reidpr, line 746: rm this blank line

test/build/50_dockerfile.bats
```diff
@@ -740,6 +731,36 @@ EOF
...
+    [[ $output = *"COPY ['fixtures/empty-file'] -> '.'"* ]]
+    
+    test -f "$CH_IMAGE_STORAGE"/img/foo/empty-file
+
+    # multiple source
+    run ch-image build -t foo -f - . <<'EOF'
+FROM 00_tiny
+COPY ["fixtures/empty-file", "fixtures/README", "."]
+EOF
+
```
reidpr, line 754: this one too

test/build/50_dockerfile.bats
```diff
@@ -740,6 +731,36 @@ EOF
...
+    echo "$output"
+    [[ $status -eq 0 ]]
+    [[ $output = *"COPY ['fixtures/empty-file', 'fixtures/README'] -> '.'"* ]]
+
```
reidpr, line 758: and this one :)

reidpr (changes requested):
Good work! Looks very close.

lib/build.py
```diff
@@ -356,11 +356,17 @@ def __init__(self, *args):
...
+         for i in range(len(paths)):
+            paths[i] = paths[i][1:-1] #strip quotes
       else:
-         assert (ch.tree_child(self.tree, "copy_list") is not None)
-         self.unsupported_yet_fatal("list form", 784)
+         ch.FATAL("not valid form of COPY")
```
reidpr, line 366: Getting here is a logic error, so IMO crashing (with a traceback) is probably better than a user-facing fatal error.

```suggestion
         assert (False, "unreachable code reached")
```

reidpr (approved):
LGTM! One tidy; I'll merge once it passes tests.

Issue #176, opened 2018-06-07
ch-build: add metadata to Docker images


reidpr:
`ch-build` should add metadata with `--label` to the built Docker images to facilitate sharing. Specifically:

* who built it
* build timestamp (is this already in the standard metadata?)
* version string
* Git commit and whether dirty working directory
* Git branch

Metadata would be added by default with a command line option to not do it.

@mcpherson, anything else you'd like to see?

mcpherson:
I don't know if there is a way to have a pointer to the source `Dockerfile`.

reidpr:
This is somewhat more complicated than I expected. Here is a concrete proposal. I need more feedback before proceeding.

### Command line option

New option `ch-build --no-metadata`. If present, the metadata listed below will not be added. (This is the current behavior.)

### Metadata labels

By default, `ch-build` will add the following key-value pairs to the image metadata using `docker build --label foo=bar`.

* `charliecloud.build-user`: Username who built the image and FQDN of build host, e.g. `charlie@foobox.example.com`.
* `charliecloud.version`: Version of Charliecloud used to build it (output of `ch-build --version`).

### Issues

* If the metadata changes, the image hash will change even if the content doesn't.
* Best practices call for a `MAINTAINER` line in the Dockerfile which appears as the `Author` field in the inspect JSON. This may be a better contact than whoever built the image.

### Metadata considered and rejected:

* Build timestamp: this is already the `Created` field in the inspect JSON.
* Content of source Dockerfile: this can be recovered with the `Cmd` field and walking the `Parent` chain.
* Path to source Dockerfile: paths can be sensitive.
* Context directory path: paths can be sensitive.
* `ch-build` command line: users probably don't expect this to be published.

PR #1028, opened 2021-04-16
`ch-convert -i docker`: include metadata

### LINKED ISSUE #276: ch-docker2tar: put all Docker metadata into tarball
This would be so users can see more about where the tarball came from.

rstyd:
Closes #276. This PR adds metadata for docker and buildah using golang template strings and their respective `inspect` commands.


bin/ch-builder2tar
```diff
@@ -128,6 +128,32 @@ EOF
     tar rf "$tar" -b1 -P --xform="s|${temp}|ch/environment|" "$temp"
     rm "$temp"
 
+    echo "adding metadata"
+    temp=$(mktemp --tmpdir ch-builder2tar.XXXXXX)
+    buildah inspect --format='{{print "{\n"}}'\
+'{{printf "  \"arch\": %q,\n" .Docker.Architecture}}'\
```
reidpr, line 134: Why is this `printf` instead of `print` like the others?

bin/ch-builder2tar
```diff
@@ -128,6 +128,32 @@ EOF
     tar rf "$tar" -b1 -P --xform="s|${temp}|ch/environment|" "$temp"
     rm "$temp"
 
+    echo "adding metadata"
+    temp=$(mktemp --tmpdir ch-builder2tar.XXXXXX)
+    buildah inspect --format='{{print "{\n"}}'\
```
reidpr, line 133: Use indentation to clarify the code structure here. You could also put the format in a variable if that helps.

bin/ch-builder2tar
```diff
@@ -128,6 +128,32 @@ EOF
...
+'{{printf "  \"arch\": %q,\n" .Docker.Architecture}}'\
+'{{print "  \"cwd\": "}}{{if .Docker.Config.WorkingDir}}'\
+'{{.Docker.Config.WorkingDir}}{{else}}{{print "\"\""}}{{end}}{{print ",\n"}}'\
```
reidpr, line 136: Wouldn't it be better to omit the `cwd` key if `.Docker.Config.WorkingDir` is not set?

bin/ch-builder2tar
```diff
@@ -128,6 +128,32 @@ EOF
...
+'{{print "}"}}{{end}}{{print ",  \n"}}'\
+'{{print "  \"shell\": [\n    "}}'\
+'{{range $i, $v := .Docker.ContainerConfig.Cmd}}'\
+'{{if $i}}{{printf ",\n    " .}}{{end}}{{printf "%q" $v}}{{end}}'\
+'{{print "\n  ]\n"}}'\
+'{{print "  \"volumes\": "}}{{if .Docker.Config.Volumes}}{{print "["}}'\
+'{{range $key, $v := .Docker.Config.Volumes}}{{printf "\n    %q," $key}}'\
+'{{end}}{{print "\n  ]"}}{{else}}{{print "[]"}}{{end}}'\
+'{{print "\n}\n"}}' "$image" > "$temp"
+    # Fix last comma corner cases for maps with {} and []
```
reidpr, line 152: Clarify the comment to include (a) what corner cases are covered and (b) what the transformation is.

bin/ch-builder2tar
```diff
@@ -128,6 +128,32 @@ EOF
...
+    sed -i ':begin;$!N;s/,\n  }/\n  }/g;tbegin;P;D' "$temp"
```
reidpr, line 153: Use `sed -Ei` for extended regular expressions which are a lot better.

bin/ch-builder2tar
```diff
@@ -155,6 +181,28 @@ docker)
...
+'{{if .Config.Labels}}{{print "\n"}}'\
+'{{range $key, $v := .Config.Labels}}'\
+'{{printf "    \"%s\": \"%s\",\n" $key $v}}{{end}}{{print "  }"}}{{else}}'\
+'{{print "}"}}{{end}}{{print ",  \n"}}'\
+'{{print "  \"shell\": "}}{{json .ContainerConfig.Cmd}}{{print "\n"}}'\
+'{{print "  \"volumes\": "}}{{if .Config.Volumes}}{{print "["}}'\
+'{{range $key, $v := .Config.Volumes}}{{printf "\n    %q," $key}}'\
+'{{end}}{{print "\n  ]"}}{{else}}{{print "[]"}}{{end}}'\
+'{{print "\n}\n"}}' > "$temp"
+    sed -i ':begin;$!N;s/,\n  }/\n  }/g;tbegin;P;D' "$temp"
```
reidpr, line 203: Add a comment here. Both should be commented but one can refer to the other.

reidpr (commented):
Thanks for your hard work Rusty. This is a good start. Several comments in-line. General items:

1. This needs a test to verify that `metadata.json` is present and correct. Possibly this is pedantic mode if something like `jq(1)` is needed. Maybe `test/make-auto.d/unpack.bats`?
2. I focused on the Buildah one since it was first and they seem similar. That said, should we bother with supporting Buildah since it's experimental? If we add untested code it will break pretty fast.

bin/ch-builder2tar
rstyd: I used `printf` when printing a variable to get the formatting right and `print` otherwise.

rstyd: `should we bother with supporting Buildah since it's experimental?` That's a good question. I mostly added it because they worked quite similarly in that buildah replicates the metada docker provides, but we could remove the buildah one for just docker. What would like to do?

bin/ch-builder2tar
rstyd: I was imitating the way that our current metadata works in ch-image. If a parameter does not exist, we place the empty set of whatever data type it is `"", [], {}`.

rstyd: We can definitely remove this if it doesn't exist instead.

reidpr (changes requested):
Thanks Rusty!

I pushed a tidy, which will definitely make the tests fail (`shell` was picking up the image history, not `SHELL`).

Also, see if you can find a standard-scope image with more complex image. `00_tiny` is pretty basic and missing most of what you output.

bin/ch-builder2tar
rstyd: Not able to indent code.

reidpr (changes requested):
Good progress, though still not quite right.

I see we have a `test/Dockerfile.metadata` that “includes all the instructions that seemed like they ought to create metadata”. That smells like what you want. I also see a test “pull image with metadata” that is probably close to the test you want.

reidpr:
I think #176 is actually something different?

rstyd:
Ah good catch. It should be #1028.

reidpr:
> Ah good catch. It should be #1028.

Do you mean #276? #1028 is this PR.

bin/ch-convert
```diff
@@ -139,10 +140,48 @@ cv_docker_tar () {
...
+# ContainerConfig.Cmd has the last command that was run in the image
+# We can use that to extract the shell that was used to build it.
+# As far as I can tell, that is always /bin/sh as with ch-image
+# Feels kludgy. But it gets the right result.
+last_cmd=$(docker_ inspect "$1" --format '{{json .ContainerConfig.Cmd}}')
+if echo "$last_cmd" | grep -q "\"/bin/sh\"\,\"\-c"; then
+    printf "  \"shell\": [\n    \"/bin/sh\",\n    \"-c\"\n  ],\n" >> "$tmpmeta"
+else
+    printf "  \"shell\": [\n    \"unknown\"\n  ],\n" >> "$tmpmeta"
+fi
```
reidpr, line 168: I think there's a non-kludgy way to get this, i.e., a direct key that gives the shell if it's not the default.

test/run/ch-run_misc.bats
```diff
@@ -1009,3 +1009,30 @@ EOF
...
+    image="centos7"
+    diff -u - "$ch_imgdir/$image/ch/metadata.json" <<'EOF'
+{
+  "arch": "amd64",
+  "cwd": "",
+  "env": {
+    "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+  },
+  "labels": {
+    "org.label-schema.build-date": "20191001",
```
reidpr, line 1025: This date likely gives a brittle diff. Jordan taught me about `diff -I` which could be used to exclude the date from comparison.

test/run/ch-run_misc.bats
```diff
@@ -1009,3 +1009,30 @@ EOF
     [[ $output = *"UID ${uid_bad} not found; using dummy info"* ]]
     [[ $output = *"GID ${gid_bad} not found; using dummy info"* ]]
 }
+
+@test 'Docker metadata' {
+    scope standard
+    [[ $CH_BUILDER = docker ]] || skip 'Docker only'
+    image="centos7"
```
reidpr, line 1016: Would [`charliecloud/metadata`](https://hub.docker.com/r/charliecloud/metadata) work? That would give a more complete test.

reidpr (changes requested):
Thanks Rusty. Some suggestions in-line.

Generally: Would it make sense to include the metadata in the other three formats that `ch-convert` supports (`squash`, `dir`, `ch-image`)? And if so, is the scope reasonable to do it in this PR?

test/run/ch-run_misc.bats
rstyd: Ooh that's nifty. I'll check it out.

bin/ch-convert
rstyd: That'd be great. I spent a bunch of time looking through the json files of various images, and this was the only common piece I could find.

test/run/ch-run_misc.bats
rstyd: Yeah that sounds good!

reidpr:
This is getting stale. Re-open if/when appropriate.

Issue #1015, opened 2021-03-29
ch-run: deal with existing mount points that are symlinks correctly


loveshack:
I tried to run an image built from a linuxcontainers.org rootfs, where /etc/resolv.conf is a link to /run/systemd/resolve/resolv.conf, and the host system doesn't have that link. The touch in ch-tar2dir gives a (non-fatal) error and, more to the point, DNS doesn't work inside the container.  (This looks rather familiar from long ago, maybe not with Charliecloud.)

Shouldn't /etc/resolv.conf be deleted in the image and replaced with an empty file on which to bind mount?  If I modify the script to do that, all is well. That should perhaps be done for other mount points which are always used.

loveshack:
Apologies. I corrected where the touch gets done in the original version of that.

reidpr:
Thanks for the bug report! We'll fix it.

loveshack:
Sorry, I missed the request somehow.  I was actually using a tar image from under https://uk.images.linuxcontainers.org/images/fedora/33/ppc64el/default/ (which will vary with time) and using tar2dir.  I think that was before I found how to do ppc64le docker-ism. It would be the same with the x86_64 version. I was going to provide the patch I've been using in a bit.

PR #1034, opened 2021-04-19
Avoid problem when /etc/resolv.conf is a symlink

loveshack:
I've used this to address #1015.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,6 +58,7 @@ Contributors:
 * Shane Goff <rgoff@lanl.gov>
 * Michael Jennings <mej@lanl.gov>
 * Christoph Junghans <junghans@lanl.gov>
+* Dave Love <dave.love@manchester.ac.uk>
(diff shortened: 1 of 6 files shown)


reidpr:
Hello @loveshack, thanks for the contribution! You've identified a real and rather sneaky problem and provided the basis for a solid fix.

I pushed a couple more commits on this branch to extend the solution:

1. The same bug occurs in a couple other places; fix those too and tie things together with common code when possible.
2. Add testing for the bug.
3. Add you to the contributors list.

Does this PR still solve your problem?

shanegoff (approved):
Good fix. I didn't realize that symlinks would trip up the current code.

loveshack:
I had replied to this by mail, but at least some mail on github issues is going to a black hole. Prod me if there's other feedback you're missing.

Anyway, that fixes what I found, thanks.  (I don't know how I missed the case with squashfs when I looked for other instances.)

Issue #1030, opened 2021-04-16
ch-image build --force: support UBI8


qwofford:
A package requires the `--force` option. The package is in epel.

When I attempt to install the package without `--force`, I can install epel but I can't install the package.

When I attempt to `--force` the package installation, I can no longer install epel.

Dockerfile reproducer:
```
FROM qwofford/nvidia-cuda-devel:summit_2020-11-25
RUN dnf install -y  https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
RUN dnf update -y
```

reidpr:
The problem is that `ch-image` tries to use the `rhel8` config but the UBI8 image doesn't have a `epel-release` package:

```
+ dnf install -y epel-release
[...]
Error: Unable to find a match: epel-release
```

I note that the Dockefile above installs EPEL with an RPM, not a package name. This is the [recommendation from Fedora](https://fedoraproject.org/wiki/EPEL).

reidpr:
I think the thing to do here is:

1. Change the configs so we install EPEL via RPM. (Possible gotcha: What happens if the Dockerfile *also* installs it via RPM?)
2. Add a UBI8 test for `--force`.

Issue #1026, opened 2021-04-14
ch-image build --force: add Fedora support


reidpr:
@ipdemes requested `--force` for a [few more distributions](https://github.com/flecsi/flecsi/tree/images/docker/os):

* Clear Linux (`latest`, distro’s versioning seems unclear)
* Fedora 30, 32, 33

heasterday:
Fedora notes:
 * We plan to support fedora 24 -> 35 (rawhide) as these are the images on dockerhub and it appears one config handles them all.  
Clear Linux notes:
  * Software is almost always installed via "bundles" which are software collections aimed at providing a specific application/toolset.  
  * `fakeroot` is packaged in their repos but it is only available via the [os-testsuite-0day bundle](https://clearlinux.org/software/bundle/os-testsuite-0day).  
  * Almost every "bundle" install fails upriv so it would be really nice if we could get fakeroot as it's own bundle.

heasterday:
In talking to Reid to decision was made to split the Fedora support and Clear Linux support into two separate issues. This will be for Fedora support.

loveshack:
I'd missed this before doing #1035, which will fix it in case it hasn't already been done elsewhere. fakeroot goes way back in Fedora, if that's a concern <https://koji.fedoraproject.org/koji/packageinfo?packageID=1704>.

heasterday:
@loveshack Thanks for the PR Dave! I also have one [here](https://github.com/hpc/charliecloud/pull/1036) that addresses some of the same things. What do you say I incorporate your PR into mine, I credit you, and you give the PR a review when it is ready? I see you additionally extended the support to Debian 11 and OpenSuse which would be awesome to have!

loveshack:
> What do you say I incorporate your PR into
> mine, I credit you, and you give the PR a review when it is ready?

Do whatever is helpful.  I didn't know whether there was anything else
in flight.  I'm not sure I should be reviewing, though.  I did test the
changes, anyway.

> I
> see you additionally extended the support to Debian 11 and OpenSuse
> which would be awesome to have!

By the way, I think there shouldn't be an upper limit on the versions,
as it's unlikely this will stop working in the near future.
Particularly, if you upper limit Fedora to the current versions, it will
only last a year.

dschwoerer:
This seems to fail for fedora 36, with:
```
workarounds: init step 1: $ dnf install -y fakeroot                                                                                                                                                                                [227/1362]
[...]

PR #1035, opened 2021-04-19
Update --force support

loveshack:
RHEL would fail because you can't install epel-release there, only on CentOS.
I added support for Fedora, which I wanted, and SUSE while I was at it. Tested with Fedora 33 and current OpenSUSE leap.


reidpr:
Addresses #1026, #1030, #1043, #1044.

reidpr:
Looks like the one remaining distro (SUSE, issue #1044) is covered by PR #1276? Can I close this?

Regardless much appreciate the help.

loveshack:
> Looks like the one remaining distro (SUSE, issue #1044) is covered by
> PR #1276? Can I close this?

Yes, I think so.

PR #1036, opened 2021-04-19
ch-image: add --force support for fedora, install epel via url

heasterday:
This addresses #1026 and #1030.

Marking draft pending further local testing.

### DIFF
--- a/lib/fakeroot.py
+++ b/lib/fakeroot.py
@@ -223,13 +223,13 @@ DEFAULT_CONFIGS = {
 
    # Fedora notes:
    #
-   # 1. The supported versions were chosed somewhat arbitrarily based on the
+   # 1. The supported versions were chosen somewhat arbitrarily based on the
    #    release versions available for building (what was on Dockerhub). To
    #    extend support to additional versions one need only adjust the match
    #    config.
    #
-   # 2. The fakeroot package is in the base repository set on so enabling EPEL
-   #    is not required.
+   # 2. The fakeroot package is in the base repository set so enabling EPEL is
+   #    not required.
 
    "fedora":
    { "name": "Fedora 24-35,",
--- a/lib/fakeroot.py
+++ b/lib/fakeroot.py
@@ -148,7 +148,7 @@ DEFAULT_CONFIGS = {
    #    repos.
    #
    # 2. Unlike on CentOS, RHEL doesn't have the epel-release rpm in the
-   #    standard repos, so install the rpm directly.
+   #    standard repos, so install via rpm for both.
    #
    # 3. Enabling EPEL can have undesirable side effects, e.g. different
    #    version of things in the base repo that breaks other things. Thus,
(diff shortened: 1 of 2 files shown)


lib/fakeroot.py
```diff
@@ -147,15 +147,18 @@
    # 1. CentOS seems to have only fakeroot, which is in EPEL, not the standard
    #    repos.
    #
-   # 2. Enabling EPEL can have undesirable side effects, e.g. different
+   # 2. Unlike on CentOS, RHEL doesn't have the epel-release rpm in the
+   #    standard repos, so install the rpm directly.
```
reidpr, line 151: Should clarify we're doing the RPM way for both. Or did you mean “unlike on Fedora”?

lib/fakeroot.py
```diff
@@ -221,6 +224,24 @@
...
+   #    release versions available for building (what was on Dockerhub). To
+   #    extend support to additional versions one need only adjust the match
+   #    config.
+   #
+   # 2. The fakeroot package is in the base repository set so enabling EPEL is
+   #    not required.
+
+   "fedora":
+   { "name": "Fedora 24-35,",
+     "match":  ("/etc/fedora-release", r"release (2[4-9]|3[0-5])"),
```
reidpr, line 239: I think Dave Love had a good point. Can we remove the high end of this range test?

test/build/50_fakeroot.bats
```diff
@@ -693,3 +709,191 @@ EOF
     [[ $output = *'will use --force'* ]]
     [[ $output = *'--force: init OK & modified 1 RUN instructions'* ]]
 }
+
+# We would prefer to test the lowest supported --force version, 24, but the
+# ancient version of dnf it has doesn't fail the transaction when a package
+# fails so we test with 26 instead.
```
reidpr, line 715: Move this comment inside the test.

test/build/50_fakeroot.bats
```diff
@@ -693,3 +709,191 @@ EOF
...
+    [[ $output = *'will use --force'* ]]
+    [[ $output = *'--force: init OK & modified 1 RUN instructions'* ]]
+}
+
+@test "${tag}: Fedora 35: unneeded, no --force, build succeeds" {
+    scope standard
+    # no commands that may need it, without --force, build succeeds
+    # also: correct config
+    run ch-image -v build -t fakeroot-temp -f - . <<'EOF'
+FROM fedora:35
```
reidpr, line 814: Similar to above, can we do `fedora:latest` here? (which seems to be Fedora 33 currently) Goal being to test the latest released Fedora, not a pre-release, though? Or maybe we really do want `fedora:rawhide` so we can fix problems early?

reidpr (changes requested):
Good work. Some comments inline.

Were you planning to add the Debian 11 (#1043) and openSUSE (#1044) from PR #1035 here?

See also #1042.

lib/fakeroot.py
heasterday: Can do, I'll just check that it is above version 24.

lib/fakeroot.py
heasterday: > Good work. Some comments inline.
> 
> Were you planning to add the Debian 11 (#1043) and openSUSE (#1044) from PR #1035 here?
> 
> See also #1042.

I was going to make an issue for Debian 11 (doesn't appear out yet) and OpenSUSE, looks like you already did which is quite helpful!

heasterday: Will clarify that we install via RPM for both.

test/build/50_fakeroot.bats
heasterday: Let's do `fedora:latest`, I know nothing about how stable their pre-release images are.

lib/fakeroot.py
```diff
@@ -221,6 +224,22 @@
      "cmds": ["apt", "apt-get", "dpkg"],
      "each": ["fakeroot"] },
 
+   # Fedora notes:
+   #
+   # 1. The minimum supported version was chosen somewhat arbitrarily based on the
+   #    release versions available for testing (what was on Dockerhub).
```
reidpr, line 230: I'm going to tidy this comment by spelling Docker Hub correctly. You would be correct to point out it's a trivial change; the purpose is to make searches for "Docker Hub" find this spot.

```suggestion
   #    release versions available for testing (what was on Docker Hub).
```

reidpr (approved):

reidpr:
We decided to put Debian 11 (#1043) and openSUSE (#1044) in a separate PR.

Issue #1040, opened 2021-04-27
ch-image pull: test registry.access.redhat.com


reidpr:
E.g., UBI images come from here. Pretty sure this works but we should add it to `40_pull.bats`.

PR #1045, opened 2021-04-27
add test for registry.access.redhat.com

heasterday:
Addresses #1040

The Redhat registry is a mixture of images that require auth and those that do not.  After looking around I haven't found a public image smaller than 31/78 MiB (compressed/uncompressed) although it's quite possible one exists.

### DIFF
--- a/test/build/40_pull.bats
+++ b/test/build/40_pull.bats
@@ -289,6 +289,10 @@ EOF
     LC_ALL=en_US.utf-8 \
     ch-image pull nvcr.io/hpc/foldingathome/fah-gpu:7.6.21
 
+    # Redhat Registry: https://catalog.redhat.com/software/containers/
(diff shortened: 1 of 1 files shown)


test/build/40_pull.bats
```diff
@@ -289,6 +289,10 @@ EOF
     LC_ALL=en_US.utf-8 \
     ch-image pull nvcr.io/hpc/foldingathome/fah-gpu:7.6.21
 
+    # Redhat Registry: https://catalog.redhat.com/software/containers/
```
reidpr, line 292: Typo; given URL was 404. Suggest:

```suggestion
    # Red Hat registry: https://catalog.redhat.com/software/containers/explore
```

reidpr (approved):

Issue #1048, opened 2021-05-03
build failure with gcc 8.4.0


loveshack:
I tried to build the current head on ppc64le RHEL 7 with gcc 8.4.0 and got a strange maybe-uninitialized warning, converted to error (below).  The message seems unrelated to the source, which presumably indicates a GCC bug.  I don't see the problem on x86 Debian 10 with gcc 8.3.  
I'll provide a workaround.

```
gcc -DHAVE_CONFIG_H -I.     -g -O2 -std=c99 -Wpedantic -Wall -Werror -MT ch_misc.o -MD -MP -MF $depbase.Tpo -c -o ch_misc.o ch_misc.c &&\
mv -f $depbase.Tpo $depbase.Po
ch_misc.c: In function 'mkdirs':
ch_misc.c:106:14: error: '__s' may be used uninitialized in this function [-Werror=maybe-uninitialized]
       next = cat(next, component);  // canonical except for last component
              ^~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

PR #1049, opened 2021-05-03
strtok_r(3): initialize saveptr to NULL

loveshack:


reidpr:
Addresses #1048.

reidpr:
Good catch @loveshack; thank you! I pushed a new commit with some style tidying. Does this still work for you? If so I will merge.

Note the [strtok_r(3)](https://man7.org/linux/man-pages/man3/strtok.3.html) man page mentions `saveptr` needing to be `NULL` sometimes. Maybe that is the problem.

reidpr:
I think the CI failure is unrelated but I will look more carefully later today.

loveshack:
Yes, it still works.

For what it's worth, POSIX says you don't need to initialize the save
pointer, and I don't think GCC could tell if you needed to.

In related news, while trying to understand what was happening I built
with -O3 and got this:

```
ch_misc.c: In function ‘msg.constprop’:
ch_misc.c:154:7: error: array subscript 4 is above array bounds of
‘const char *[4]’ [-Werror=array-bounds]
  154 |       fputs(VERBOSE_LEVELS[level], stderr);
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ch_misc.c:28:13: note: while referencing ‘VERBOSE_LEVELS’
   28 | const char *VERBOSE_LEVELS[] = { "error", "warning", "info", "debug" };
      |             ^~~~~~~~~~~~~~
```

Issue #1050, opened 2021-05-03
mount /etc/machine-id as optional default to support D-Bus


loveshack:
I tried to run a graphical program (paraver) and got a dbus-related error about /etc/machine-id being empty.  Mounting that into the container fixes it.

reidpr:
Thanks @loveshack. I was not familiar with this file. Looks like it is a unique ID of the OS installation ([man page](https://man7.org/linux/man-pages/man5/machine-id.5.html), [Debian wiki](https://wiki.debian.org/MachineId)). So by bind-mounting it from the host, we declare that the host and the container are the same OS installation. That's certainly a kludge, but it could be a good kludge. Questions:

* Are you aware of any arguments that this is or is not a best practice for containers?
* `--bind /etc/machine-id` is available now as a workaround; is that sufficient for your use case?
* If we do want to initialize the file with unique but properly formatted value, it's straightforward to do this with `/dev/urandom` and POSIX tools.
* Would a FAQ entry be sufficient for your use case?

reidpr:
I spoke with my colleague @mej about this, who's an expert on such matters. His opinions:

* There does not appear to be a consensus on whether container `/etc/machine-id` should match the host or be generated new for each container.
* It does seem that the file should contain a valid value when the container is running. On the other hand, when not running the file should be empty. (This implies that if we don't match the host, we need to keep track of the value some other way.)
* This ID is sometimes used for software licensing, which points to matching the host.
* Bind-mounting from the host seems to be a common approach (e.g. [1](https://github.com/denisbrodbeck/machineid/issues/5) [2](https://forums.docker.com/t/host-machine-id-visible-from-containers/100533/4)).

Given that, I'm inclined to proceed with the proposed solution.

loveshack:
https://www.redhat.com/en/resources/cloud-native-container-design-whitepaper
says in support:  "Some containerized applications require the container
to be synchronized with the host on certain attributes such as time and
machine ID".  (Not that I've gone "cloud native".)

loveshack:
In case it's still relevant, I should have said that an explicit --bind
works, of course.

[The best I came up with for generating the id with plain POSIX utils
was an od | cut | tr with appropriate args.]

mej:
[...]

PR #1051, opened 2021-05-03
Maybe mount /etc/machine-id

loveshack:


reidpr:
Addresses #1050.

reidpr:
Hello @loveshack; I pushed a style tweak and some documentation corrections. If it looks good to you, I'll merge.

doc/ch-run_desc.rst
```diff
@@ -100,31 +109,30 @@ In addition to any directories specified by the user with :code:`--bind`,
...
-    :code:`/home/$USER` by default. This is accomplished by mounting a new
-    :code:`tmpfs` at :code:`/home`, which hides any image content under that
-    path. If :code:`--no-home` is specified, neither of these things happens
-    and the image's :code:`/home` is exposed unaltered.
+  * :code:`/etc/hosts`
+  * :code:`/etc/machine-id`
+  * :code:`/etc/resolv.conf`
+  * :code:`/var/lib/hugetlbfs` at guest :code:`/var/opt/cray/hugetlbfs`
+  * :code:`/var/opt/cray/alps/spool`
+  * :code:`$PREFIX/bin/ch-ssh` at guest :code:`/usr/bin/ch-ssh`
```
heasterday, line 127: Do we already document why we mount these files/directories? If not, I think that would be helpful for some of the less obvious ones.

I'm thinking these could use some explanation:
 * `/etc/machine-id`
 * `/var/lib/hugetlbfs`
 * `/var/opt/cray/alps/spool`
 * `$PREFIX/bin/ch-ssh`
 
 Thoughts?

heasterday:
Tested this locally and it looks good, only one minor comment (which may be a separate PR).

doc/ch-run_desc.rst
reidpr: (1) No (2) I agree

Issue #1060, opened 2021-05-11
FTBFS with Sphinx 4.0


reidpr:
Simple fix is to set [`man_make_section_directory`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-man_make_section_directory) to `False`. I think we can still do this with old Sphinx (we support down to 1.2.3).

PR #1062, opened 2021-05-12
configure Sphinx 4.x behavior to match 3.x

reidpr:
Addresses #1060.

@j-ogas, can you make sure this works in old-school Red Hat Sphinxes? E.g. is that covered by the test suite RPM builds?


j-ogas (approved):
I tested building out the charliecloud rpms on Centos 7 (Sphinx v1.2.3). The changes here do not interfere with man page or html creation.

PR #1072, opened 2021-05-14
PR #1072: add runtime check for Python version (closes #1070)

### LINKED ISSUE #1070: validate Python version as a dependency
We validate Python version in `configure` and include in the report, but it's very easy for users to get a old Python interpreter, often through no real fault of their own. Add this check to `charliecloud.dependencies_check()`.

heasterday:
Addresses issue #1070

Example output:
```
error: outdated dependency: found Python 3.5; Needs 3.6 minimum
       executable: /home/heasterday/anaconda3/envs/python-3.1/bin/python3
```

NOTE: This doesn't add a dependency check to ch-run-oci as it is a bit of an oddball. @reidpr let me know if you would like that added.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -315,7 +315,7 @@ AS_IF([test -n "$NVIDIA_CLI"],
 AC_MSG_RESULT($have_nvidia_libs)
 
 # Python
-vmin_python=3.6
+vmin_python=3.6  # NOTE: Keep in sync with lib/charliecloud.py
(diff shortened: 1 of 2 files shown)


reidpr (commented):
Pushed a tidy; if it looks OK for you then I'll merge.

Printing out the Python executable in use seems very helpful and I would not have thought of that.

heasterday:
@reidpr Looks good to me

Issue #1071, opened 2021-05-14
testing: validate we can pull from Microsoft Container Registry


reidpr:
Docs: https://github.com/microsoft/containerregistry

PR #1073, opened 2021-05-14
ch-test: test Microsoft Container Registry (MCR)

heasterday:
Addresses #1071 

The Microsoft Container Registry is interesting in that they don't have a public UI to their catalog, instead they leverage Dockerhub as outlined in [this blog post](https://cloudblogs.microsoft.com/opensource/2019/01/17/improved-discovery-experience-microsoft-containers-docker-hub/).

### DIFF
--- a/test/build/40_pull.bats
+++ b/test/build/40_pull.bats
@@ -293,6 +293,11 @@ EOF
     # FIXME: 77 MiB unpacked, should find a smaller public image
     ch-image pull registry.access.redhat.com/ubi7-minimal:latest
 
+    # Microsoft Container Registry:
+    # MCR doesn't have a catalog, Docker Hub is the way to browse images
+    # https://hub.docker.com/publishers/microsoftowner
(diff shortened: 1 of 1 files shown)


test/build/40_pull.bats
```diff
@@ -293,6 +293,12 @@ EOF
     # FIXME: 77 MiB unpacked, should find a smaller public image
     ch-image pull registry.access.redhat.com/ubi7-minimal:latest
 
+    # Microsoft Container Registry:
+    # https://hub.docker.com/publishers/microsoftowner
```
reidpr, line 297: This link is a bit confusing because it goes to Docker Hub but we're pulling from `mcr.microsoft.com`. Would the GitHub link I sent you work here?

reidpr (changes requested):

test/build/40_pull.bats
heasterday: Sounds good. I will also remove the comment about Dockerhub being the official method to browse images since that information will be in the github link.

Issue #1023, opened 2021-04-05
ch-image list: better error message if no builder storage


reidpr:
```
$ bin/ch-image -s /tmp/bar list
error: can't list directory: /tmp/bar/img: No such file or directory
```

Should be a friendlier error that doesn't include storage directory layout (`img` subdirectory).

PR #1074, opened 2021-05-17
ch-image list: better error msg

mphinney1100:
Addresses #1023

### DIFF
--- a/lib/misc.py
+++ b/lib/misc.py
@@ -43,9 +43,11 @@ def list_(cli):
    ch.dependencies_check()
    imgdir = ch.storage.unpack_base
    if (not os.path.isdir(ch.storage.root)):
-      ch.FATAL("does not exist: %s" % ch.storage.root)
+      ch.INFO("does not exist: %s" % ch.storage.root)
+      return;
    if (not ch.storage.valid_p()):
-       ch.FATAL("doesn't contain subdirectory img: %s" % ch.storage.root)
+       ch.INFO("not a storage directory: %s" % ch.storage.root)
+       return;
    imgs = ch.ossafe(os.listdir, "can't list directory: %s" % ch.storage.root, imgdir)
    for img in sorted(imgs):
       print(ch.Image_Ref(img))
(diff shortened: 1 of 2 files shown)


lib/misc.py
```diff
@@ -42,7 +42,8 @@ def delete(cli):
 def list_(cli):
    ch.dependencies_check()
    imgdir = ch.storage.unpack_base
-   imgs = ch.ossafe(os.listdir, "can't list directory: %s" % imgdir, imgdir)
+   imgs = ch.ossafe(os.listdir, "not a storage directory: %s" 
+        % ch.storage.root, imgdir)
```
reidpr, line 46: I think there's actually two separate error conditions here:

1. `ch.storage.root` does not exist; error should just say that.
2. `ch.storage.root` does exist but doesn't contain a subdirectory `img`. Error should say something to that effect. (I remember seeing a method in `Storage` that's yes/no on whether the storage directory looks valid. Possibly part of PR #1024?)

lib/misc.py
```diff
@@ -42,7 +42,11 @@ def delete(cli):
 def list_(cli):
    ch.dependencies_check()
    imgdir = ch.storage.unpack_base
-   imgs = ch.ossafe(os.listdir, "can't list directory: %s" % imgdir, imgdir)
+   if (not os.path.isdir(ch.storage.root)):
+      ch.FATAL("does not exist: %s" % ch.storage.root)
+   if (not ch.storage.valid_p()):
+       ch.FATAL("doesn't contain subdirectory img: %s" % ch.storage.root)
+   imgs = ch.ossafe(os.listdir, "can't list directory: %s" % ch.storage.root, imgdir)
```
j-ogas, line 49: A few thoughts:
1. If the storage root doesn't exist, why not create it and then say there are no images?
2. Do we need to exit with a failure? For example:
```
$ ch-image list
No images in storage.
$ echo $?
0
```

reidpr (approved):
Looks good to go! Once you're happy please mark as ready for review and I'll merge.

PR #1076, opened 2021-05-18
issue organization overhaul

reidpr:
Here is a concrete proposal to organize issues, per discussions.

See also internal google doc: https://docs.google.com/document/d/1PiaIJOBZ0Drxp1lXMBCLEz8Qu4vbAC3Q1nQ2bzO_AHw/edit

Summary of the proposal:

* There are several dimensions of labels, including priority and category (called *type* here).
* This leaves Projects unused; they remain open for possible future use as scoped units of work larger than issues.
* All label colors will be overhauled with ColorBrewer.
* Yes, the plan is to re-label every open issue (~200 of them, plus some internal ones). The 450 closed issues will not be reviewed.
* Enable Discussions.
* Plan:
  1. Revise PR until it looks good.
  2. Re-label everything.
  3. Adjust PR again based on what's learned in previous step.
  
Renamed labels:

* *Unprivileged build* project → *image*
* *erroneous* → *jurisdiction*
* *obsolete* → *moot*

Deleted labels:

* *dumpster fire* — needless editorializing, no open issues
* *ready to merge* — replaced by draft/ready PRs, any remaining use is an error even in closed PRs

### DIFF
--- a/doc/dev.rst
+++ b/doc/dev.rst
@@ -236,8 +236,9 @@ re-tagged).
 Issue labeling
 --------------
 
-We use the following labels (a.k.a. tags) to organize issues. Each issue
-should have appropriate label(s) from each category.
+We use the following labels (a.k.a. tags) to organize issues. Each issue (or
+stand-alone PR) should have label(s) from every category, with the exception
+of disposition which only applies to closed issues.
 
 .. note::
 
@@ -260,9 +261,8 @@ Choose *one type* from:
   desired and actual behavior.
 
 :code:`refactor`
-  Change that will improve Charliecloud but does not materially affect the
-  documented behavior. However, user-visible behavior may change if
-  undocumented or unspecified, e.g. the text of error messages.
+  Change that will improve Charliecloud but does not materially affect
+  user-visible behavior.
 
 How important/urgent is it?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -277,7 +277,7 @@ Choose *one priority* from:
   Medium priority.
 
 :code:`low`
-  Low priority. Note, unfortunately due to resource limitations, that complex
+  Low priority. Note: Unfortunately, due to resource limitations, complex
   issues here are likely to wait a long time, perhaps forever. If that makes
   you particularly sad on a particular issue, please comment to say why. Maybe
   it's mis-prioritized.
@@ -288,10 +288,10 @@ Choose *one priority* from:
   risky; you probably want to argue successfully that it should be done before
   starting work on it.
 
-Note that (1) the priority levels are indeed fuzzy, and we have current plans
-to make them more precise; (2) issues are not a priority queue, partly for
-this reason, i.e., we do work on lower-priority issues while higher-priority
-ones are still open; and (3) issues often move between priority levels. In
+Note that (1) the priority levels are indeed fuzzy, and we have no plans to
+make them more precise; (2) issues are not a priority queue, partly for this
+reason, i.e., we do work on lower-priority issues while higher-priority ones
+are still open; and (3) issues often move between priority levels. In
 particular, if you think we picked the wrong priority level, please say so.
 
 What part of Charliecloud is affected?
@@ -302,13 +302,16 @@ Choose *one or more components* from:
 :code:`runtime`
   The container runtime itself; largely :code:`ch-run`.
 
-:code:`build/mgmt`
+:code:`image`
   Image building and interaction with image registries; largely
-  :code:`ch-image`. Not to be confused with image management tasks done by
-  glue code.
+  :code:`ch-image`. (Not to be confused with image management tasks done by
+  glue code.)
 
 :code:`glue`
   Largely shell scripts in :code:`bin`.
+
+:code:`build`
+  Build system and packaging. (Not to be confused with image building.)
 
 :code:`doc`
   Documentation.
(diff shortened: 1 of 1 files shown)


doc/dev.rst
```diff
@@ -249,49 +260,139 @@ What kind of issue is it?
...
+
+:code:`runtime`
+  The container runtime itself; largely :code:`ch-run`.
+
+:code:`image`
+  Image building and interaction with image registries; largely
+  :code:`ch-image`. (Not to be confused with image management tasks done by
+  glue code.)
+
+:code:`glue`
```
j-ogas, line 310: I don't think `glue` is descriptive or informative. Unlike `image`, `runtime`, `build`, and `doc`, it doesn't immediately correspond to some action or idea. What about `script`?

doc/dev.rst
```diff
@@ -249,49 +260,139 @@ What kind of issue is it?
...
+  Largely shell scripts in :code:`bin`.
+
+:code:`build`
+  Build system and packaging. (Not to be confused with image building.)
```
j-ogas, line 314: Could `package` work here? I think `build` will most certainly be confused with image building.

doc/dev.rst
```diff
@@ -249,49 +260,139 @@ What kind of issue is it?
   a new feature proposal or refactoring. Should have steps to reproduce with
```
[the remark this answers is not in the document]
reidpr: Good catch. The proposal is that `enhancement` is user-visible, `refactoring` is not. The intent is `refactoring` would be for internal cleanup stuff that doesn't affect how things work; users won't care about `refactoring` issues except in the general sense of Charliecloud being more reliable is good for users. The word *refactoring* should be removed from `enhancement`.

doc/dev.rst
reidpr: Well `bin/ch-image` is unfortunately a script too. I agree `glue` isn't the best.

reidpr: Though `glue` has an emoji (🍝) while `script` does not.

doc/dev.rst
reidpr: Possibly `install`?

doc/dev.rst
```diff
@@ -233,11 +233,22 @@ images. The former is generally preferred, as it lets you update only those
...
-What kind of issue is it?
-~~~~~~~~~~~~~~~~~~~~~~~~~
+We use the following labels (a.k.a. tags) to organize issues. Each issue (or
+stand-alone PR) should have label(s) from every category, with the exception
+of disposition which only applies to closed issues.
+
+.. note::
+
+   This scheme is designed to organize open issues only. There have been
+   previous schemes, and we have not re-labeled closed issues.
```
heasterday, line 246: I think a statement on who is encouraged to label issues would be good. Is the goal for the issue submitter to self-label, the internal team label, or just you, Reid?

doc/dev.rst
```diff
@@ -249,49 +260,139 @@ What kind of issue is it?
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+:code:`medium`
+  Medium priority.
 
-For all of these, leave other tags in place, e.g. :code:`bug`.
+:code:`low`
+  Low priority. Note: Unfortunately, due to resource limitations, complex
+  issues here are likely to wait a long time, perhaps forever. If that makes
+  you particularly sad on a particular issue, please comment to say why. Maybe
+  it's mis-prioritized.
```
heasterday, line 283: Thoughts on including an example for each priority?

doc/dev.rst
heasterday, line 314: What about `packaging` or `distribution`?

doc/dev.rst
```diff
@@ -249,49 +260,139 @@ What kind of issue is it?
...
+  Same as some other issue. In addition to this tag, duplicates should refer
+  to the other issue in a comment. Of the duplicates, the better one should
+  stay open (e.g., clearer reproduction steps); if they are roughly equal in
+  quality, the older one should stay open.
+
+:code:`jurisdiction`
+  The issue is not something we can resolve. Typically problems with other
+  software or problems with containers in general that we can't work around.
+  *Use caution when blaming a problem on user error. Often (or usually) there
+  is a documentation or usability bug that caused the "user error".*
```
heasterday, line 364: Minor thing but I'm not sure I like `jurisdiction` although I'm having a tough time articulating why exactly. Perhaps `scope` or `external` instead?

doc/dev.rst
reidpr: Discussion 5/19:

* emphasize priority is required
* explain how to choose priority
* who? project lead is responsible for assigning priority, though everyone should give it a shot?

doc/dev.rst
reidpr: Other less friendly alternatives I came up with:

* `notcharliecloud`
* `notourproblem`
* `yourmonkeys`

doc/dev.rst
mphinney1100: I was confused by `build` but `install` was the least confusing for me.

doc/dev.rst
reidpr: I took a crack at this; please take a look when I push.

I hesitate to include examples because I'm concerned about them being representative, i.e., avoid over-generalizing.

doc/dev.rst
reidpr: So far I don't see anything I like better than jurisdiction. I agree it's a tad awkward. How much do folks dislike it?

reidpr: * `nothingtodo`
* `notactionable`
 
?

doc/dev.rst
j-ogas: > Well `bin/ch-image` is unfortunately a script too. I agree `glue` isn't the best.

Hmm, since we are typically referring the bash/sh scripts when we refer to glue, can we do something like `POSIX sh` and/or `bash`?

doc/dev.rst
j-ogas: I think `cantfix` is more friendly than `erroneous` and a bit more helpful than `jurisdiction`.

heasterday: I like `cantfix` from that new list.

doc/dev.rst
reidpr: In the interest of moving forward, I'm going to call it `glue` for now. But we can easily change it later.

Issue #1067, opened 2021-05-13
ch-build: --force causes failures when target Dockerfile also installs EPEL


heasterday:
Currently the `--force` flag for ch-build installs EPEL, installs fakeroot, then disabled EPEL. This leaves the EPEL repo installed but disabled. The side effect of this is that if the Dockerfile also tries to install EPEL, while that operation won't fail, it won't result in EPEL being enabled, leading to the build failing later when trying to install packages from EPEL.

heasterday:
This failure can be reproduced using this example Dockerfile:
```dockerfile
FROM centos:8
RUN dnf install -y epel-release && dnf install -y acme-tiny
```

reidpr:
Some background that we discussed offline but would be good to include here:

* The reason for installing EPEL disabled is that leaving it enabled causes a different but related problem: EPEL has packages with the same name as the main repo but different versions, so EPEL unexpectedly enabled can cause the wrong versions of things to be installed later in the Dockerfile.
* We are investigating a replacement workflow: install EPEL, install fakeroot, uninstall EPEL.

PR #1084, opened 2021-05-28
lib/fakeroot.py: uninstall EPEL when appropriate after installing fakeroot

heasterday:
Addresses #1067

### DIFF
--- a/examples/Dockerfile.centos7
+++ b/examples/Dockerfile.centos7
@@ -6,7 +6,6 @@ FROM centos:7
 
 # Install our dependencies, ensuring we fail out if any are missing.
 RUN yum install -y epel-release \
- && yum-config-manager --enable epel \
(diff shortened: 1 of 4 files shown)


reidpr (changes requested):
Nice work.

A few questions/recommendations:

* I recommend you get a review from Tim because he knows better the implications here.
* In the case where we install and then remove EPEL, does fakeroot become un-upgradeable, and if so do we care?
* Tests where EPEL is already installed should verify we didn't remove it.

heasterday:
Hey Reid,
 * Re un-upgradable fakeroot: I believe it does become un-upgradeable but I don't think we care, although I'm curious for Tim's thoughts.
 * Re already install EPEL testing: Looks like we already do this for CentOS 8, added the test for CentOS 7.

reidpr (approved):

Issue #1086, opened 2021-06-02
ch-image delete: doesn't works with -- storage


mphinney1100:
`ch-image --storage DIR delete IMG` doesn't delete the right directory.  
``` 
vm-user@mphinney:~/charliecloud$ bin/ch-image -vv -s /tmp/ch-test.tmp.vm-user/bats.tmp/ delete alpine
verbose level: 2
image_ref
  ir_name   alpine

error: storage directory seems broken: not an image: alpine
```

PR #1087, opened 2021-06-02
ch-image delete --storage works

mphinney1100:
Addresses #1086.

### DIFF
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -632,7 +632,7 @@ class Image:
    def unpack_delete(self):
       if (not self.unpack_exist_p()):
          FATAL("%s image not found" % self.ref)
-      VERBOSE("image path: %s" % self.unpack_path)
+      VERBOSE("unpack path: %s" % self.unpack_path)
       if (self.unpacked_p(self.unpack_path)):
          INFO("deleting image: %s" % self.ref)
          rmtree(self.unpack_path)
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -223,7 +223,10 @@ class Image:
    def __init__(self, ref, unpack_path=None):
       assert isinstance(ref, Image_Ref)
       self.ref = ref
-      self.unpack_path = storage.unpack(self.ref)
+      if (unpack_path is not None):
+         self.unpack_path = Path(unpack_path)
+      else:
+         self.unpack_path = storage.unpack(self.ref)
       self.metadata_init()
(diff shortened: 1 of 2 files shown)


lib/charliecloud.py
```diff
@@ -223,10 +223,7 @@ class Image:
    def __init__(self, ref, unpack_path=None):
       assert isinstance(ref, Image_Ref)
       self.ref = ref
-      if (unpack_path is not None):
-         self.unpack_path = Path(unpack_path)
-      else:
-         self.unpack_path = storage.unpack(self.ref)
```
reidpr, line 229: We do need to keep this conditional; `unpack_path` being not None is used when we want to pull somewhere other than the storage directory.

reidpr (changes requested):
I think you're on to something but this doesn't quite fix it.

One thing to add that may help debug, and would be useful in general, is adding `VERBOSE("unpack path: %s" % self.unpack_path")` to the beginning of `Image.unpack_delete()`.

You might also examine my recent changes to `Image.unpacked_p` (e.g., with `git blame`). It's possible those changes introduced this bug.

lib/charliecloud.py
```diff
@@ -630,6 +630,7 @@ def unpack_create_ok(self):
          rmtree(self.unpack_path)
 
    def unpack_delete(self):
+      VERBOSE("unpack path: %s" % self.unpack_path)
```
reidpr, line 633: I moved this to the beginning of the function so it's output even if the image doesn't exist.

lib/misc.py
```diff
@@ -36,7 +36,7 @@ def __call__(self, *args, **kwargs):
 def delete(cli):
    ch.dependencies_check()
    img_ref = ch.Image_Ref(cli.image_ref)
-   img = ch.Image(img_ref, cli.storage)
+   img = ch.Image(img_ref)
```
reidpr, line 39: So this was a little subtle. I think with your solution you added a second error that compensated for the first, so and it ended up working, i.e. it was a “two wrongs make a right” approach. ;)

The initial error was passing the storage directory when `Image` expected the complete path. If that's not given, then it computes the unpack path based on the storage directory. It worked with `--storage` omitted because in that case it gets `unpack_path=None`, which is the same as the default.

reidpr (approved):
Added a tidy and explanation. Please give it a whirl, and if it looks good, set it ready for review and I'll merge.

Issue #1041, opened 2021-04-27
testing: delete temp images after use


reidpr:
There are a lot of BATS tests that build throw-away images not used after completion of the test. Now that we have `ch-image delete`, remove those images at the end of their tests.

PR #1094, opened 2021-06-14
delete temp images in testing 

mphinney1100:
Addresses #1041.


test/build/40_pull.bats
```diff
@@ -255,6 +255,8 @@ EOF
                   "$img"
     [[ $status -eq 0 ]]
     grep -F '"schemaVersion": 1' "${cache}/${img}.manifest.json"
+
+    run ch-image --storage="$unpack" delete 'debian:squeeze'
```
reidpr, line 259: I recommend changing this test to:

1. use a subdirectory of `$BATS_TMPDIR` for the storage directory, not `$BATS_TMPDIR` itself
2. remove `--no-cache`
3. just delete the whole storage directory with `rm -Rf`

With those changes, the following will be moot, but: Note also that if you wrap a command with `run`, you need to check the exit value and print the output manually. Here I'd suggest just omitting `run`.

test/build/50_ch-image.bats
```diff
@@ -250,6 +250,8 @@ RUN test -f /mnt/0/empty-file
 EOF
     echo "$output"
     [[ $status -eq 0 ]]
+
+    run ch-image delete build-bind
```
reidpr, line 254: Looks like the same error with `run` needs to be fixed throughout.

reidpr (changes requested):
Thanks, good work! See comments in-line.

test/build/50_dockerfile.bats
```diff
@@ -393,6 +399,8 @@ EOF
     [[ $output = *'warning: not supported, ignored: STOPSIGNAL instruction'* ]]
     [[ $output = *'warning: not supported, ignored: USER instruction'* ]]
     [[ $output = *'warning: not supported, ignored: VOLUME instruction'* ]]
+
+    run ch-image delete unsupported
```
reidpr, line 403: Several of these hanging around still.

```suggestion
    ch-image delete unsupported
```

test/build/50_dockerfile.bats
```diff
@@ -471,6 +479,10 @@ EOF
   echo "$output"
   [[ $status -eq 0 ]]
   diff -u <(echo "$env_expected") <(echo "$output" | grep -E "^\('chse_")
+
+  if [[ $CH_BUILDER = ch-image ]]; then
+    ch-image delete env-syntax
+  fi
```
reidpr, line 485: So this raises something interesting I had not thought of — it seems awkward to delete the images we don't need for only `ch-image` (and leave them for other builders). I think there are at least two options:

1. Write several lines of code for every multi-builder test to select the right delete command. You'd probably want to include Buildah but it won't be tested.
2. Change all these tests to use the *same* temporary image name — and don't delete it. Also document the name in the contributor's guide. That will give us one image hanging around, which seems OK.

What do you think?

reidpr (changes requested):
Thanks Megan. Two comments in-line, one of which proposes a significant change to the PR.

test/build/50_dockerfile.bats
mphinney1100: I am leaning towards the 2nd idea. It seems cleaner and nicer to use one variable name that overwrites the image vs adding extra code to decide how to delete said image.

reidpr: make it so

test/build/40_pull.bats
```diff
@@ -243,18 +243,19 @@ EOF
...
     # thus keeps test time down.
     img=debian:squeeze
 
     ch-image pull --storage="$unpack" \
-                  --no-cache \
                   "$img"
     [[ $status -eq 0 ]]
+
     grep -F '"schemaVersion": 1' "${cache}/${img}%skinny.manifest.json"
+    rm -Rf "$unpack"
```
reidpr, line 258:
```suggestion
    grep -F '"schemaVersion": 1' "${cache}/${img}%skinny.manifest.json"

    rm -Rf "$unpack"
```

test/common.bash
```diff
@@ -198,6 +198,9 @@ mkdir -p "$btnew"
 chmod 700 "$btnew"
 export BATS_TMPDIR=$btnew
 [[ $(stat -c %a "$BATS_TMPDIR") = '700' ]]
+# all temporary images all tagged TMPIMG so there is only 1 extra image at end
+# test case.
+export TMPIMG="foo"
```
reidpr, line 203: I think we don't need a variable here; the indirection doesn't buy us much (unless I am missing something?). I recommend you just call it `tmpimg` directly.

reidpr (changes requested):
Good progress!

Couple things in-line.

Also, please document this image briefly in the contributor's guide. You could create a new §5.2.3 (before the existing one) called "Special images" and just write a couple sentences about it. Eventually it would be good to document `chtest` here too.

reidpr (approved):
LGTM! Let's make sure CI passes and then I'll merge.

Issue #1095, opened 2021-06-15
ch-image build --force: support CentOS Stream


sbiae5:
With the death of centos8 are there any plans to support centos stream with ch-image build --force ?
Or is the recommendation to stick to centos 7 ?

reidpr:
Thanks for the request! Let us look into it.

heasterday:
As a note, it doesn't appear that there are official CentOS Stream images on Dockhub but I was able to test with [this Dockerfile](https://github.com/tgagor/docker-centos-stream/blob/master/Dockerfile).

Looks like the issue is that our current regex match expects the version string to be of the form "x." but CentOS Stream only has a major version. I will whip up a PR to relax our regex to allow matching.

sbiae5:
We have been pulling the stream from quay.io (quay.io/centos/centos:stream8) rather than going back to centos7. If the regex can be updated that would be handy, I did have a quick look at changing it but trying to keep as close to source as possible so students can replicate it easily in future. Thanks

heasterday:
@sbiae5 Could you give the `1095-support-centos-stream` branch a try and see if it works for you?

sbiae5:
All working perfectly so far, thanks for the quick response.
Have just tested with the basics but seems to be working the same as centos8 so should be good.

PR #1096, opened 2021-06-16
lib/fakeroot.py: adjust regex match to account for CentOS Stream

heasterday:
Addresses #1095 

NOTE: I went to add a test for this but since there aren't official images for CentOS Stream that I could find I saw three options:

1. Use a third party base image (I don't like this route)
2. Use a CentOS 8 base image and [convert to Stream](https://www.centos.org/centos-stream/) for testing locally (this requires `--force` support to accomplish so it doesn't seem like an effective test without a more complicated process).
3. Don't add a test 

I went with 3 for now but can chase 2 down if desired.

### DIFF
--- a/lib/fakeroot.py
+++ b/lib/fakeroot.py
@@ -176,7 +176,7 @@ DEFAULT_CONFIGS = {
 
    "rhel8":
    { "name": "CentOS/RHEL 8",
-     "match":  ("/etc/redhat-release", r"release 8\."),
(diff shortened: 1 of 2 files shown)


heasterday:
Looks like Quay.io is the place to get these images per [the wiki](https://wiki.centos.org/FAQ/CentOSStream#What_artifacts_are_built.3F). I will add a test

reidpr:
Reporter said the solution worked for them in the issue.

PR #1098, opened 2021-06-22
Support HTTP status 401 in authentication flow

### LINKED ISSUE #1097: ch-image: some registries return 401 when rejecting anon token requests
On line 1017 of `charliecloud.py`, in `authenticate_bearer()`, we check for status code 403 to see if the anonymous token request was rejected. It turns out that some registries return 401, so accept that too.

nspark:
Add support for HTTP status 401 in the authentication flow attempting anonymous access.

Addresses #1097

### DIFF
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -984,9 +984,9 @@ class Registry_HTTP:
          VERBOSE("won't request anonymous token for %s" % res.request.method)
       else:
          VERBOSE("requesting anonymous auth token")
-         res = self.request_raw("GET", auth_d["realm"], {200,403},
+         res = self.request_raw("GET", auth_d["realm"], {200,401,403},
                                 params=params)
-         if (res.status_code == 403):
+         if (res.status_code in {401,403}):
             VERBOSE("anonymous access rejected")
          else:
             token = res.json()["token"]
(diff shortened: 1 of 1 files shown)


lib/charliecloud.py
```diff
@@ -984,9 +984,9 @@ def authenticate_bearer(self, res, auth_d):
          VERBOSE("won't request anonymous token for %s" % res.request.method)
       else:
          VERBOSE("requesting anonymous auth token")
-         res = self.request_raw("GET", auth_d["realm"], {200,403},
+         res = self.request_raw("GET", auth_d["realm"], {200,401,403},
                                 params=params)
-         if (res.status_code == 403):
+         if (res.status_code in {401,403}):
```
reidpr, line 989:
```suggestion
         if (res.status_code != 200):
```

reidpr (approved):
Nick, this looks great, thank you!

I have one suggestion to fix an existing problem since we're in there. Also I'm going to edit the description slightly to match our workflow (we don't use the GitHub PR/issue linking keywords).

If that suggestion looks good to you, I'll merge.

reidpr:
Once you're happy, mark the PR "ready for review" as we use that to indicate ready for merge.

Issue #1081, opened 2021-05-26
ch-run: auto-mount squashfs


mphinney1100:
Have squashfs as a valid input for `ch-run` to mount, execute and unmount. This is an extension of Charliecloud SI project 2020.

reidpr:
This would replace `ch-mount` and `ch-unmount` by moving the functionality into `ch-run`.

mphinney1100:
- [ ] update `Makefile`
- [ ] fix rpm test cases
- [ ] update documentation
- [ ] finalize new test cases

PR #1102, opened 2021-06-25
ch-run: auto-mount squashfs

mphinney1100:
Addresses #1081.

### DIFF
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -109,7 +109,7 @@ int main(int argc, char *argv[])
    int c_argc;
    char ** c_argv;
 
-   atexit(sqfs_ll_clean);
+   //atexit(sqfs_ll_clean);
    privs_verify_invoking();
 
    verbose = 1;  // in charliecloud.h
@@ -144,6 +144,7 @@ int main(int argc, char *argv[])
    if (!argp_help_fmt_set)
       Z_ (unsetenv("ARGP_HELP_FMT"));
    if(sqfs_ll_check(argv[arg_next], 0)) {
+      Ze (atexit(sqfs_ll_clean), "exit handler set up failed");
       argv[arg_next] = sqfs_mount(sqfs_unpack, argv[arg_next]);
       sqfs_p = true;
       DEBUG("new argv: %s", argv[arg_next]);
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -174,8 +174,7 @@ int main(int argc, char *argv[])
    INFO("private /tmp: %d", args.c.private_tmp);
 
    fix_environment(&args);
-   if (!sqfs_p)
-      containerize(&args.c);
+   containerize(&args.c);
    run_user_command(c_argv, args.initial_dir); // should never return
    exit(EXIT_FAILURE);
 }
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -143,7 +143,7 @@ int main(int argc, char *argv[])
    Z_ (argp_parse(&argp, argc, argv, 0, &arg_next, &args));
    if (!argp_help_fmt_set)
       Z_ (unsetenv("ARGP_HELP_FMT"));
-   if(sqfs_ll_check(argv[arg_next], 0)) {
+   if(sqfs_ll_check(argv[arg_next])) {
       Ze (atexit(sqfs_ll_clean), "exit handler set up failed");
       argv[arg_next] = sqfs_mount(sqfs_unpack, argv[arg_next]);
       sqfs_p = true;
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -97,6 +97,7 @@ void privs_verify_invoking();
 const struct argp argp = { options, parse_opt, args_doc, usage };
 extern char **environ;  // see environ(7)
 char *sqfs_unpack = "/var/tmp";
+bool sqfs_p = false;
 
 /** Main **/
 
@@ -108,6 +109,7 @@ int main(int argc, char *argv[])
    int c_argc;
    char ** c_argv;
 
+   atexit(sqfs_ll_clean);
    privs_verify_invoking();
 
    verbose = 1;  // in charliecloud.h
@@ -143,6 +145,7 @@ int main(int argc, char *argv[])
       Z_ (unsetenv("ARGP_HELP_FMT"));
    if(sqfs_ll_check(argv[arg_next], 0)) {
       argv[arg_next] = sqfs_mount(sqfs_unpack, argv[arg_next]);
+      sqfs_p = true;
       DEBUG("new argv: %s", argv[arg_next]);
    }
 
@@ -170,6 +173,8 @@ int main(int argc, char *argv[])
    INFO("private /tmp: %d", args.c.private_tmp);
 
    fix_environment(&args);
+   if (sqfs_p)
+      sqfs_run_user_command(c_argv, args.initial_dir);
    containerize(&args.c);
    run_user_command(c_argv, args.initial_dir); // should never return
    exit(EXIT_FAILURE);
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -141,7 +141,7 @@ int main(int argc, char *argv[])
    Z_ (argp_parse(&argp, argc, argv, 0, &arg_next, &args));
    if (!argp_help_fmt_set)
       Z_ (unsetenv("ARGP_HELP_FMT"));
-   if(0) {
+   if(sqfs_ll_check(argv[arg_next], 0)) {
       argv[arg_next] = sqfs_mount(sqfs_unpack, argv[arg_next]);
       DEBUG("new argv: %s", argv[arg_next]);
    }
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -13,7 +13,7 @@
 #include "config.h"
 #include "ch_misc.h"
 #include "ch_core.h"
-
+#include "ch_fuse.h"
 
 /** Constants and macros **/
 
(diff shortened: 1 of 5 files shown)


bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+
+/* Function called to start sqfs_ll_clean */
+void kill_fuse()
+{
+   DEBUG("end fuse loop: %d", sq.pid);
+   exit(0);
+}
+
+/* Exit handler for sqfs */
+void sqfs_ll_clean()
```
reidpr, line 70: Don't use the `sqfs_ll` prefix because it implies the functions are in the SquashFUSE library (it's an informal namespace).

bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+{
+   fuse_remove_signal_handlers(sq.chan.session);
+   sqfs_ll_destroy(sq.ll);
+   DEBUG("unmounting: %s", sq.mountdir);
+   sqfs_ll_unmount(&sq.chan, sq.mountdir);
+   Ze(rmdir(sq.mountdir) == -1, "unable to remove directory");
```
reidpr, line 76: The fact that we create and remove the mount point every time makes me a little queasy. What do you think? Is this the right approach?

bin/ch-run.c
```diff
@@ -95,7 +96,7 @@ void privs_verify_invoking();
 
 const struct argp argp = { options, parse_opt, args_doc, usage };
 extern char **environ;  // see environ(7)
-
+char *sqfs_unpack = "/var/tmp";
```
reidpr, line 99: I wonder if the default mount point should be `/var/tmp/$USER/ch-mnt`? Or maybe we should change the default image storage to something like `/var/tmp/$USER.ch/ch-image` and then for this use `/var/tmp/$USER.ch/mnt`? The difference is that `/var/tmp/$USER` isn't Charliecloud specific while `/var/tmp/$USER.ch` is.

bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/* Path is a sqfs*/
+bool sqfs_ll_check(const char *path)
+{
+   sqfs_ll *ll;
```
reidpr, line 82: I wonder if this should just be a global so it doesn't need to be allocated. We just ever have one, right?

bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+{
+   fuse_remove_signal_handlers(sq.chan.session);
+   sqfs_ll_destroy(sq.ll);
+   DEBUG("unmounting: %s", sq.mountdir);
+   sqfs_ll_unmount(&sq.chan, sq.mountdir);
+   Ze(rmdir(sq.mountdir) == -1, "unable to remove directory");
+}
+
+/* Path is a sqfs*/
+bool sqfs_ll_check(const char *path)
```
reidpr, line 80: The name here disagrees a bit with what it does (more than checking).

bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+void sqfs_ll_clean()
+{
+   fuse_remove_signal_handlers(sq.chan.session);
+   sqfs_ll_destroy(sq.ll);
+   DEBUG("unmounting: %s", sq.mountdir);
+   sqfs_ll_unmount(&sq.chan, sq.mountdir);
+   Ze(rmdir(sq.mountdir) == -1, "unable to remove directory");
+}
+
+/* Path is a sqfs*/
```
reidpr, line 79: I recommend more detailed comments for the functions. What does it do? What are the inputs and return value? What errors can happen? Any corner cases? etc.

bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+bool sqfs_ll_check(const char *path)
+{
+   sqfs_ll *ll;
+   sqfs_fd_t fd;
+
+   ll = malloc(sizeof(*ll));
+   Tf (ll, "can't allocate memory");
+   memset(ll, 0 , sizeof(*ll));
+   fd = open(path, O_RDONLY);
+   if(fd != -1 && sqfs_init(&ll->fs, fd, 0) == SQFS_OK)
```
reidpr, line 89: These are two separate errors (can't open file, can't init squashfs) and should be reported separately.

bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+   sqfs_destroy(&ll->fs);
+   free(ll);
+   return false;
+}
+
+/* Run user command with an extra process for sqfs */
+void sqfs_run_user_command(char *argv[], const char *initial_dir)
+{
+   int status;
+   if (fork() == 0)
```
reidpr, line 100: Need to check for `fork(2)` error.

Also, this is an example of a function that definitely needs some careful documentation, because it's got rather quirky behavior: (1) forks and (2) does not return in either the child or parent but (3) for different reasons in each.

Somewhat unrelated, I think you should future-proof this by investigating `prctl(PR_SET_CHILD_SUBREAPER)` and making sure the process that does *not* run the user command (i.e., parent or child) is ready to be the subreaper if we go down that route.

bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+      run_user_command(argv, initial_dir);
+   wait(&status);
```
reidpr, line 102: Should probably respond to the status value somehow.

bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+   kill(sq.pid, SIGINT);
```
reidpr, line 103: The expected signal here is probably SIGTERM (it means "please stop yourself cleanly").

bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+   _Exit(0);
```
reidpr, line 104: Comment why `_Exit(2)` instead of `exit(3)` or returning.

bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+   struct fuse_args args;
+   struct fuse_lowlevel_ops sqfs_ll_ops;
+   args.argc = 1;
+   args.argv = &filepath;
+   args.allocated = 0;
+   get_fuse_ops(&sqfs_ll_ops);
+
+   // mount sqfs
+   sq.ll = sqfs_ll_open(filepath, 0);
+   Te(sq.ll, "%s does not exist", filepath);
```
reidpr, line 128: I like to call these macros with a space (i.e., `Te (...)`) to make more clear that they are not normal functions.

bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+   Ze(opendir(sq.mountdir), "%s already exists", sq.mountdir);
+   Ze(mkdir(sq.mountdir, 0777), "failed to create: %s", sq.mountdir);
+   Te(SQFS_OK == sqfs_ll_mount(&sq.chan, sq.mountdir, &args, &sqfs_ll_ops, sizeof(sqfs_ll_ops), sq.ll), "failed to mount");
+   Ze((sq.chan.session == NULL), "failed to create fuse session");
```
reidpr, line 132: These need to pass through the diagnostics from libsquashfuse if at all possible on failure.

bin/ch_fuse.h
```diff
@@ -0,0 +1,15 @@
+/* Copyright © Triad National Security, LLC, and others.
+   This interface contains Charliecloud's core containerization features. */
```
reidpr, line 2: It doesn't though. ;)

test/run/ch-run_sqfs.bats
```diff
@@ -0,0 +1,21 @@
+load ../common
+
+@test 'ch-run: squash' {
```
reidpr, line 3: I think we probably do not need a separate Bats file but rather should send the whole test suite through your code. We can talk about how to make that happen; I've thought about it a bit but would need to refresh my memory.

reidpr (commented):
Excellent work. Lots of comments in-line.

I think the major thing missing here is some documentation on how it works. It seems important to me to have a clear summary, since reading the code it's hard to see the forest for the trees. It would also help compare what's implemented to your presentation.

I'd like to do the next review (next week?) live.

We could throw this into tidy right now, but let's have you iterate a little more.

Question: Do we want to support SquashFUSE at all if we can't link with libsquashfuse? I.e., are we going to keep the `ch-mount` and `ch-unmount` scripts? Talk with Tim and Shane about this.

bin/ch_fuse.c
```diff
@@ -0,0 +1,157 @@
...
+/** Function prototypes (private) **/
+void get_fuse_ops(struct fuse_lowlevel_ops *sqfs_ll_ops);
+void kill_fuse();
+
+/** Functions **/
+
+/* Initalize fuse loop */
+void fuse_loop_init()
+{
+   Ze((fuse_set_signal_handlers(sq.chan.session) == -1), "failed to set signal handlers");
```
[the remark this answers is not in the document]
mphinney1100: `Te` with -1 doesn't fail which is why its comparing to -1 with `Ze`.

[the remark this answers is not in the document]
reidpr: Aha, so it's a function that returns ≥ 0 on success and –1 on failure? In that case you might consider `Te (... != -1)` or `Te (... >= 0)` to keep the equality and its negation closer together.

bin/ch_fuse.c
mphinney1100: This method is checking if the input path is a sqfs so it should never error. I'm going to clarify that solidify that method a bit more so it more understandable.

bin/ch_fuse.c
mphinney1100: So there isn't any information from  libsquashfuse for the failure besides the two lines above. FUSE does log errors but it looks like there are all set to error so they will all be printed.

reidpr: 👍 If you know how to trigger a failure in `sqfs_ll_mount` then testing the presence of the error message might be a good test.

reidpr (approved):
Great work as always. Several tidy commits pushed. I did only a quick spot test and it seems to still work.

Work remaining I can think of:

1. Tests:
   1. CI for libsquashfuse version.
   2. Specific tests for various libsquashfuse errors.
2. `-w` should fail for SquashFS.
3. Document:
   1. processing structure (parent/child, signals, user command exit value passed through from parent). Maybe one paragraph or so?
   2. libsquashfuse can be opaque about errors; `-vv` enables debug output

PR #1111, opened 2021-06-30
configure: remove bashism from SquashFUSE version check.

olifre:
This fixes `configure` using e.g. `dash` as `SHELL` and `CONFIG_SHELL`. Issue originally reported at:
https://bugs.gentoo.org/799377


reidpr (approved):
LGTM, thanks so much for the PR! Once CI passes I'll merge.

PR #1117, opened 2021-07-07
`ch-run`: honor `$TMPDIR`

sc13-bioinf:
Our HPC setup has a tiny /tmp directory. The creation of charliecloud housekeeping files e.g. copy of /etc/passwd is enough to cause 'no space left on device' errors. This patch allows use of a user specified tmp dir instead.


reidpr:
Hi, welcome to Charliecloud and thanks for the contribution!

What do you think about honoring `$TMPDIR` rather than a custom environment variable `$CH_TMP`? It seems to be [standard](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_03).

Marking this PR draft since that's how we indicate work-in-progress not yet ready for merge.

sc13-bioinf:
Now using $TMPDIR

bin/ch_core.c
```diff
@@ -179,11 +179,20 @@ void enter_udss(struct container *c)
...
+      T_ ( host_tmp = strdup("/tmp") );
+   }
+   else
+   {
+      T_ ( host_tmp = strdup(c->tmp) );
+   }
    if (c->private_tmp) {
       tmpfs_mount("/tmp", c->newroot, NULL);
    } else {
       bind_mount("/tmp", "/tmp", BD_REQUIRED, c->newroot, 0);
```
reidpr, line 193: For your use case, would it also be good to bind-mount the host's `$TMPDIR` at container `/tmp` rather than always using the hosts's `/tmp`?

reidpr:
Also, would it be OK with you if we made a pass on this code to make the style consistent with the rest of the project and add a test?

sc13-bioinf:
Please feel free to change the code however you wish. Thank you for looking at this.

reidpr:
@sc13-bioinf, does it still work for you after my commit?

bin/ch-run.c
```diff
@@ -274,6 +274,9 @@ void fix_environment(struct args *args)
       INFO("new $PATH: %s", new_value);
    }
 
+   // $TMPDIR: Unset.
+   Z_ (unsetenv("TMPDIR"));
```
mphinney1100, line 278: Why are you unsetting $TMPDIR before `containerize()`?

reidpr: The justification is in `ch-run_desc.rst` though that's of course open to critique.

Note this makes the `host_tmp()` function approach fail, so see the latest commit. Sorry for the churn.

reidpr (approved):

Issue #1118, opened 2021-07-07
`ch-image build`: pass through SSH agent


reidpr:
`SSH_AUTH_SOCK` is not currently passed into the build by default, so one must add `ARG SSH_AUTH_SOCK` to the Dockerfile and `--build-arg=SSH_AUTH_SOCK` to the build command line in order for SSH authentication via the agent to work.

We should instead pass through this variable by default, like the proxy variables.

I thought we already did this, but I guess not, or maybe something changed.

Issue #1014, opened 2021-03-24
ch-image build: git pull over SSH can stop build


reidpr:
E.g.:

```
-bash-4.2$ cat ./Dockerfile 
FROM centos:7
RUN git clone git@example.com:foo/bar.git

-bash-4.2$ ch-image build -t reproducer -f ./Dockerfile .
  1 FROM centos:7
  2 RUN ['/bin/sh', '-c', 'git clone git@example.com:foo/bar.git']
Cloning into 'bar'...
The authenticity of host 'example.com (8.67.53.09)' can't be established.
ECDSA key fingerprint is SHA256:[...].
ECDSA key fingerprint is MD5:[...].
Are you sure you want to continue connecting (yes/no)? 
```

And then the build stops because SSH wants to be interactive.

In this case, the user has `example.com` in their `~/.ssh/known_hosts`, but this isn't passed into the build because it's in the home directory but `ch-image` runs `ch-run` with `--no-home`.

reidpr:
I think solutions include turning off host verification in the SSH config or adding the desired host to the system known hosts file (though this is sort of a secret).

PR #1119, opened 2021-07-07
SSH authentication docs & improvement

reidpr:
Addresses #1014 and #1118.

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -763,7 +763,7 @@ This often occurs during an SSH-based Git clone. For example:
   RUN apk add git openssh
   RUN git clone git@github.com:hpc/charliecloud.git
 
-.. code-block::
+.. code-block:: console
 
   $ ch-image build -t foo -f ./Dockerfile .
   [...]
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -824,8 +824,8 @@ Solutions include:
 
        FROM alpine:latest
        RUN apk add git openssh
-       ARG GIT_SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
-       RUN git clone https://github.com/hpc/charliecloud.git
+       ARG GIT_SSH_COMMAND="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
+       RUN git clone git@github.com:hpc/charliecloud.git
 
   5. Add the remote host to the system known hosts file, e.g.:
 
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -378,6 +378,100 @@ password fields give no feedback, not even whether a character has been typed.
 Try using the number row instead, toggling Num Lock key, or SSHing into the
 virtual machine.
 
+SSH stops :code:`ch-image` build with interactive queries
+---------------------------------------------------------
+
+This often occurs during an SSH-based Git clone. For example::
+
+  $ cat ./Dockerfile
+  FROM alpine:latest
+  RUN apk add git openssh
+  RUN git clone git@github.com:hpc/charliecloud.git
+  $ ch-image build -t foo -f ./Dockerfile .
+  [...]
+  3 RUN ['/bin/sh', '-c', 'git clone git@github.com:hpc/charliecloud.git']
+  Cloning into 'charliecloud'...
+  The authenticity of host 'github.com (140.82.113.3)' can't be established.
+  RSA key fingerprint is SHA256:nThbg6kXUpJWGl7E1IGOCspRomTxdCARLviKw6E5SY8.
+  Are you sure you want to continue connecting (yes/no/[fingerprint])?
+
+At this point, the build stops while SSH waits for input.
+
+This happens even if you have :code:`github.com` in your
+:code:`~/.ssh/known_hosts`. This file is not available to the build because
+:code:`ch-image` runs :code:`ch-run` with :code:`--no-home`, so :code:`RUN`
+instructions can't see anything in your home directory.
+
+Solutions include:
+
+  1. Approve the connection interactively by typing :code:`y`. Note this will
+     record details of the connection within the image, including IP address
+     and the fingerprint.
+
+  2. Turn off host key checking in the image's system `SSH config
+     <https://man.openbsd.org/ssh_config>`_. There are multiple ways to do
+     this, but one is::
+
+       $ cat ./Dockerfile
+       FROM alpine:latest
+       RUN apk add git openssh
+       RUN printf 'StrictHostKeyChecking=no\nUserKnownHostsFile=/dev/null\n' \
+           >> /etc/ssh/ssh_config
+       RUN git clone git@github.com:hpc/charliecloud.git
+       $ ch-image build -t foo -f ./Dockerfile .
+       [...]
+         5 RUN ['/bin/sh', '-c', 'git clone git@github.com:hpc/charliecloud.git']
+       Cloning into 'charliecloud'...
+       Warning: Permanently added 'github.com' (RSA) to the list of known hosts.
+       remote: Enumerating objects: 10337, done.
+       remote: Counting objects: 100% (601/601), done.
+       remote: Compressing objects: 100% (299/299), done.
+       remote: Total 10337 (delta 400), reused 441 (delta 297), pack-reused 9736
+       Receiving objects: 100% (10337/10337), 12.77 MiB | 22.31 MiB/s, done.
+       Resolving deltas: 100% (7061/7061), done.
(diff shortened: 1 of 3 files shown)


doc/faq.rst
```diff
@@ -378,6 +378,100 @@ password fields give no feedback, not even whether a character has been typed.
...
+       grown in 4 instructions: foo
+
+     Check your institutional policy on whether this is permissible, though
+     it's worth noting that users `almost never
+     <https://www.usenix.org/system/files/login/articles/105484-Gutmann.pdf>`_
+     verify the host fingerprints anyway.
+
+     This will not record details of the connection in the image.
+
+     More complex edit commands will be needed if either of these directives
```
mej, line 441: As you note, automated editing of the SSH config can get hairy.  Instead of recommending that, I'd replace this 2nd solution with setting `${GIT_SSH}` to something like `"ssh -o 'StrictHostKeyChecking=no' -o 'UserKnownHostsFile=/dev/null'"` to start off with.  You can always mention in a note that adding those settings to an appropriate `Host foo` block in either the system or the user SSH config file is a more robust way of accomplishing the same thing, but this way the detailed solution is something that will "always just work" as opposed to something potentially error-prone.

doc/faq.rst
```diff
@@ -378,6 +378,100 @@ password fields give no feedback, not even whether a character has been typed.
...
+       grown in 4 instructions: foo
+
+     This records connection details in both the Dockerfile and the image.
+
+.. note:: This FAQ does not cover the related issue of SSH authentication. One
+          approach is to simply run the SSH agent on the host.
+          :code:`ch-image` passes environment variable :code:`SSH_AUTH_SOCK`
+          into the build and bind-mounts host :code:`/tmp` to guest
+          :code:`/tmp`, which is where the SSH agent's listening socket
+          usually resides. Thus, SSH within the container will use this
```
mej, line 471: Might want to recommend checking the location (e.g., via `echo ${SSH_AUTH_SOCK%/ssh*/agent*}` or `df ${SSH_AUTH_SOCK}` or similar) and doing a manual host mount if needed?

doc/faq.rst
```diff
@@ -378,6 +378,100 @@ password fields give no feedback, not even whether a character has been typed.
...
+  Are you sure you want to continue connecting (yes/no/[fingerprint])?
+
+At this point, the build stops while SSH waits for input.
+
+This happens even if you have :code:`github.com` in your
+:code:`~/.ssh/known_hosts`. This file is not available to the build because
+:code:`ch-image` runs :code:`ch-run` with :code:`--no-home`, so :code:`RUN`
+instructions can't see anything in your home directory.
+
+Solutions include:
```
mej, line 405: While it could seem like I'm stating the obvious here, it may not be obvious to all end-users...

I would recommend pointing out that most public repositories on most public Git hosting services can be cloned anonymously via HTTPS, so simply changing the URL that's being `git clone`'d might just be the quickest & easiest solution here.

doc/faq.rst
```diff
@@ -378,6 +378,100 @@ password fields give no feedback, not even whether a character has been typed.
...
+
+.. note:: This FAQ does not cover the related issue of SSH authentication. One
+          approach is to simply run the SSH agent on the host.
+          :code:`ch-image` passes environment variable :code:`SSH_AUTH_SOCK`
+          into the build and bind-mounts host :code:`/tmp` to guest
+          :code:`/tmp`, which is where the SSH agent's listening socket
+          usually resides. Thus, SSH within the container will use this
+          existing SSH agent on the host to authenticate without further
+          intervention.
+
```
mej, line 475: I'm not sure how clever/crazy you want this to get, but some other possibilities you may want to consider at least mentioning:
 - Git credentials (see [gitcredentials(7)](https://git-scm.com/docs/gitcredentials) and related subcommand/helper docs)
 - [Personal Access Tokens](https://docs.github.com/en/github/authenticating-to-github/keeping-your-account-and-data-secure/creating-a-personal-access-token) or equivalent
 - `SSH_ASKPASS`, `GIT_ASKPASS`, the GnuPG agent's `pinentry-program`, et al., for out-of-band password/PIN interrogation/caching options
 - Automatic host key fingerprint verification - [SSHFP via DNSSEC](https://weberblog.net/sshfp-authenticate-ssh-fingerprints-via-dnssec/)

There are potentially other options as well (process substitution, socket multiplexing, etc.), depending on how much one might trust the container image, but the above at least should be robust enough to qualify as "safe," IMHO, with varying degrees of simplicity/reliability/dad-proofing....

mej (commented):
I'm doing this as a "Comment" rather than an "Approve" or "Request changes" because I think this could go either way.  It's fine as-is, but I do have some suggestions that I feel are at least worth a couple minutes of mulling.  So I'd rather start a dialogue than draw a conclusion...hope that makes sense! 😺

doc/faq.rst
reidpr: Good point; now issue #1120.

doc/faq.rst
reidpr: Also good points and in issue #1120.

mej (approved):
This is good; I'd just take a 2nd read over that "interactive build" line and see if it can't be wordsmithed to avoid the double-negative.  Other than that, I think you're good to merge!

The only other thing that comes to mind is, it can sometimes be a challenge for beginners to even know what exactly to Google; would it make sense to end the FAQ item with a quick TMTOWTDI note and 1 or 2 suggested Google searches?  Like "automating unattended SSH" or "ssh in cron jobs" or similar?

I'm going to mark "Approved" this time; it's perfectly fine in its current state irrespective of whether you agree with any of the above or decide to address them now or later or never. 😁

doc/faq.rst
```diff
@@ -404,13 +404,26 @@ instructions can't see anything in your home directory.
...
+       RUN apk add git
+       RUN git clone https://github.com/hpc/charliecloud.git
+       $ ch-image build -t foo -f ./Dockerfile .
+       [...]
+       grown in 3 instructions: foo
+
+  2. Approve the connection interactively by typing :code:`y`. Note this will
      record details of the connection within the image, including IP address
-     and the fingerprint.
+     and the fingerprint. It also doesn't make the build non-interactive.
```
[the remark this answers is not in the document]
qwofford: I did get a little confused here. I prefer "Keep in mind that this means building the container will need to remain interactive; you won't be able to fully automate it without defeating it as a security measure,"

qwofford:
This directly addresses confusion I've had regarding interactive prompts during Dockerfile builds using ch-image. 

It indirectly addresses confusion I've had regarding access controlled  git clones (ssh-agent). Thanks for including this in the documentation! One possible issue though...

I thought I had to add a line in my Dockerfile for SSH_AUTH_SOCK?
```
ARG SSH_AUTH_SOCK
```

And also supply:
```
ch-image build --build-arg SSH_AUTH_SOCK ...
```

Am I mistaken about that?

doc/faq.rst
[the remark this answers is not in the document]
reidpr: Challenging for everyone IMO. I'll revise.

reidpr:
> Am I mistaken about that?

No, this accurately describes the current behavior. This PR removes the need to specify `ARG` in the Dockerfile or `--build-arg=SSH_AUTH_SOCK` on the command line.

reidpr:
@qwofford, is this clearer?

qwofford:
Yes, it's clearer. I think there is still potential for confusion since there is a special list of environment vars that are passed along, while the rest are ignored. Something like a:

"Note that most of the user environment is sanitized during a build procedure, the following variables are exceptions to the rule: SSH_AUTH_SOCK, https_proxy, ..."

reidpr:
> following variables are exceptions to the rule

I agree. We have the full list in the `ch-image` man page.

doc/faq.rst
```diff
@@ -707,4 +707,139 @@ Alpine 3.9 image pulled from Docker hub::
...
+  4. Turn off host key checking on the SSH command line. (See caveats in the
+     previous item.) The wrapping tool should provide a way to configure this
+     command line. For example, for Git:
+
+     .. code-block:: docker
+
+       FROM alpine:latest
+       RUN apk add git openssh
+       ARG GIT_SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
+       RUN git clone https://github.com/hpc/charliecloud.git
```
mej, line 828: Just noticed this typo; sorry I missed it before!  It's a `GIT_SSH` example but is cloning via `https`; should read `RUN git clone git@github.com:hpc/charliecloud.git` instead.

heasterday (approved):

Issue #794, opened 2020-06-30
better error message with image tags containing upper-case letters


samuelkgutierrez:
Hi,

I noticed this behavior when using master at 254fa07a2241ea0349179ea42471b1abdefafeea.

Given the following command:
`ch-build -b ch-grow -t CapName -f /home/samuel/devel/bueno/examples/build-n-run/Dockerfile.mpich /home/samuel/devel/bueno/examples/build-n-run`

I see the following error:
```
building with: ch-grow
error: image ref syntax, at end: CapName
```

May I suggest providing a nicer error message that points to the problematic tag name? I don't know if that's really the underlying problem here, but when I provide a tag name without capitalization (anywhere), things work as expected.

Thank you.

reidpr:
Yeah, image names have to be lower case but I agree we could be clearer about it.

FWIW, gory details: https://hpc.github.io/charliecloud/faq.html#how-do-i-specify-an-image-reference

PR #1121, opened 2021-07-09
Add hint option to log

mphinney1100:
Addresses #794.


lib/charliecloud.py
```diff
@@ -1803,7 +1806,10 @@ def log(*args, color=None, prefix="", **kwargs):
                    datetime.datetime.now().isoformat(timespec="milliseconds"),
                    prefix))
    print(prefix, file=log_fp, end="")
-   print(flush=True, file=log_fp, *args, **kwargs)
+   print(flush=True, file=log_fp, *args)
+   hint = kwargs.get('hint', None)
+   if (hint is not None):
+      print("hint: ", hint, file=log_fp)
```
reidpr, line 1812: This will also need `flush=True`.

test/build/40_pull.bats
```diff
@@ -175,31 +175,37 @@ EOF
     # invalid character in image name
     cat <<'EOF' | image_ref_parse 'name*' 1
 error: image ref syntax, char 5: name*
+hint:  https://hpc.github.io/charliecloud/faq.html#how-do-i-specify-an-image-reference
```
reidpr, line 178: I recommend just one space after `hint:` because while it looks tidy here you could also plausibly also use the hint with other log levels.

reidpr (changes requested):
Good work.

I think this reveals a deficiency in the existing design of the logging functions (`INFO()` et al.) That is, I originally envisioned them as being a wrapper around the built-in `print()` and so they should have the same interface. That's why they're all declared with arguments `*args, **kwargs`. However I think they have already (before this PR) diverged from that.

Consider changing `log()` to only take the arguments we actually care about, i.e. remove `*args` and `**kwargs`. (You could probably leave the upper-case wrappers to just pass through everything.) This will let you simplify the code a bit (I think currently there are callers that provide keyword arguments that will mess up the hint) and, if you put `hint` as the first keyword argument, make the log calls with the hint as a positional instead of keyword argument.

This will require auditing all the calls to see if any of them provide more than one object to print or give keyword arguments we pass to `print()`. At the same time you could add hints elsewhere they seem useful (I think there are a few calls with an embedded hint), or create a new issue to do that.

reidpr (approved):
Tidied. Please LMK what I broke.

Issue #1127, opened 2021-07-22
autogen.sh: add check for Python wheel package


shanegoff:
People get an error with make if they don't have the wheel package installed. The error is really confusing and is caused by differences in the output from installing lark without wheel.

PR #1133, opened 2021-07-26
autogen.sh: fail if Lark embed seems broken

reidpr:
Addresses #1127.


j-ogas (approved):
Looks good to me.

autogen.sh
```diff
@@ -74,7 +75,7 @@ rm -rf Makefile \
        configure
 
 if [[ $lark_shovel ]]; then
-    rm -Rfv lib/lark lib/lark-stubs lib/lark*.dist-info
+    rm -Rfv lib/lark lib/lark-stubs lib/lark*.dist-info lib/lark*.egg-info
 fi
```
j-ogas, line 79: Question: why do we avoid removing it by default? What makes the inverse, e.g., remove by default unless specified otherwise more appealing?

reidpr: It has to download the packages from the internet, which I didn't want to do every time `autogen.sh` is run.

shanegoff (approved):
This looks pretty straight forward and should help give people useful feedback when running into the "you need wheel" problem

Issue #1000, opened 2021-03-09
ch-image build: crashes if no image name


reidpr:
This should be an error, not a traceback:

```
$ cp examples/hello/Dockerfile /tmp/dockertest
$ bin/ch-image build -f /tmp/dockertest /tmp
[...]
TypeError: can only concatenate str (not "NoneType") to str
```

reidpr:
Increasing priority because this is kind of a brown paper bag bug. I hope it should be easy to fix.

PR #1140, opened 2021-08-06
lib/build.by: update name inference

heasterday:
Addresses #1000 

Error output using original reproducer:
```
ch-build -b ch-image -f dockertest /tmp
building with: ch-image
error: no image tag provided and unable to infer one
```


reidpr (changes requested):
Nice two-liner!

I think there's a couple documentation improvements that should be added somewhere.

1. Do we document how the inference works? If so, add that.
2. This really could use a hint; something like "use -t, or see: [url to inference docs]".
3. I know Docker calls image names tags, but I sure think of them as names. Are we consistent?

heasterday:
> Nice two-liner!
> 
> I think there's a couple documentation improvements that should be added somewhere.
> 
>     1. Do we document how the inference works? If so, add that.
> 
>     2. This really could use a hint; something like "use -t, or see: [url to inference docs]". This could go on a comment for easy pickup when in PR [Add hint option to log #1121](https://github.com/hpc/charliecloud/pull/1121).
> 
>     3. I know Docker calls image names tags, but I sure think of them as names. Are we consistent?

1. It looks like there is information under the docs for`ch-image`'s `build` subcommand [here](https://hpc.github.io/charliecloud/command-usage.html#build) which states:
>Name of image to create. If not specified, use the final component of path CONTEXT. Append :latest if no colon present. 

  Which seem lacking imo, I will add to it.
2. That's a great idea
3. Perusing through our docs it seems like we have quite a few statements of the form "image name (tag)" but a few where just "image name" is used, I will tidy some.

reidpr (commented):
Good work. Tidied. Haven't tested much so please give it a workout. Let me know what you think.

Seems short on testing, especially for errors (e.g. what if the extension is *only* bad chars?). However there was no testing before either so it's not worse.

lib/build.py
```diff
@@ -78,13 +78,28 @@ def main(cli_):
...
+         cli.tag = base_all
+         ch.VERBOSE("inferring name from Dockerfile basename: %s" % cli.tag)
+      elif (os.path.abspath(cli.context) != "/"):
+         cli.tag = os.path.basename(os.path.abspath(cli.context))
+         ch.VERBOSE("inferring name from context directory: %s" % cli.tag)
+      else:
+         assert (os.path.abspath(cli.context) == "/")
+         cli.tag = "root"
+         ch.VERBOSE("inferring name with root context directory: %s" % cli.tag)
+      cli.tag = re.sub(r"[^a-z0-9_.-]", "", cli.tag.lower())
```
heasterday, line 101: Currently `:` is stripped so that Dockerfile.foo:v3.5 results in an inferred tag of `foov3.5` is this desired?

reidpr: Good catch. Probably not. The name and tag are colon separated which is why I didn't put it in. Might add a comment saying why the regex doesn't match our documented image reference syntax.

heasterday:
@reidpr Outside of the comment I left this looks good to me.

lib/build.py
reidpr: Issue #1167.

Issue #1057, opened 2021-05-11
ch-image build --force: don't enforce maximum version


reidpr:
@loveshack points out in #1043:

> I think restricting the releases supported for --force is a mistake, as it needlessly stops things working. It already will reject Debian unstable and Ubuntu 21.04, for instance, and other Debian derivatives which would just work. It seems unlikely that the support will stop working in distributions any time soon and if it does, users are no worse off. (You could presumably ensure fakeroot support at least doesn't disappear without warning from a distribution with a maintained package by testing the recipe for the native system in the package build.)

This issue is to investigate any downsides or excessive difficulty in not enforcing maximum version for the `--force` configurations, and if it's reasonable, implement it.

See also #862.

PR #1141, opened 2021-08-06
lib/fakeroot.py: don't enforce a max distro version on --force

heasterday:
Addresses #1057 

NOTE: this is a proof of concept and is not final

@reidpr Please let me know your thoughts on this approach offline.


reidpr (commented):
Discussed offline.

PR #1142, opened 2021-08-09
packaging: fedora: update upstream spec

j-ogas:
Update `upstream.spec` with the live changes to https://src.fedoraproject.org/rpms/charliecloud/blob/rawhide/f/charliecloud.spec from #1012.


reidpr (approved):

reidpr:
Approved/merging but I feel like there's got to be a better way to maintain this log.

Issue #1039, opened 2021-04-22
COPY: no error if some sources don't exist


reidpr:
e.g.:

```
$ touch /tmp/doesexist
$ rm -f /tmp/doesnotexist
$ printf 'FROM alpine:3.9\nCOPY doesexist doesnotexist /\n' | ch-image build -t foo -f - /tmp
  1 FROM alpine:3.9
  2 COPY ['doesexist', 'doesnotexist'] -> '/'
grown in 2 instructions: foo
```

This should fail, complaining that `doesnotexist` doesn't exist.

PR #1144, opened 2021-08-09
COPY: error if some sources not found

reidpr:
Addresses #1039.

### DIFF
--- a/lib/build.py
+++ b/lib/build.py
@@ -519,11 +519,13 @@ class I_copy(Instruction):
       # Expand source wildcards.
       srcs = list()
       for src in self.srcs:
-         for i in glob.glob("%s/%s" % (context, src)):  # glob can't take Path
+         matches = glob.glob("%s/%s" % (context, src))  # glob can't take Path
+         if (len(matches) == 0):
+            ch.FATAL("can't copy: not found: %s" % src)
+         for i in matches:
             srcs.append(i)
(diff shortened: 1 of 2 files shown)


lib/build.py
```diff
@@ -519,11 +519,13 @@ def execute_(self):
...
-         for i in glob.glob("%s/%s" % (context, src)):  # glob can't take Path
+         matches = glob.glob("%s/%s" % (context, src))  # glob can't take Path
+         if (len(matches) == 0):
+            ch.FATAL("can't copy: not found: %s" % src)
+         for i in matches:
             srcs.append(i)
             ch.VERBOSE("source: %s" % i)
-      if (len(srcs) == 0):
-         ch.FATAL("can't COPY: no sources found")
+      assert (len(srcs) > 0)
```
j-ogas, line 528: Do we think the extra line of code saved is worth the hairy traceback output from an assertion error? FWIW, I prefer the FATAL message to a messy traceback.

j-ogas (changes requested):

lib/build.py
reidpr: In this case it really is a "can't happen" situation — it should have been caught by the test on line 523, but I felt nervous about it.

I do see now that if `self.srcs` has length zero then `srcs` will as well. Maybe that is why I felt nervous. Thanks for pointing it out.

j-ogas (approved):

Issue #962, opened 2021-01-20
ch-image: version storage directory?


reidpr:
Maybe we should warn if the storage directory was created by a previous version of Charliecloud. We don't track changes to it carefully so it's quite plausible something will get messed up.

Probably should be done after #961 because we could recommend a `ch-image reset` (or whatever the command ends up being) as a fix.

Issue #1146, opened 2021-08-09
charliecloud.mkdirs: change to two arguments


reidpr:
Currently `charliecloud.mkdirs` takes one argument: the sequence of directories to create if they don't exist (just like `mkdir -p`). However, in most cases only part of this path do we really want to create if it doesn't exist. For example, if it's something within an image, only directories within the image should be created; the rest of the path should exist.

The function could be changed to take two arguments: the portion of the path that should already exist, and the portion that should be created if not present.

I looked into this while addressing #962, but it was a little more refactoring than I wanted to deal with at the time.

Some of the existing calls could perhaps be changed to `mkdir()` also.

PR #1148, opened 2021-08-10
add storage directory versioning

reidpr:
Addresses #962.

Note this just adds a warning to existing behavior. There's already no compatibility between different Charliecloud versions; we just don't tell anyone.

### DIFF
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -23,6 +23,8 @@ import tarfile
 import time
 import types
 
+import version
+
 
 ## Imports not in standard library ##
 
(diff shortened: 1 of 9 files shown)


trandles-lanl:
I'm confused by this in general. The "versioning" being discussed is the version of _Charliecloud_ that created the storage directory. It's not a version of the storage directory itself. It sounds like from the documentation change and the OP that this PR isn't changing existing behavior or fixing any problem...so...why?

trandles-lanl (approved):
I like the new wording. It's less ambiguous and doesn't imply that the user should simply always start by doing `ch-image reset` even when it's not required.

lib/charliecloud.py
```diff
@@ -1480,6 +1485,33 @@ def root_env():
...
+            FATAL("storage directory seems invalid: %s" % self.root)
+         if (os.path.isfile(self.version_file)):
+            v_found = int(file_read_all(self.version_file))
+         else:
+            v_found = 1
+      if (v_found == STORAGE_VERSION):
+         VERBOSE("found storage dir v%d: %s" % (STORAGE_VERSION, self.root))
+      elif (v_found in {None, 1}):  # initialize/upgrade
+         INFO("%s storage directory: v%d %s" % (op, STORAGE_VERSION, self.root))
+         mkdirs(self.root)  # FIXME: use mkdir() when don't need two levels
```
j-ogas, line 1506: Can we not make this `mkdirs()` `mkdir()` now?

j-ogas (changes requested):

lib/charliecloud.py
reidpr: No — because the default is `/var/tmp/$USER/ch-image` it will usually have to create both `$USER` and `ch-image`. Though it won't actually help if we change it to `/var/tmp/$USER.ch/ch-image`.

reidpr: See #1146 though.

mphinney1100 (approved):

j-ogas (approved):

Issue #1043, opened 2021-04-27
ch-image build --force: support Debian 11


reidpr:

reidpr:
Waiting on #1042.

PR #1164, opened 2021-08-26
Add support for debian 11

dschwoerer:
Not a nice solution, but it works.

I haven't tested, but wouldn't it be easier to test for debian/ubuntu, and just let it fail if it doesn't work? The user can still remove `--force` and try without, so they are not worse of than now?

Addresses #1043

### DIFF
--- a/lib/fakeroot.py
+++ b/lib/fakeroot.py
@@ -213,8 +213,8 @@ DEFAULT_CONFIGS = {
    #      | egrep '^(fakeroot|fakeroot-ng|pseudo)$'
 
    "debderiv":
-   { "name": "Debian (9, 10) or Ubuntu (16, 18, 20)",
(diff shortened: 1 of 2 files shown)


heasterday:
@dschwoerer This looks correct to me, would you be willing to adjust the tests to include Debian 11 and not test 10? We generally test the minimum supported version and the maximum supported version. This would just be changing the Debian 10 section -> 11 and updating the release name. The relevant sections are [here](https://github.com/hpc/charliecloud/blob/master/test/force-auto.py.in#L252) and [here](https://github.com/hpc/charliecloud/blob/master/test/force-auto.py.in#L282).

dschwoerer:
I have updated the tests :+1:

heasterday (approved):

heasterday:
If CI passes this looks good to merge.

PR #1165, opened 2021-09-01
lib/charliecloud.py bump max support lark version

heasterday:
This should resolve CI failures. Another route would be to pin the lark version that we install in the CI environment but it sounds like we are hoping to only test with the bundled version of lark in the near future.


reidpr:
New Lark 0.12.0 came out a couple days ago: https://github.com/lark-parser/lark/releases/tag/0.12.0

PR #1166, opened 2021-09-03
lib/fakeroot.py: --force, don't enforce max supported version

heasterday:
Addresses #1057 

This PR removes the max supported version logic from `--force`. Instead, `fakeroot.py` checks that the target os version isn't a known unsupported version. In addition, I adjusted the relevant tests to test image releases tagged `latest`. I also changed the Ubuntu tests to use image tags based on release numbers instead of their release names (20.04 instead of focal).

### DIFF
--- a/lib/fakeroot.py
+++ b/lib/fakeroot.py
@@ -175,8 +175,8 @@ DEFAULT_CONFIGS = {
      "each": ["fakeroot"] },
 
    "rhel8":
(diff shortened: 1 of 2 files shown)


reidpr (commented):
LGTM, thanks!

I'll wait to merge until after 0.25.

reidpr (approved):

Issue #1139, opened 2021-08-03
examples: add distroless


reidpr:
Distroless seems popular and it would be straightforward to add an example.

https://github.com/GoogleContainerTools/distroless

PR #1168, opened 2021-09-14
examples: add Distroless

heasterday:
Addresses #1139


heasterday:
Blocked by #1170

examples/distroless/Dockerfile
```diff
@@ -0,0 +1,8 @@
+# ch-test-scope: standard
+# Distroless is a Google project providing slim images that contain runtime
+# dependencies only. https://github.com/GoogleContainerTools/distroless
+FROM gcr.io/distroless/python3
```
j-ogas, line 4: Of the distroless tags what reason, if any, do we pick the python3 one? Perhaps comment on our choice here.

j-ogas (changes requested):

examples/distroless/Dockerfile
heasterday: I chose the`python3` image because of all of their base images it is the simplest to test. I will add a comment to this effect.

examples/distroless/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# ch-test-scope: standard
+# Distroless is a Google project providing slim images that contain runtime
+# dependencies only. https://github.com/GoogleContainerTools/distroless
+# The python3 image was chosen for ease of testing.
+FROM gcr.io/distroless/python3
+COPY . /distroless
+
+# Without the CMD directive ch-builder2tar will fail when using docker
+CMD ["/distroless/hello.py"]
```
reidpr, line 9: Interestingly once PR #698 is merged, this will become no longer needed.

examples/distroless/test.bats
```diff
@@ -0,0 +1,17 @@
+true
+# shellcheck disable=SC2034
+CH_TEST_TAG=$ch_test_tag
+
```
reidpr, line 4: I think these four lines are not needed? They aren't used and the test title prints out properly.

examples/distroless/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# ch-test-scope: standard
+# Distroless is a Google project providing slim images that contain runtime
+# dependencies only. https://github.com/GoogleContainerTools/distroless
+# The python3 image was chosen for ease of testing.
+FROM gcr.io/distroless/python3
+COPY . /distroless
```
reidpr, line 6: I suggest (a) copy only `hello.py` since that's all that's used, and (2) put it at the root directory.

reidpr (changes requested):
Nice work, a few minor suggestions.

examples/distroless/Dockerfile
heasterday: Would you like me to block this PR pending #698 ?

examples/distroless/test.bats
heasterday: Every other `test.bats` I looked at has these lines, is this something that should by tidied up outside of this specific test?

examples/distroless/Dockerfile
heasterday: This was done to be consistent with the `hello` and `mpihello` tests. Is this something that you would like corrected in the other tests?

examples/distroless/Dockerfile
reidpr: No, that one will take a while.

examples/distroless/test.bats
reidpr: Probably. Some might actually need it, or maybe it's completely obsolete.

examples/distroless/Dockerfile
reidpr: Some of the others do put multiple files in. I would leave the others alone for now.

examples/distroless/test.bats
heasterday: I'll create a separate issue to track this.

heasterday: It looks like the `true` may not be needed. Without line 3 there is this error:  
```
 ✗ NO_TAG_SET/hello
   (in test file /home/heasterday/charliecloud/examples/distroless/test.bats, line 13)
     `[[ $status -eq 0 ]]' failed
   ch-run[657842]: can't find image: /var/tmp/images/NO_TAG_SET: No such file or directory (ch-run.c:146 2)
   ```  
And line 2 suppresses this shellcheck warning
```
   CH_TEST_TAG=$ch_test_tag
^---------^ SC2034: CH_TEST_TAG appears unused. Verify use (or export if used externally).
```

reidpr: Bizarre; it must be some `ch-test` magic I've forgotten about.

The `true` is needed: without it, the `shellcheck disable` applies to the whole file, not just the subsequent line.

Maybe these gotchas should be documented in the contributors' guide?

heasterday: Added back the `true`. Documenting the process for adding test does indeed sound like a good idea.

reidpr (approved):

PR #1174, opened 2021-09-15
fix progress meter

reidpr:
In PR #1121 I broke the progress meter: it put all updates on a new line rather than same-line updates if stdout is a TTY. This PR restores the desired behavior.


heasterday (approved):
Tested default behavior and redirecting output to a file (looks like the former should output progress to one line and the latter to multiple lines) and this works on my dev system. Approved.

PR #1181, opened 2021-09-18
`ch_misc.h`: mark `host_tmp` as extern to fix linking errors

olifre:
Without marking this as `extern`, `ld` fails with:
```
ld: ch_run-ch_core.o:DIR/bin/ch_misc.h:70: multiple definition of `host_tmp'; ch_run-ch-run.o:DIR/bin/ch_misc.h:70: first defined here
ld: ch_run-ch_misc.o:DIR/bin/ch_misc.h:70: multiple definition of `host_tmp'; ch_run-ch-run.o:DIR/bin/ch_misc.h:70: first defined here
collect2: error: ld returned 1 exit status
```

### DIFF
--- a/bin/ch_misc.h
+++ b/bin/ch_misc.h
@@ -67,7 +67,7 @@
 /** External variables **/
 
 extern int verbose;
-char * host_tmp;
+extern char * host_tmp;
 
 
 /** Function prototypes **/


reidpr:
Interesting, I think you're right but it links fine for me (and CI). What compiler are you using?

olifre:
This was with `gcc 10.3.0-r2 p3` on Gentoo, but I tried just now on Debian Bullseye which ships `gcc 10.2.1-6` and can also reproduce it there — so probably, GCC 10 and higher make this appear.

reidpr (approved):

PR #1188, opened 2021-09-29
rstudio server test

### LINKED ISSUE #588: add Rstudio Server demo/example
As a follow-up to #569, this seems like a cool example/demo. See that issue for some background and example Dockerfiles.

@jhawe, I recall we discussed on the phone that you had a simple R script that would demonstrate things are working OK (e.g., simple computation and plot)? Is that something you can share?

rstyd:
This PR addresses #588. The only issue currently is that the first test freezes up. Need to find a way to run the server in the background without disrupting BATS.


examples/rstudio/compare.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+percent_diff=$(compare -metric MSE "$1" "$2" null: 2>&1 > /dev/null | awk '{print $1}')
```
reidpr, line 2: This needs an explanation. What is `compare`? Does it introduce new dependencies, which should probably be added to `configure`? Why `-metric MSE`? What is the selected threshold (is it really zero, or rounded)? (If it's really zero, maybe we want `-metric AE`.) What is `awk` pulling out? Is it really a percentage (docs suggest the range is 0–1)? What is `null:`? Why are we discarding stderr?

examples/rstudio/compare.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
reidpr, line 1: Maybe this should be a function in `common.bash` instead of a stand-alone script?

examples/rstudio/compare.sh
```diff
@@ -0,0 +1,7 @@
...
+percent_diff=$(compare -metric MSE "$1" "$2" null: 2>&1 > /dev/null | awk '{print $1}')
+if [[ $percent_diff == 0 ]];then
```
reidpr, line 3:
```suggestion
if [[ $percent_diff = '0' ]];then
```

to emphasize that it's a string compare, not numeric compare. Also the shell comparison operator is a single equals (`==` is a Bash extension and will work but the rest of the code uses `=`).

examples/rstudio/compare.sh
```diff
@@ -0,0 +1,7 @@
...
+    echo "Images are equal"
+else
+    echo "Images don't match"
+fi
```
reidpr, line 7: Use exit code with no output, like `grep -q`.

examples/rstudio/model.R
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env Rscript
```
reidpr, line 1: Add a couple of sentences comment saying what this example does.

examples/rstudio/rstudio_auth
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
reidpr, line 1: Why use `bash` from the path instead of the usual `#!/bin/bash`?

examples/rstudio/rstudio_auth
```diff
@@ -0,0 +1,24 @@
...
+# rstudio_auth method from https://github.com/nickjer/singularity-rstudio
+
+# Confirm username is supplied
+if [[ $# -ne 1 ]]; then
+  exit 1
+fi
+USERNAME="${1}"
```
reidpr, line 8: Does this need to be interactive? Can we use some other method to ensure the caller is part of the test? I'd think we don't actually need to authenticate for the purposes of this test, since we're just passing around dummy passwords stored elsewhere in the test.

examples/rstudio/start_rstudio
```diff
@@ -0,0 +1,20 @@
...
+  echo "You must specify a port. Suggested range: 8000-9000"
+  exit 1
+fi
+
+RSTUDIO_AUTH="/bin/rstudio_auth"
+
+/usr/lib/rstudio-server/bin/rserver \
+  --www-port="${port}" \
+  --auth-none=0 \
+  --auth-pam-helper-path=${RSTUDIO_AUTH} \
```
reidpr, line 19: What are the implications of running an open Rstudio server during the test? Looks like it's listening to all interfaces too.

examples/rstudio/rstudio_auth
```diff
@@ -0,0 +1,24 @@
...
+
+# Read in the password from user
+read -s -p "Password: " PASSWORD
+echo ""
+
+if [[ ${USERNAME} == ${USER} && ${PASSWORD} == ${RSTUDIO_PASSWORD} ]]; then
+  exit 0
+else
+  exit 1
+fi
```
reidpr, line 23: Don't need an if statement for this, or so many brackets.
```suggestion
[[ $USERNAME = "$USER" && $PASSWORD = "$RSTUDIO_PASSWORD" ]] || exit 1
```

examples/rstudio/start_rstudio
```diff
@@ -0,0 +1,20 @@
...
+#password=$(openssl rand -base64 16)
+# Just use a basic password but in practice we'd want to create a random one like above
+password=charliecloud
+export RSTUDIO_PASSWORD=${password}
+
+port=${1}
+if [[ -z ${port} ]]; then
+  echo "You must specify a port. Suggested range: 8000-9000"
+  exit 1
+fi
```
reidpr, line 12: Why do we need a configurable port?

examples/rstudio/start_rstudio
```diff
@@ -0,0 +1,20 @@
...
+
+RSTUDIO_AUTH="/bin/rstudio_auth"
```
reidpr, line 14: I'd just put the path directly below, since this is the only place the variable is used.

examples/rstudio/test.bats
```diff
@@ -0,0 +1,32 @@
+true
+# shellcheck disable=SC2034
+CH_TEST_TAG=$ch_test_tag
+
+load "${CHTEST_DIR}/common.bash"
+
+setup () {
+    scope full
```
reidpr, line 8: I believe you can set the scope using keywords in the Dockerfile.

examples/rstudio/test.bats
```diff
@@ -0,0 +1,32 @@
...
+    prerequisites_ok "rstudio"
+}
+
+@test "${ch_tag}/test rstudio-server" {
+    # Start up on port 8991
+    run ch-run "$ch_img" -- start_rstudio 8991 &
```
reidpr, line 14: Why port 8991?

examples/rstudio/test.bats
```diff
@@ -0,0 +1,32 @@
...
+    rstudio_server=$!
+    run ch-run "$ch_img" -- python3 test_rstudio.py
+    echo $output
+    [[ $status -eq 0 ]]
+    [[ $output = *'Good'* ]] 
+    kill -9 $rstudio_server
```
reidpr, line 20: 1. Don't `kill -9` unless plain `kill` doesn't work, because it won't let the process clean up.
2. How will the next test work if the Rstudio server has been stopped?

examples/rstudio/test.bats
```diff
@@ -0,0 +1,32 @@
...
+}
+
+
+# Test R installation 
+@test "${ch_tag}/test R installation" {
+    ch-run -b "$BATS_TMPDIR":/mnt "$ch_img" -- bash -c "Rscript model.R /mnt/plot.png"
+    run ch-run -b "$BATS_TMPDIR":/mnt/0 -b "$CHTEST_EXAMPLES_DIR"/rstudio:/mnt/1 \
+          "$ch_img" -- sh -c "./compare.sh /mnt/0/plot.png /mnt/1/plot.png"
```
reidpr, line 28: What are the subshells doing here? Why is one bash and the other sh?

examples/rstudio/test_rstudio.py
```diff
@@ -0,0 +1,33 @@
+import getpass
+import time
+import selenium
+from selenium import webdriver
+from selenium.webdriver import FirefoxOptions
+from selenium.webdriver.common.keys import Keys
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.support import expected_conditions as EC
```
reidpr, line 9: Our style is to `import` rather than `from ... import`. You can `import ... as ...` if you need a shorthand.

examples/rstudio/test_rstudio.py
```diff
@@ -0,0 +1,33 @@
...
+
+opts = FirefoxOptions()
+opts.add_argument("--headless")
+driver = webdriver.Firefox(log_path='/tmp/geckodriver.log',firefox_options=opts)
```
reidpr, line 13: Put the log in `$BATS_TMPDIR` so the filesystems stays tidy and different test runs don't clash.

examples/rstudio/test_rstudio.py
```diff
@@ -0,0 +1,33 @@
+import getpass
```
reidpr, line 1: Is there a reason not to make this script executable with a shebang line?

Also it could use some comments to explain what is going on. There's nothing else like it in the Charliecloud test suite, so the audience won't be familiar.

examples/rstudio/test_rstudio.py
```diff
@@ -0,0 +1,33 @@
...
+elem = driver.find_element_by_name("username")
+elem.clear()
+user = getpass.getuser()
+elem.send_keys(user)
+
+elem = driver.find_element_by_name("password")
+elem.clear()
+password = "charliecloud"
+elem.send_keys(password)
+elem.send_keys(Keys.RETURN)
```
reidpr, line 25: What does this line do?

examples/rstudio/test_rstudio.py
```diff
@@ -0,0 +1,33 @@
...
+
+driver.find_element_by_xpath("//button[text()='Sign In']").click()
+time.sleep(30)
```
reidpr, line 28: Why 30 seconds?

examples/rstudio/test_rstudio.py
```diff
@@ -0,0 +1,33 @@
...
+try:
+    elem = driver.find_element_by_id("rstudio_container")
+    print('Good')
+except selenium.common.exceptions.NoSuchElementException:
+    print('Bad')
```
reidpr, line 33: I recommend output that's more descriptive, with the success/failure in the exit code.

examples/rstudio/test.bats
```diff
@@ -0,0 +1,32 @@
...
+CH_TEST_TAG=$ch_test_tag
+
+load "${CHTEST_DIR}/common.bash"
+
+setup () {
+    scope full
+    prerequisites_ok "rstudio"
+}
+
+@test "${ch_tag}/test rstudio-server" {
```
reidpr, line 12: Looks like this just tests whether it's possible to log into Rstudio? Can you actually compute something? In either case, a comment explaining what's being tested would be helpful.

examples/rstudio/model.R
```diff
@@ -0,0 +1,12 @@
...
+
+args <- commandArgs(trailingOnly = TRUE)
+# Load vehicle data
+data(cars)
+
+# Create linear model
+lm <- lm(dist~speed, cars)
+
+# plot linear model results
+png(args[1])
```
reidpr, line 11: I think you could just hard-code the output path.

reidpr (changes requested):
Thanks Rusty. This indeed a very cool new example.

I see the tests fail on `cloc`, so that will need to be fixed.

One general comment. This test needs to be documented; it's very different from what we have already. I would suggest a README in `examples/rstudio`. E.g.: what is Rstudio, what is selenium, what are the principles the test is built on and summary of how does it work.

As for your question about the first test hanging, I don't see any fundamental problems. I think you just need to figure out the right way to shut down Rstudio and do that, possibly waiting until it's stopped. The Rstudio docs should help. See the Spark example for how it's done for that daemon.

examples/rstudio/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM rocker/verse:3.6.1
```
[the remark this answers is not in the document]
rstyd: This image was created by the Rocker community R docker image effort along with r-base. These images are intended to be more version stable for better reproducibility compared to the r-base image. Looking back at their page [here](https://hub.docker.com/r/rocker/verse) I think I want to use their `rstudio` image instead since this `verse` one is all about tex and publishing which we don't care about in this example. 

Initially, I tried using`r-base` but ran into some issues. I also tried using an ubuntu base and then these instructions [here](https://www.rstudio.com/products/rstudio/download-server/debian-ubuntu/), but that also didn't quite work. Now, that it _should_ be working correctly I can give those another shot.

examples/rstudio/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+     python3 \
+     python3-pip \
+     firefox-esr \ 
+     imagemagick \
+     wget 
+
+
+RUN pip3 install selenium
+
+# Need to install newest version of geckodriver for selenium
```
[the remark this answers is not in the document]
rstyd: The base image we're currently using is Debian which has no `geckodriver` package. Ubuntu does have one.

examples/rstudio/test.bats
rstyd: Yes. The test just checks if we can login in to rstudio. My initial plan was to use selenium to perform a computation in rstudio after logging in by putting some code into the code editor then clicking the run button. The way rstudio is written made it basically impossible to get the frame that the code editor was in with selenium. It appears they just used a transpiler from java to javascript which renders everything a mess. I figured that since rstudio uses the local R installation as a backend testing that installation and that a user could access rstudio itself is sufficient. I'll add a comment.

examples/rstudio/compare.sh
rstyd: I'll add a comment and add that to the readme. Compare is a tool that's part of the image magick suite. I install image magick inside the container then do the comparison in the container with the output images to avoid requiring image magick as a new dependency. Although, image magick does come standard in most distros. So, it probably wouldn't affect many users if we set it as a dependency and ran this outside of the container. What do you prefer?

I chose MSE pretty arbitrarily. AE does look like a better fit. `awk` is just pulling out the MSE value. Image magick returns `0 (0)` but we only want the first value.

examples/rstudio/rstudio_auth
rstyd: All the issues with this file are related to me not altering it from the source. :)

examples/rstudio/rstudio_auth
rstyd: It doesn't, and as far as I can tell there's no way for the user to interact with the interactive parts. This script is essentially a kludge that gets around having to setup some authentication rstudio requires. I believe the important part is just that it exits with `0` to rstudio. I'll set it to just `exit 0` always and not check the password. The server being on a particular port ensures the caller is part of the test.

examples/rstudio/test.bats
rstyd: I chose it arbitrarily. As far as I can tell, no application generally use it. I had tried writing a script that looks for an unused port and uses that, but it's not really possible to do that in bash without race conditions since you can't access the socket interface directly. We could use `8080` which I believe is the canonical port for something like this but a lot of applications use it as their default port.

examples/rstudio/test.bats
rstyd: 1. Ah! Good point. 
2. The next test is independent from the rstudio server. It tests the R installation on the system that Rstudio uses. I explain it further [here](https://github.com/hpc/charliecloud/pull/1188/files/3f26830644f93a6aa59284818d2f1ba18ef481d2#r721523859).

examples/rstudio/compare.sh
rstyd: Changed the implementation so we don't need this script anymore.

examples/rstudio/test.bats
rstyd: That was an oversight.

examples/rstudio/Dockerfile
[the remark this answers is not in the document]
reidpr: Yes, please add a brief comment to this effect.

The comment above says the `rstudio` image seems better but the code is still `verse`?

examples/rstudio/Dockerfile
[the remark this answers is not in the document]
reidpr: OK. Please add a comment to this effect. I think I was confused by “newest”, when based on your comment it sounds like it’s just not available as a distro package.

examples/rstudio/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+     firefox-esr \ 
+     imagemagick \
+     strace \
+     wget 
+
+RUN pip3 install selenium
+
+# Need to install newest version of geckodriver for selenium
+ARG GECKODRIVER_VERSION=v0.30.0
+RUN wget https://github.com/mozilla/geckodriver/releases/download/$GECKODRIVER_VERSION/geckodriver-v0.30.0-linux32.tar.gz
```
reidpr, line 16: Use `wget -nv` per the style guide, to reduce chatter in the build.

examples/rstudio/rstudio_auth
reidpr: OK, what is the source? Are we able to re-distribute it? This information should be commented.

examples/rstudio/rstudio_auth
```diff
@@ -0,0 +1,25 @@
...
+#
+## Read in the password from user
+#read -s -p "Password: " PASSWORD
+#echo ""
+#
+#if [[ ${USERNAME} == ${USER} && ${PASSWORD} == ${RSTUDIO_PASSWORD} ]]; then
+#  exit 0
+#else
+#  exit 1
+#fi
```
reidpr, line 24: Why is the whole script commented out?

examples/rstudio/rstudio_auth
```diff
@@ -0,0 +1,25 @@
...
+exit 0
```
reidpr, line 25: Need an explanation somewhere on how this script works and why it's just exiting successfully. (And if that's all we need, we could just specify `/bin/true` instead.)

examples/rstudio/run.py
```diff
@@ -0,0 +1,4 @@
+import subprocess
+# Running rstudio as a background process with & causes BATS to hang
+# Forking it as a subprocess instead avoids this issue
+subprocess.Popen(['/rstudio/start_rstudio', '8991'])
```
reidpr, line 4: I can try to figure out what is going on here. I'd prefer not to have a Python indirection.

examples/rstudio/test.bats
reidpr: I don't see a comment?

examples/rstudio/run.py
rstyd: That'd be great thanks!

examples/rstudio/rstudio_auth
rstyd: You had discussed previously just turning off authentication. I thought having access to how that might work could be beneficial to users, but it's probably better to just use bin/true and provide a link in a comment to the PAM helper script from the singularity example. https://github.com/nickjer/singularity-rstudio/blob/master/rstudio_auth.sh

examples/rstudio/Dockerfile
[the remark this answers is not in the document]
rstyd: Thanks! I've updated it.

reidpr (commented):
I worked on this for a while but wasn't able to make it work to my satisfaction and felt I ran out of time for it.

The current problem is that `rserver` writes the PID file but then exits. There is no error message, and I couldn't see how to generate any debug output. Previously it started up fine, with what seemed to me exactly the same start procedure.

I'm not sure where to go from here.

reidpr:
I'd love to get this in, but this PR is getting stale. Closing to reduce clutter in PR list. Re-open or replace if/when appropriate.

PR #1192, opened 2021-10-11
ch-image build: no context on stdin

### LINKED ISSUE #769: ch-image build: implement no context if Dockerfile on stdin

mphinney1100:
Closes #769.


reidpr (changes requested):
Thanks! I tweaked the docs. Please take a look and verify this matches what we talked about and what Docker does. If so, update the implementation and tests to match. If not, let's discuss.

reidpr (approved):
Great work. Much smaller and tidier solution than I expected!

I pushed a small tidy commit. If that looks OK LMK and I'll merge.

Issue #1184, opened 2021-09-28
ch-image pull fails for some images


dschwoerer:
Trying to pull a specific docker image fails - while the same call works fine with e.g. podman
```
👎 ch-image pull docker.io/oi4ai/emc3-test:test02
pulling image:    docker.io/oi4ai/emc3-test:test02
requesting arch:  amd64
error: not in registry: docker.io:443/oi4ai/emc3-test:test02
👎 podman pull docker.io/oi4ai/emc3-test:test02
Trying to pull docker.io/oi4ai/emc3-test:test02...
Getting image source signatures
Copying blob 9a1ae74008d5 skipped: already exists  
Copying blob 1cfaf5c6f756 [--------------------------------------] 0.0b / 0.0b
Copying config 7331783432 done  
Writing manifest to image destination
Storing signatures
733178343237731211ec02bab68fa55ac0667c0efdacbba6af7a5bac0b87ad79
```

```
👍 bin/ch-image pull oi4ai/emc3-test:test02 -v
```
<details>
```
verbose level: 1
host architecture from uname: x86_64
host architecture for registry: amd64
found storage dir v2: /var/tmp/dave/ch-image
pulling image:    oi4ai/emc3-test:test02
requesting arch:  amd64
destination:      /var/tmp/dave/ch-image/img/oi4ai%emc3-test:test02
downloading image: oi4ai/emc3-test:test02
initializing session
auth: no authorization
GET: https://registry-1.docker.io:443/v2/oi4ai/emc3-test/manifests/test02
response status: 401
HTTP 401 unauthorized
authorizing
WWW-Authenticate raw: Bearer realm="https://auth.docker.io/token",service="registry.docker.io",scope="repository:oi4ai/emc3-test:pull"
WWW-Authenticate parsed: Bearer {'realm': 'https://auth.docker.io/token', 'service': 'registry.docker.io', 'scope': 'repository:oi4ai/emc3-test:pull'}
authenticating using Bearer
requesting anonymous auth token
GET: https://auth.docker.io/token
response status: 200
received auth token: eyJhbGciOiJSUzI1NiIsInR5cCI6IkpX
retrying with auth: Bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpX
GET: https://registry-1.docker.io:443/v2/oi4ai/emc3-test/manifests/test02
response status: 404
RateLimit-Limit: 100;w=21600
RateLimit-Remaining: 93;w=21600
[...]

PR #1195, opened 2021-10-13
accept OCI manifest content types

reidpr:
Addresses #1184.

### DIFF
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -191,10 +191,15 @@ HTTP_CHUNK_SIZE = 256 * 1024
 PYTHON_MIN = (3,6)
 
 # Content types for some stuff we care about.
-TYPE_MANIFEST = "application/vnd.docker.distribution.manifest.v2+json"
-TYPE_MANIFEST_LIST = "application/vnd.docker.distribution.manifest.list.v2+json"
-TYPE_CONFIG =   "application/vnd.docker.container.image.v1+json"
-TYPE_LAYER =    "application/vnd.docker.image.rootfs.diff.tar.gzip"
+# See: https://github.com/opencontainers/image-spec/blob/main/media-types.md
+TYPES_MANIFEST = \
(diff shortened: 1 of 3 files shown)


reidpr:
@dschwoerer, can you try this branch and see if it solves #1184 for you?

lib/charliecloud.py
```diff
@@ -1365,7 +1372,7 @@ def manifest_upload(self, manifest):
       INFO("manifest: uploading")
       url = self._url_of("manifests", self.ref.tag)
       self.request("PUT", url, {201}, data=manifest,
-                   headers={ "Content-Type": TYPE_MANIFEST })
+                   headers={ "Content-Type": TYPES_MANIFEST["docker"] })
```
j-ogas, line 1375: Why "docker" instead of something like "v2"?

lib/charliecloud.py
```diff
@@ -1352,8 +1358,9 @@ def manifest_to_file(self, path, msg, digest=None):
          digest = "sha256:" + digest
       url = self._url_of("manifests", digest)
       pw = Progress_Writer(path, msg)
+      accept = "%s;q=0.5" % ",".join(TYPES_MANIFEST.values())
```
j-ogas, line 1361: What is the `q=0.5` magic?

test/Dockerfile.ocimanifest
```diff
@@ -0,0 +1,18 @@
+# This Dockerfile is used to test image with an OCI manifest (issue #1184). It
+# must be built with Podman; version 3.0.1 works.
```
j-ogas, line 2: Do we have version protection for this test? What happens when a non-`3.0.1` version is tested?

j-ogas (changes requested):

lib/charliecloud.py
reidpr: It has to do with `Accept:` weighting. I hesitate to touch it since I think I would have put it in for a good reason, but I agree it should be documented. New issue #1196.

PR #1207, opened 2021-10-18
ch-test: don't use misc/version

### LINKED ISSUE #678: tests: ch-test fails when git isn't available
`ch-test` calls `misc/version` which uses git to check the build version during all scopes leading to failure on systems where git isn't available. This was discovered on a system that had git available on the host where charliecloud's source was built but not on the backend nodes where it was ran.

This appears at odds with what is documented within `/misc/version`
```
# This script should usually be run once, by Autotools ...
```

j-ogas:
Closes #678.

### DIFF
--- a/bin/ch-test
+++ b/bin/ch-test
@@ -900,10 +900,9 @@ if [[ -z $CHTEST_INSTALLED ]]; then
     if [[ $ch_version != "$cf_version" ]]; then
         warning "inconsistent configure version: ${cf_version}"
     fi
-    src_version=$("${ch_base}/misc/version" 2> /dev/null || echo '')
-    if [[ $src_version = '' ]]; then
-       src_version=$(cat "${ch_base}/VERSION")
-    fi
+    # Charliecloud version. Prefer git; otherwise use simple version.
+    src_version=$(   "${ch_base}/misc/version" 2> /dev/null \
+                  || cat "${ch_base}/VERSION")
     if [[ $ch_version != "$src_version" ]]; then
         warning "inconsistent source version: ${src_version}"
     fi
--- a/bin/ch-test
+++ b/bin/ch-test
@@ -900,7 +900,7 @@ if [[ -z $CHTEST_INSTALLED ]]; then
     if [[ $ch_version != "$cf_version" ]]; then
         warning "inconsistent configure version: ${cf_version}"
     fi
-    src_version=$("${ch_base}/misc/version")
+    src_version=$(cat ./VERSION)
     if [[ $ch_version != "$src_version" ]]; then
         warning "inconsistent source version: ${src_version}"
     fi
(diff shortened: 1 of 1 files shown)


bin/ch-test
```diff
@@ -900,7 +900,7 @@ if [[ -z $CHTEST_INSTALLED ]]; then
     if [[ $ch_version != "$cf_version" ]]; then
         warning "inconsistent configure version: ${cf_version}"
     fi
-    src_version=$("${ch_base}/misc/version")
+    src_version=$(cat ./VERSION)
```
j-ogas, line 903: Unclear whether this new comparison will always pass now, making this if statement useless. @reidpr, @heasterday, is checking the VERSION file a suitable alternative for the git usage in `./misc/version`?

reidpr (commented):
Can you remind me what the goal is here? `VERSION` doesn't include the whole version number except for released versions, so I hesitate to relax the test. But I'm sure there's a different way to solve your problem.

bin/ch-test
heasterday: I'm still familiarizing myself with the original check. As a note though I believe you want `cat ../VERSION` or `cat 
"${ch_base}/VERSION"`as this file lives up a directory from `ch-test` itself.

reidpr (changes requested):
I think it's valuable to have the Git information if it's available, so I don't want to just exclude it. I would be OK with falling back to the non-Git simple version if the Git info isn't available.

Looks like like `misc/version` currently handles the case where the source is not a Git working directory.

I'd suggest a few options:

1. Modify `ch-test` to only call `misc/version` if Git is available.
2. Add a switch to `misc/version` that returns the simple version if Git isn't available, and keep the current default behavior of silently returning the simple version if not a Git working directory. This would mean you still can't run `autogen.sh` if the source is a Git working directory but Git is not installed (currently this will crash just like `ch-test`).
3. Add a switch to `misc/version` that returns the simple version if *either* Git isn't available or the source isn't a Git working directory; if the switch isn't provided and one of those conditions is true, error out. This would mean you can't run `autogen.sh` if Git isn't installed, which seems bad.
4. Modify `misc/version` so it returns the simple version if *either* Git isn't available or the source isn't a Git WC (without any new switch). This would mean `autogen.sh` silently drops Git info from the version number if run from a Git WD but Git isn't installed.

reidpr (commented):
Discussed offline.

bin/ch-test
```diff
@@ -900,7 +900,10 @@ if [[ -z $CHTEST_INSTALLED ]]; then
     if [[ $ch_version != "$cf_version" ]]; then
         warning "inconsistent configure version: ${cf_version}"
     fi
-    src_version=$("${ch_base}/misc/version")
+    src_version=$("${ch_base}/misc/version" 2> /dev/null || echo '')
+    if [[ $src_version = '' ]]; then
+       src_version=$(cat "${ch_base}/VERSION")
+    fi
```
reidpr, line 906: I'd recommend checking for the availability of Git directly, if that's what you're after. As written, this uses `VERSION` no matter why `misc/version` failed.

Maybe that's what you want? If so, then `|| true` would be more idiomatic, and a commend explaining why it's what you want would be good.

reidpr (changes requested):

PR #1215, opened 2021-10-26
ch-image: add history metadata

### LINKED ISSUE #1002: ch-image push: fails for Quay.io
Pushing an image to Quay.io causes HTTP 500 while uploading manifest:

```
$ ch-image push 00_tiny quay.io/charliecloud/test:latest
pushing image:   00_tiny
destination:     quay.io/charliecloud/test:latest
layer 1/1: gathering
layer 1/1: preparing
preparing metadata
starting upload
layer 1/1: 0f73d3d: checking if already in repository
layer 1/1: 0f73d3d: already present
config: 0db8e4b: checking if already in repository
config: 0db8e4b: not present, uploading

Username: reidpr
Password:
manifest: uploading
error: PUT failed; expected status {201, 401} but got 500: Internal Server Error
```

j-ogas:
Closes #1002.


test/build/50_ch-image.bats
```diff
@@ -396,16 +396,69 @@ ch_foo=foo-ev
...
+    },
+    {
+      "created": "2021-11-08T17:55:01Z",
+      "created_by": "SHELL ['/bin/sh', '-v', '-c']",
+      "empty_layer": "True"
+    },
+    {
+      "created": "2021-11-08T17:55:01Z",
+      "created_by": "RUN ['/bin/sh', '-v', '-c', 'echo \"shell2: $0\"']",
+      "empty_layer": "True"
```
reidpr, line 459: Shouldn't the last one be `"empty_layer": "False"`?

test/build/50_ch-image.bats
```diff
@@ -396,16 +396,69 @@ ch_foo=foo-ev
...
     "PATH": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
     "ch_bar": "bar-ev",
     "ch_baz": "baz-ev",
     "ch_foo": "foo-ev"
   },
+  "history": [
+    {
+      "created": "2021-11-08T17:55:00Z",
+      "created_by": "FROM charliecloud/metadata:2021-01-15",
+      "empty_layer": "True"
```
reidpr, line 414: Do we want the string "True" or the boolean value true?

reidpr (commented):
Good progress. Some comments in-line. Two general notes:

* Use case 1: We still flatten the image, so don't we need to modify the image so we don't get a layer count mismatch?

* Rejection 3: Charliecloud treats a multi-stage image as multiple independent builds that happen to be in the same Dockerfile. So I don't think there's a special case here.

test/build/50_ch-image.bats
j-ogas: I haven't found a good way to tell if `image_i` is the last image. So far we've been deferring to `push.py` to append that final non-empty entry.

j-ogas: Good catch.

reidpr:
FWIW at present this PR is incompatible with pre-PR storage directories:

```
$ ch-test -b ch-image --pack-fmt=tar-unpack all
[...]
executing build phase tests ...
 ✗ ch-build 00_tiny
   (in test file /tmp/ch-test.tmp.reidpr/build_auto.bats, line 13)
     `ch-build $force -t 00_tiny --file="[...]/charliecloud/test/Dockerfile.00_tiny" "[...]/charliecloud/test"' failed
   building with: ch-image
   warning: not yet supported, ignored: issue #777: .dockerignore file
     5 FROM alpine:3.9
   warning: --force not available (no suitable config found)
     File "[...]/charliecloud/bin/ch-image", line 240, in <module>
       main()
     File "[...]/charliecloud/bin/ch-image", line 234, in main
       cli.func(cli)
     File "[...]/charliecloud/bin/../lib/charliecloud/build.py", line 181, in main
       ml.visit_topdown(tree)
     File "[...]/charliecloud/bin/../lib/charliecloud/lark/visitors.py", line 269, in visit_topdown
       self._call_userfunc(subtree)
     File "[...]/charliecloud/bin/../lib/charliecloud/lark/visitors.py", line 241, in _call_userfunc
       return getattr(self, tree.data, self.__default__)(tree)
     File "[...]/charliecloud/bin/../lib/charliecloud/build.py", line 224, in __default__
       inst.metadata_history_update(self.instruction_ct)
     File "[...]/charliecloud/bin/../lib/charliecloud/build.py", line 278, in metadata_history_update
       hist = images[image_i].metadata_history_return()
     File "[...]/charliecloud/bin/../lib/charliecloud/charliecloud.py", line 490, in metadata_history_return
       hist = self.metadata["history"]
   KeyError: 'history'
```

This should be fixed. I think likely it can be made to work with no storage format version bump.

lib/build.py
```diff
@@ -221,6 +221,7 @@ def __default__(self, tree):
                ch.FATAL("first instruction must be ARG or FROM")
          inst.execute()
          if (image_i != -1):
+            inst.metadata_history_update(self.instruction_ct)
```
reidpr, line 224: This would be better in `execute()` I think. Then it wouldn't need to care about the instruction index.

lib/build.py
```diff
@@ -265,6 +266,19 @@ def execute(self):
...
+      images[image_i].metadata_append(hist)
+
+   def metadata_history_remove(self):
+      images[image_i].metadata_history_remove()
+
+   def metadata_history_update(self, inst_ct):
+     hist = images[image_i].metadata_history_return()
+     images[image_i].metadata_history_replace(hist)
+     self.metadata_history_add(inst_ct)
+
```
reidpr, line 281: These can probably be eliminated in favor of calling the `Image` methods directly.

lib/charliecloud.py
```diff
@@ -446,6 +447,18 @@ def set_(dst_key, *src_keys):
...
+      # true.
+      try:
+         hist = config["history"]
+      except AttributeError:
+         FATAL("config missing key 'history'")
+      hist = self.metadata_empty_layer_history(hist)
+      try:
+         hist[-1].pop("empty_layer")
+      except (IndexError, KeyError):
+         pass
```
reidpr, line 460: This can probably be eliminated by leaving the `empty_layer` workaround until push time.

lib/charliecloud.py
```diff
@@ -446,6 +447,18 @@ def set_(dst_key, *src_keys):
             except AttributeError:
                FATAL("can't parse config: bad Env line: %s" % line)
             self.metadata["env"][k] = v
+      # History. Set the empty_layer value of each non-final history entry to
+      # true.
+      try:
+         hist = config["history"]
```
reidpr, line 453: I'd suggest an explicit translation step here. I.e., do we really want all fields in each history entry from the config? Which ones do we care about and which do we not?

lib/charliecloud.py
```diff
@@ -456,6 +469,31 @@ def set_(dst_key, *src_keys):
          for k in config["config"]["Volumes"].keys():
             self.metadata["volumes"].append(k)
 
+   def metadata_append(self, data):
+      DEBUG("appending metadata: %s" % data)
+      for k in data:
+         if (k in self.metadata.keys()):
+            self.metadata[k] += data[k]
+         else:
+            FATAL("config has no key '%s'" % k)
```
reidpr, line 478: Not clear to me what this function does. Looks like a dictionary merge of some kind? If so, there's idiomatic ways to do it w/o defining a function.

Also, if the translation from OCI config moves to pull, this won't need to care about config keys being present.

lib/charliecloud.py
```diff
@@ -456,6 +469,31 @@ def set_(dst_key, *src_keys):
...
+
+   def metadata_empty_layer_history(self, hist):
+      if (len(hist) > 0 ):
+         for i in range(len(hist)):
+            hist[i]["empty_layer"] = True
+      return hist
```
reidpr, line 484: This translation to OCI config should happen in `push.py` I'd think.

lib/charliecloud.py
```diff
@@ -456,6 +469,31 @@ def set_(dst_key, *src_keys):
...
+
+   def metadata_history_replace(self, hist):
+      self.metadata["history"] = hist
```
reidpr, line 487: Why a function instead of just assigning the dictionary key?

lib/charliecloud.py
```diff
@@ -456,6 +469,31 @@ def set_(dst_key, *src_keys):
...
+
+   def metadata_history_return(self):
+      "Return config history with all entires having empty_layer set to True"
+      try:
+         hist = self.metadata["history"]
+      except KeyError:
+         return []
```
reidpr, line 494: I'm seeing a lot of these. I would recommend moving the workaround for missing history to when `metadata.json` is read from disk. Coupled with the translate on pull/push, then all the internal history manipulation code can assume the data structures are consistent.

lib/build.py
```diff
@@ -265,6 +266,19 @@ def execute(self):
    def execute_(self):
       ...
 
+   def metadata_history_add(self, inst_ct):
+      hist = { "history": [ { "created": ch.now_utc_iso8601(),
+                              "created_by": "%s %s" % (self.str_name(), self.str_())}]}
+      images[image_i].metadata_append(hist)
```
reidpr, line 272: I'd recommend that the function that adds a history entry take the two? history fields directly, with names chosen by you (IMO `created` and `created_by` are confusing) e.g. `history_append(when, what)`.

lib/push.py
```diff
@@ -114,6 +114,8 @@ def prepare(self):
                                      "digest": "sha256:" + hash_c })
       # Prepare metadata.
       ch.INFO("preparing metadata")
+      self.image.metadata_load()
+      config["history"] = self.image.metadata["history"]
```
reidpr, line 118: This should IMO become an explicit translation as well.

reidpr (changes requested):
Nice progress and thanks for your patience on my review.

The main suggestion I have is to figure out a history format that's convenient to use, independent of the OCI config, and use that format internally. Factors to consider include (1) what's important to us and (2) what's needed to generate an OCI config that satisfies the registries. Then, translate from OCI config to our format at pull time, and back to OCI config at push time. This will eliminate the need for anything other than the translation code to care about stuff like the `empty_layer` hack.

lib/charliecloud.py
j-ogas: @heasterday argues that something generic like `FROM TAG` in leou of the base image's actual history isn't very helpful; the tag TAG referenced could be from a locally built image and unavailable for inspection from a registry like Dockerhub.

Determining what is useful and what isn't seems strange for the base image. I personally prefer to replace it with something simple like `created_by: FROM TAG` OR pass all the base image history unedited aside from the `empty_layer` hack.

lib/charliecloud.py
```diff
@@ -446,6 +447,12 @@ def set_(dst_key, *src_keys):
             except AttributeError:
                FATAL("can't parse config: bad Env line: %s" % line)
             self.metadata["env"][k] = v
+      # History.
+      try:
+         hist = config["history"]
+      except AttributeError:
+         hist = []
```
reidpr, line 454: 1. Wouldn't this be `KeyError`? Regardless I don't personally like the "ask forgiveness" idiom of Python, preferring instead the "ask permission" model. The reasoning is that it can be tricky to know exactly where somewhat generic exceptions like `KeyError` come from, so you might be responding to a different problem than you expected (though not in this case I don't think).

2. Does the exception ever happen? Can we treat a config missing `history` key as malformed?

lib/build.py
```diff
@@ -260,6 +258,11 @@ def announce(self):
    def execute(self):
       if (not cli.dry_run):
          self.execute_()
+      hist = [ { "created": ch.now_utc_iso8601(),
+                 "created_by": "%s %s" % (self.str_name(), self.str_())}]
+      if (image_i != -1):
+         images[image_i].metadata_history_append(hist)
+         images[image_i].metadata_save()
```
reidpr, line 265: Should this be under the `if`?

lib/charliecloud.py
```diff
@@ -456,6 +463,11 @@ def set_(dst_key, *src_keys):
          for k in config["config"]["Volumes"].keys():
             self.metadata["volumes"].append(k)
 
+   def metadata_history_append(self, hist):
+      DEBUG("appending history: %s" % hist)
+      for i in hist:
+         self.metadata['history'].append(i)
```
reidpr, line 469: Couple suggestions here:

1. Call it `history_append`?
2. Why does it take a list of items? It's only called once with a singleton.
3. I'd suggest more structured arguments, rather than arbitrary objects appended to `self.metadata["history"]`. E.g., the call I see could just pass the string that ends up in `created_by`.

lib/push.py
```diff
@@ -114,6 +114,15 @@ def prepare(self):
...
       ch.INFO("preparing metadata")
+      self.image.metadata_load()
+      hist = self.image.metadata["history"]
+      for i in range(len(hist)):
+         hist[i]["empty_layer"] = True
+      try:
+         hist[-1].pop("empty_layer")
+      except IndexError:
+         ch.FATAL("metadata: missing history")
+      config["history"] = hist
```
reidpr, line 125: Would be good I think to document what we're doing with history. For example, it looks like we're using the OCI format. Why? What alternatives were considered and rejected?

reidpr (changes requested):
Good progress. Some food for thought in-line.

lib/charliecloud.py
j-ogas: > Call it history_append?

I don't really like the huge name `metadata_history_append` either, however, it allows us to group it alphabetically in charliecloud.py with other related metadata functions.

> Why does it take a list of items? It's only called once with a singleton.

Good catch. I'll update it to take a single dictionary.

reidpr (approved):
Great, thanks. If the tidy looks OK to you and the tests pass, I'll merge.

lib/push.py
```diff
@@ -116,12 +116,8 @@ def prepare(self):
       ch.INFO("preparing metadata")
       self.image.metadata_load()
       hist = self.image.metadata["history"]
-      for i in range(len(hist)):
+      for i in range(len(hist) - 1):
```
j-ogas, line 119: If the last history entry happens to have the `empty_layer` present then we can hit a corner case where will have a layerless image. IIRC the nginx image history has a final entry with an empty layer. So if we built an image with something silly like:

```
FROM nginix
```

We may bump into this issue. That is why I popped the last entry instead of iterating to n -1. Perhaps there is a better way to handle this?

j-ogas (commented):
Looks good except for the layerless edge case. See comment.

reidpr (commented):
Looks like there's a comment that needs to be resolved.

lib/push.py
```diff
@@ -114,6 +114,32 @@ def prepare(self):
...
+      #
+      # Thus, to push images built (or pulled) with Charliecloud we ensure the
+      # the total number of non-empty layers always totals one (1). To do this
+      # we iterate over the history entires backward searching for the first
+      # non-empty entry and preserve it; all others are set to empty.
+      non_empty_winner = None
+      for i in range(len(hist) - 1, -1, -1):
+         if (   "empty_layer" not in hist[i].keys()
+             or (    "empty_layer" in hist[i].keys()
+                 and hist[i]["empty_layer"] == 'false')):
```
reidpr, line 135: Is it really a string? Below it's a boolean.

```suggestion
                 and not hist[i]["empty_layer"])):
```

reidpr (commented):
One question in-line.

heasterday:
@j-ogas This looks good to me. I would like to test it but the infrastructure that I was using to spin up the test Quay instance appears to be down. If you are alright with waiting until next week to merge I can test then.

reidpr:
Probably worth waiting IMO since pushing to Quay was the key motivation.

test/build/50_ch-image.bats
```diff
@@ -406,6 +406,146 @@ EOF
...
+    {
+      "author": "charlie@example.com",
+      "created": "2021-01-16T00:12:12.868439691Z",
+      "created_by": "/bin/echo world",
+      "empty_layer": true
+    },
+    {
+      "author": "charlie@example.com",
+      "created": "2021-01-16T00:12:13.055783024Z",
+      "created_by": "/bin/ash -c #(nop)  SHELL [/bin/ash -c]",
```
heasterday, line 481: I'm not familiar with `/bin/ash` what shell is this?

test/build/50_ch-image.bats
```diff
@@ -406,6 +406,146 @@ EOF
...
+      "created_by": "/bin/ash -c #(nop)  VOLUME [/mnt/foo /mnt/bar /mnt/foo]",
+      "empty_layer": true
+    },
+    {
+      "created": "2021-11-30T20:40:24Z",
+      "created_by": "FROM charliecloud/metadata:2021-01-15"
+    },
+    {
+      "created": "2021-11-30T20:40:24Z",
+      "created_by": "RUN ['/bin/ash', '-c', 'echo \"cwd1: $PWD\"']"
```
heasterday, line 514: While it doesn't appear to cause issues I notice a difference between the `created_by` lines between Docker and Charliecloud:
Docker follows a format like this:
`/bin/sh -c  #(nop) RUN echo hello`
Charliecloud:
`"RUN ['bin/sh', '-c', 'echo hello']"`.  

This difference is intended, correct?

heasterday:
This appears to resolve the issue but I do have a couple minor questions above.

test/build/50_ch-image.bats
reidpr: Yes.

test/build/50_ch-image.bats
reidpr: It's a bare-bones shell aimed at POSIX compliance with not much else. https://en.wikipedia.org/wiki/Almquist_shell

heasterday (approved):

Issue #1221, opened 2021-11-02
`examples/spack`: test `spec` fails


reidpr:
Test `spec` in `examples/spack/test.bats` currently fails:

```
 ✗ spack/spec
   (in test file /tmp/ch-test.tmp.reidpr/spack_example.bats, line 30)
     `ch-run "$ch_img" -- spack spec hdf5' failed with status 3
   Input spec
   --------------------------------
   hdf5

   Concretized
   --------------------------------
   ==> Bootstrapping clingo from pre-built binaries
   ==> Bootstrapping clingo from sources
   ==> Installing libiconv-1.16-eixrlsf3gm5drzsyfh64zkjrjojfh52y
[...]
   ==> Error: cannot bootstrap the "clingo" Python module from spec "clingo-bootstrap@spack+python %gcc target=x86_64" due to the following failures:
       'spack-install' raised InstallError: Terminating after first install failure: OSError: [Errno 30] Read-only file system: '/spack/var/spack/cache'
       Please run `spack -d spec zlib` for more verbose error messages
```

It's trying to install things during `spack spec`. This is apparently because it's trying to [bootstrap the clingo concretizer](https://spack.io/changes-spack-v017/).

Adding `spack install clingo` to the Dockerfile did not help.

reidpr:
Steps to reproduce error above w/o running the test suite:

```
$ cd examples/spack
$ ch-build -t spack .
$ ch-convert -i docker spack /var/tmp/spack
$ ch-run /var/tmp/spack -- /spack/bin/spack spec hdf5
```

reidpr:
@sly clarifies that we don't need to *install* clingo but rather *bootstrap* it. This happens during `spack spec` in the Dockerfile, but it gets bootstrapped into root's home directory:

```
$ ls /var/tmp/spack/root/.spack
bootstrap  cache  reports
```

However, redirecting Spack to the “correct” home directory still fails:

```
$ ch-run /var/tmp/spack -- sh -c 'HOME=/root /spack/bin/spack spec hdf5'
Input spec
--------------------------------
hdf5

Concretized
--------------------------------
==> Error: Error writing to config file: '[Errno 30] Read-only file system: '/root/.spack/bootstrap/config/linux/bootstrap.yaml''
```

Issue #193, opened 2018-06-29
re-enable Spack tests on Cray


reidpr:
`examples/other/spack/test.bats` is disabled on Cray hosts because of spack/spack#8618. Once that bug is fixed, re-enable it.

PR #1224, opened 2021-11-05
re-enable Spack tests on Cray

rstyd:
This PR addresses #193. Spack test works on cray again. There are some other issues with #1221, but it doesn't affect cray machines individually.


reidpr (approved):
LGTM, thank you. Tests don't pass but I think that's because it needs some fixes in PR #1220, so I'll wait to merge until that's in master.

Issue #731, opened 2020-04-30
ch-build: add --verbose


reidpr:
I keep trying to use this, expecting it will show the builder commands it is running.

reidpr:
Mooted by PR #1225.

PR #1225, opened 2021-11-05
deprecate `ch-build`

reidpr:
Per discussion, this PR deprecates `ch-build` and puts a proposed alternative shell function in the FAQ.

It would moot #731.


shanegoff (commented):
I won't miss ch-build. I expect we will drop the ch-build tests at a later date?

shanegoff (approved):

Issue #1190, opened 2021-09-29
`ch-image`: exception during auth prompt causes traceback


reidpr:
`ch-image` crashes with a traceback if there is an exception when prompting for authentication, e.g. if you press Control-C instead of giving a username. It should instead fail with an error message.

PR #1227, opened 2021-11-09
ch-image: handle keyboard interrupt

j-ogas:
Addresses #1190.


lib/charliecloud.py
```diff
@@ -259,7 +259,11 @@ def get(self):
          except KeyError:
             # Finally, prompt the user.
             # FIXME: This hangs in Bats despite sys.stdin.isatty() == True.
-            username = input("\nUsername: ")
+            try:
+               username = input("\nUsername: ")
+            except KeyboardInterrupt:
+               INFO('')
+               FATAL("keyboard interrupt")
```
reidpr, line 266: I'd recommend wording that's more user-facing, e.g. "cancelled" or something like that.

lib/charliecloud.py
```diff
@@ -259,7 +259,11 @@ def get(self):
          except KeyError:
             # Finally, prompt the user.
             # FIXME: This hangs in Bats despite sys.stdin.isatty() == True.
-            username = input("\nUsername: ")
+            try:
+               username = input("\nUsername: ")
+            except KeyboardInterrupt:
+               INFO('')
```
reidpr, line 265: Is this because without the blank message, the error message is on the same line as the prompt? I think you can remove it, because (1) it's very common for control-C to generate weirdly aligned error messages and (2) the mixed level (INFO vs. FATAL) is jarring and seems brittle.

reidpr (changes requested):
Thanks! Couple suggestions in-line.

reidpr (approved):

Issue #1059, opened 2021-05-11
C code: multiple representations of lists


reidpr:
We currently have a mix of sentinel-terminated arrays and unterminated arrays with a second length variable. I'm not sure if this is bad or not.

Possibly we need a generic list data structure?

Being able to initialize our lists with literals is important.

reidpr:
Review of the code yields (roughly):

| location | variable | terminator | contents | notes |
| --- | --- | --- | --- | --- |
| `ch-run.c:44` | `options` | 0 | array of `struct argp_option` | required by `argp` API |
| `ch-run.c:77` | `args.env_delta` | `ENV_END` = 0 | array of `struct env_delta` | has datatype-specific list functions |
| `ch-run.c:178` | `c_argc`, `c_argv` | count | array of strings | `execvp(3)` takes null-terminated `argv` and I don't see how that happens |
| `ch_core.c:42` | `BINDS_DEFAULT` | `NULL` | array of `struct bind` | not pointers despite `NULL` terminator |
| `ch_core.c:70` | `bind_mount_paths` | count | array of strings | |
| `ch_core.c:136` | `bind_mounts()` args | `NULL` in `.src` | array of `struct bind` | || `ch_misc.c:28` | `VERBOSE_LEVELS` | none | array of strings | never iterated |
| `ch_misc.c:68` | `supp_gids` | none | fixed-length array of `gid_t` | 0 is valid entry, required by `getgroups(2)` |
| `ch_misc.c:95` | `mkdirs()` args | count | array of strings | didn't audit callers |

reidpr:
My initial guess is that we could standardize on zero-terminated arrays and write some functions that treat the contents as opaque buffers, `realloc(3)` as needed. This would require:

1. Lists must be on the heap if modified.
2. Value zero not allowed in list contents.

PR #1230, opened 2021-11-09
standardize C lists

reidpr:
Addresses #1059.

### DIFF
--- a/bin/ch_fuse.c
+++ b/bin/ch_fuse.c
@@ -121,7 +121,7 @@ void sq_fork(struct container *c)
       T_ (asprintf(&subdir, "/%s.ch/mnt", username) > 0);
       c->newroot = cat("/var/tmp", subdir);
       INFO("using default mount point: %s", c->newroot);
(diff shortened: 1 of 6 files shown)


bin/ch_fuse.c
```diff
@@ -119,7 +119,7 @@ void sq_fork(struct container *c)
       T_ (asprintf(&subdir, "/%s.ch/mnt", username) > 0);
       c->newroot = cat("/var/tmp", subdir);
       INFO("using default mount point: %s", c->newroot);
-      mkdirs("/var/tmp", subdir, NULL, 0);
+      mkdirs("/var/tmp", subdir, NULL);
```
j-ogas, line 122: Seems like the other files have more `NULL`s converted to `0` this PR. Why is this one opposite?

j-ogas (approved):
LGTM

bin/ch_fuse.c
reidpr: This one was and is a null pointer. The zero that was removed was the count of items.

PR #1233, opened 2021-11-10
CI: fix variable

j-ogas:


reidpr (changes requested):
Doesn't pass CI. I think it will if you also add the Fedora updates from PR #1220.

reidpr:
With the merge of PR #1220, this should be moot.

PR #1242, opened 2021-11-23
move storage directory to new path

reidpr:
This PR updates the default `ch-image` storage directory to `/var/tmp/$USER.ch` (e.g., `/var/tmp/reidpr.ch`). If the storage directory path in use is the default, and a valid storage directory exists at the old default path, move it to the new default path.

### DIFF
--- a/doc/ch-image_desc.rst
+++ b/doc/ch-image_desc.rst
@@ -139,8 +139,10 @@ In descending order of priority, this directory is located at:
   :code:`$CH_IMAGE_STORAGE`
     Environment variable.
 
-  :code:`/var/tmp/$USER/ch-image`
-    Default.
+  :code:`/var/tmp/$USER.ch`
+    Default. (Previously, the default was :code:`/var/tmp/$USER/ch-image`. If
+    a valid storage directory is found at the old default path,
+    :code:`ch-image` tries to move it to the new default path.)
 
 Unlike many container implementations, there is no notion of storage drivers,
 graph drivers, etc., to select and/or configure.
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -1467,6 +1467,13 @@ class Storage:
       return self.root // "ulcache"
 
    @property
+   def valid_p(self):
+      # FIXME: require version file too when appropriate (#1147)
+      """Return True if storage present and seems valid, False otherwise."""
+      return (os.path.isdir(self.unpack_base) and
+              os.path.isdir(self.download_cache))
+
+   @property
    def version_file(self):
       return self.root // "version"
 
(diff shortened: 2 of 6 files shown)


doc/ch-image_desc.rst
```diff
@@ -139,8 +139,10 @@ In descending order of priority, this directory is located at:
   :code:`$CH_IMAGE_STORAGE`
     Environment variable.
 
-  :code:`/var/tmp/$USER/ch-image`
-    Default.
+  :code:`/var/tmp/$USER.ch`
+    Default. (Previously, the default was :code:`/var/tmp/$USER/ch-image`. If
+    a valid storage directory is found at the old default path,
+    :code:`ch-image` tries to move it to the new default path.)
```
heasterday, line 145: Do you anticipate Charliecloud using `/var/tmp/$USER.ch` for operations besides those driven by `ch-image`? If so, any thoughts on having `/var/tmp/$USER.ch/ch-image` be the default?

lib/charliecloud.py
```diff
@@ -1519,6 +1523,50 @@ def init(self):
...
+            try:
+               shutil.move(src, dst)
+            except OSError as x:
+               FATAL("can't move: %s -> %s: %s"
+                     % (x.filename, x.filename2, x.strerror))
+      ossafe(os.rmdir, "can't rmdir: %s" % old.root, old.root)
+      if (len(ossafe(os.listdir, "can't list: %s" % old.root.parent,
+                     old.root.parent)) == 0):
+         WARNING("storage dir: parent of old now empty: %s" % old.root.parent,
+                 hint="consider deleting it")
```
heasterday, line 1568: Minor note, I found this phrasing to be confusing when I first saw it. Maybe something like "previous storage dir now empty" or "migration complete".

lib/charliecloud.py
```diff
@@ -2032,3 +2074,10 @@ def tree_terminals_cat(tree, tname):
 def unlink(path, *args, **kwargs):
    "Error-checking wrapper for os.unlink()."
    ossafe(os.unlink, "can't unlink: %s" % path, path)
+
+def user():
+   "Return the current username; exit with error if it can't be obtained."
+   try:
+      return os.environ["USER"]
+   except KeyError:
+      FATAL("can't get username: $USER not set")
```
heasterday, line 2083: Are there any scenarios where USER might not be set and we would want to use something like `whoami` as a fallback?

heasterday (approved):
Looks good to me. Added some comments but they are not blockers to merging.

doc/ch-image_desc.rst
reidpr: Yes. In fact we already have `/var/tmp/$USER.ch/mnt` for the default SquashFS. The thinking is that adding another directory level makes maintenance more complicated because `mkdir(2)`/`rmdir(2)` aren't recursive, and the recursive options (analogous to `mkdir -p` or `rm -Rf`) can be more high powered than needed, and this complexity isn't outweighed by clear benefits.

lib/charliecloud.py
reidpr: Yeah it really is the parent of the storage dir though, e.g. `/var/tmp/reidpr`, which I didn't feel we owned so I couldn't just delete it. I'll try to come up with better phrasing.

lib/charliecloud.py
reidpr: I've wondered about this. I think all of Charliecloud is trash if `$USER` isn't set though, so it's a bigger problem than right here.

shanegoff (approved):
This looks good. Tested quickly and it did convert my old storage space to the new one.

PR #1258, opened 2021-12-19
add musl compatibility

olifre:
On musl systems, argp functionality is not provided by libc.
If that is detected, try to add `-largp` which allows to use
`libargp` or `argp_standalone` to provide this functionality.

Fixes build on Alpine Linux and Gentoo with musl:
https://bugs.gentoo.org/829607


reidpr:
Good catch @olifre, much appreciated!

Do you know a way to add musl build to our GitHub Actions CI? That runs on Ubuntu virtual machines, version 20.04 IIRC.

olifre:
> Do you know a way to add musl build to our GitHub Actions CI? That runs on Ubuntu virtual machines, version 20.04 IIRC.

I was also wondering about this. I don't know of a straightforward way — the way I tested it was to pull an alpine container (which uses musl), run it with `ch-run`, and then compile charliecloud within. So in the end, I executed:
```
  dockerhub2oci --repo library --image alpine --tag latest 
  ch-run -w library/alpine/latest -- bin/sh
  git clone https://github.com/hpc/charliecloud.git
  apk add alpine-sdk
  apk add automake
  apk add autoconf
  apk add python3
  apk add py3-pip
  apk add py3-wheel
  apk add argp-standalone
  sh ./autogen.sh && ./configure
  make 
```
[DockerHub2OCI](https://github.com/olifre/dockerhub2oci) is a basic unprivileged pull tool I wrote some time ago (with many limitations), but of course `ch-pull2dir` should work, too.

reidpr:
Thanks. Looks like Ubuntu does provide a `musl-gcc` wrapper. Do you mind if I mess around on this branch to get a test added to it?

olifre:
> Looks like Ubuntu does provide a `musl-gcc` wrapper also.

That sounds even better, I did not know about this :-). 



> Do you mind if I mess around on this branch to get a test added to it?

I don't mind, of course I welcome having a test — please feel free to go ahead :+1: .

olifre:
This is great, and easier than expected :+1: . Getting a `musl` toolchain setup on Gentoo is significantly harder, so it's great Debian (and thus, Ubuntu) ship it out of the box in functional state and overriding `CC` is sufficient to use it. 
Thanks again!

reidpr:
I got it to build, but not link with the system `libfuse3` and the `libsquashfuse` we install, both of which were built against glibc — that seems to me out of scope for this PR or  just not a supported configuration. I think tomorrow I will move the `libsquashfuse` version back to glibc and see if the tests pass. This process shook out a couple other bugs so again I appreciate the report/PR!

olifre:
> I got it to build, but not link with the system `libfuse3` and the `libsquashfuse` we install, both of which were built against glibc — that seems to me out of scope for this PR or just not a supported configuration.

Indeed, that is quite likely to be an unsupported configuration. Probably, building against a full `musl`-based stack is only reasonably possible by using a `musl`-based distribution such as Alpine (otherwise, all dependencies would have to be rebuilt). I did not think of this beforehand, but this seems very reasonable. 


> This process shook out a couple other bugs so again I appreciate the report/PR!

Sure, you are welcome :smile: .

reidpr:
OK, I think this is ready for review. Below is a summary of what's changed. @olifre, can you test it in your environment?

**Build**

1. musl does not provide `argp_parse(3)` and friends, which are a GNU extension. However an external `libargp` can be used. `configure` now detects this and adds appropriate linking flags. (@olifre, I used your logic 100% but changed the style to be more idiomatic for the project.)

**Test**

1. GitHub Actions now builds against musl for some of the tests, using the `musl-gcc` wrapper included with Ubuntu. This build won't link with any of the glibc stuff already on the system, which is the main reason why it's only some of the tests.

**Bugs fixed**

The test suite identified several bugs that caused `ch-run` to fail at runtime when built against musl. I'll list them comprehensively for reference. Under musl:

1. `limits.h` must be included for `INT_MIN` and `INT_MAX` (it must have been incidentally included with glibc).
2. `strtol(3)` sets `errno` to `EINVAL` if the input contains no digits (glibc does not set `errno` in this case).
3. `strtol(3)` fails with `ERANGE` in more cases than glibc's.
4. `sem_open(3)` and `shm_open(3)` share a namespace (filenames in `/dev/shm`).
5. `getline(3)` segfaults if provided an uninitialized `char *`.
6. `statvfs(2)` passes through flag bit 0x20, which is `ST_VALID` in the kernel, undefined in `sys/statvfs.h` for both glibc and musl, and inappropriate for `mount(2)`. glibc unsets it from the flags. (This one smells like a bug in musl.)

olifre:
@reidpr This looks great, I can confirm it runs fine in an Alpine musl environment :+1: . I also prefer the code style in `configure.ac` over my own :smile:. 
I still do not have a Gentoo musl environment to test (this seems to require rebuilding most parts of the system), but I guess this will not give us another data point anyways ;-).

j-ogas (approved):

Issue #1262, opened 2022-01-04
`ch-image`: crashes during parse in C locale on non-ASCII characters under Python 3.6


reidpr:
For example:

```
$ env | egrep '^LC_'
LC_TERMINAL=iTerm2
LC_TERMINAL_VERSION=3.4.15
LC_ALL=C
$ python3 --version
Python 3.6.12
$ ch-image build --parse-only -t spack -f examples/spack/Dockerfile examples/spack
Traceback (most recent call last):
  File "bin/ch-image", line 240, in <module>
    main()
  File "bin/ch-image", line 234, in main
    cli.func(cli)
  File "/host/p/Charliecloud/charliecloud/bin/../lib/charliecloud/build.py", line 139, in main
    text = ch.ossafe(fp.read, "can't read: %s" % cli.file)
  File "/host/p/Charliecloud/charliecloud/bin/../lib/charliecloud/charliecloud.py", line 1984, in ossafe
    return f(*args, **kwargs)
  File "/usr/local/stow/python-3.6.12/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1896: ordinal not in range(128)
```

The offending character is the curly quote on line 43 of the Dockerfile:

```
# Depending on what's commented below, we get either Spack’s “develop” branch
```

However, with Python 3.7 it works:

```
$ python3 --version
Python 3.7.3
$ ch-image build --parse-only -t spack -f examples/spack/Dockerfile examples/spack
$
```

And with a different locale it works:

```
$ export LC_ALL=en_US.utf8
$ python3 --version
Python 3.6.12
$ ch-image build --parse-only -t spack -f examples/spack/Dockerfile examples/spack
$
```

I have not tried locales other than `C`, `en_US.utf8`, and whatever is the default on my system, i.e. `LC_ALL` unset.

I suspect the cause is the same as #970.

Given that Python 3.6 is rather old and a straightforward workaround is available, I want to record the problem but I'm closing this immediately as WONTFIX.

Issue #970, opened 2021-01-25
ch-image: cannot pull some images with LC_ALL=C on Python 3.6


reidpr:
`ch-image` crashes while pulling (a) certain images under (b) Python 3.6 with:

```
UnicodeEncodeError: 'ascii' codec can't encode character '\u0151' in position 28: ordinal not in range(128)
```

One such image is `nvcr.io/hpc/foldingathome/fah-gpu:7.6.21`.

A workaround is to set the locale to something UTF-8, e.g. `LC_ALL=en_US.utf-8`.

The cause seems to be some pretty hairy locale/encoding stuff in Python that seems to have been fixed in 3.7.

reidpr:
3.6 is getting old and I don’t know to fix this. Closing.

PR #1263, opened 2022-01-04
fix tests

### LINKED ISSUE #1172: examples: investigate "fuzzy" image comparisons
For our Paraview and obspy tests we generate images and compare them against a known good image (I believe the term is a gold file) using `cmp`. This appears to be brittle when generating images across machines. We should investigate a fuzzy matching solution so we tolerate a small amount of difference.

### LINKED ISSUE #1266: `ch-test`: use locale `en_US.utf8` if available
Locale `C` causes problems on Python 3.6; see issues #970 and #1262, which in turn causes the test suite to fail.

### LINKED ISSUE #1288: examples: images using centos:8 are failing
Since centos 8 is EOL we need to migrate this images.

related #1279

### LINKED ISSUE #1289: examples: tests that revert images after "cray-ifying" them still use `ch-tar2dir`
`ch-tar2dir` is deprecated and should be replaced by `ch-convert`. This leads to failures in the tests because the initial unpack is done with `ch-convert` which doesn't place the sentinel file that `ch-tar2dir` expects.

### LINKED ISSUE #1290: examples: serial MPI tests fail on Cray xc40
After some recent changes serial MPI tests fail on our Cray xc40. It sounds like this is the way that things will be moving forward, we should skip these tests.

reidpr:
This works around locale issues for the test suite (#1262 and possibly #970, closes #1266) and adds fuzzy image compare (closes #1172).
Additionally, this PR closes #1290, closes #1289, and closes #1288.

### DIFF
--- a/test/common.bash
+++ b/test/common.bash
@@ -171,9 +171,6 @@ unpack_img_all_nodes () {
     fi
 }
 
-# Predictable sorting and collation
-export LC_ALL=C
-
 # Do we have what we need?
 env_require CH_TEST_TARDIR
 env_require CH_TEST_IMGDIR
--- a/test/common.bash
+++ b/test/common.bash
@@ -116,6 +116,34 @@ pedantic_fail () {
     fi
 }
 
+# If the two images (graphics, not container) are not "almost equal", fail.
+# The first argument is the reference image; the second is the test image. The
+# third argument, if given, is the maximum number of differing pixels (default
+# zero). Also produce a diff image, which highlights the differing pixels in
+# red, based on the sample, e.g. foo.png -> foo.diff.png.
+pict_assert_equal () {
+    ref=$1
+    sample=$2
+    pixel_max_ct=${3:-0}
+    sample_base=${sample%.*}
+    sample_ext=${sample##*.}
+    diff_=${sample_base}.diff.${sample_ext}
+    echo "reference:   ${ref}"
+    echo "sample:      ${sample}"
+    echo "diff image:  ${diff_}"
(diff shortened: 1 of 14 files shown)


test/common.bash
```diff
@@ -116,6 +116,34 @@ pedantic_fail () {
...
+    sample=$2
+    pixel_max_ct=${3:-0}
+    sample_base=${sample%.*}
+    sample_ext=${sample##*.}
+    diff_=${sample_base}.diff.${sample_ext}
+    echo "reference:   ${ref}"
+    echo "sample:      ${sample}"
+    echo "diff image:  ${diff_}"
+    # See: https://imagemagick.org/script/command-line-options.php#metric
+    pixel_ct=$(compare -metric AE "$ref" "$sample" "$diff_" 2>&1 || true)
```
heasterday, line 135: Should ImageMagick be add as a dependency for the "complete test suite" in `configure.ac`?

heasterday (commented):
I would like to test this on my dev machine but this looks good. One minor comment.

test/common.bash
reidpr: Yeah probably. There's a lot of that stuff. Why don't we defer it to issue #1267.

heasterday:
@reidpr It's not allowing me to request your review, can you take a look at this PR?

heasterday:
@reidpr Updated issues that this resolves. @j-ogas I would appreciate you taking a look at this one.

j-ogas (approved):

Issue #1158, opened 2021-08-24
tell compiler `FATAL` can't return


reidpr:
`FATAL()` and possibly a few other functions/macros can't return. If we tell the compiler this, e.g. with GCC's `__attribute__ ((noreturn))`, then it will avoid some warnings we currently work around awkwardly.

reidpr:
Closed in PR #1264 but typo’ed the link.

PR #1264, opened 2022-01-05
harmonize C and Python logging

### LINKED ISSUE #1019: harmonize verbose levels between Python and C
In Python the default is `verbose == 0` and the value is the number of `-v` given. In C the default is `verbose == 1` and the value is the number if `-v` given plus 1.

Also, in Python `INFO` is printed by default while in C it is not. Harmonize the levels; I think probably all `INFO` in C could be changes to `VERBOSE`. The difference here is that the C code is pretty quiet by default while the Python code is chatty, but the calls should mean the same thing.

reidpr:
Closes #1019 and #1158.

### DIFF
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -180,14 +180,14 @@ int main(int argc, char *argv[])
    for (int i = 0; i < argc - arg_next; i++)
       c_argv[i] = argv[i + arg_next];
 
-   INFO("verbosity: %d", verbose);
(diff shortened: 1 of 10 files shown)


reidpr:
One thing I'm currently struggling with is how to articulate the distinction between VERBOSE and DEBUG.

j-ogas (approved):
LGTM.

PR #1273, opened 2022-01-24
packaging: update spec files for 0.26 release

j-ogas:
This PR updates the `upstream.spec` with our fedora build targeting 0.26. Note the `printf.patch` was required to address warnings treated as errors (#1274). When the issue is fixed we will want to remove the patch for the 0.27 update, thus I don't update the `charliecloud.spec` for CI testing because: 1) the change is trivial, and 2) it will just be reverted anyway. Instead I've included the patch in the packaging directory as a record/reference.

### DIFF
--- a/packaging/fedora/upstream.spec
+++ b/packaging/fedora/upstream.spec
@@ -19,6 +19,7 @@ Source0:       https://github.com/hpc/%{name}/releases/downloads/v%{version}/%{n
 BuildRequires: gcc rsync bash
 Requires:      squashfuse squashfs-tools
 Patch1:        el7-pkgdir.patch
+Patch2:        printf.patch
 
 %description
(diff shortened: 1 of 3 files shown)


packaging/fedora/upstream.spec
```diff
@@ -190,6 +188,9 @@ ln -s "${sphinxdir}/js"    %{buildroot}%{_pkgdocdir}/html/_static/js
 
 
 %changelog
+* Mon Jan 24 2021 Jordan Ogas <jogas@lanl.gov 0.26-1
```
j-ogas, line 191:
```suggestion
* Mon Jan 24 2022 Jordan Ogas <jogas@lanl.gov 0.26-1
```

reidpr (approved):

PR #1276, opened 2022-01-25
Add more fakeroot support

loveshack:
Covers Debian derivatives, (Open)SUSE, Alpine, and Arch (older
releases, or on a host which supports the syscalls it uses).

Note that os-release is a Freedesktop standard.

### DIFF
--- a/lib/fakeroot.py
+++ b/lib/fakeroot.py
@@ -139,6 +139,9 @@ DEFAULT_CONFIGS = {
    #      seems a maintenance headache, :latest changes.
    #
    #    * grep the same file for each distro: No standardized file for this.
+   #      [Fixme: /etc/os-release is sufficiently standardized and
+   #       covers derivatives (but see Arch below)?
+   #       https://www.freedesktop.org/software/systemd/man/os-release.html]
    #
    #    * Ask lsb_release(1): Not always installed, requires executing ch-run.
 
@@ -180,6 +183,7 @@ DEFAULT_CONFIGS = {
      "init": [ ("command -v fakeroot > /dev/null",
                 "set -ex; "
                 "if ! grep -Eq '\[epel\]' /etc/yum.conf /etc/yum.repos.d/*; then "
+                # Fixme: Wrong for EPEL9
                 "dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm; "
                 "dnf install -y fakeroot; "
                 "dnf remove -y epel-release; "
@@ -212,8 +216,8 @@ DEFAULT_CONFIGS = {
    #      | egrep '^(fakeroot|fakeroot-ng|pseudo)$'
 
    "debderiv":
-   { "name": "Debian 9+ or Ubuntu 14+",
-     "match": ("/etc/os-release", r"Debian GNU/Linux (?![0-8] )|Ubuntu (?![0-9]\.|1[0-3]\.)"),
+   { "name": "Debian 9+, Ubuntu 14+, or other recent Debian derivative",
+     "match": ("/etc/os-release", r"Debian GNU/Linux (?![0-8] )|Ubuntu (?![0-9]\.|1[0-3]\.)|ID_LIKE=debian"),
      "init": [ ("apt-config dump | fgrep -q 'APT::Sandbox::User \"root\"'"
                 " || ! fgrep -q _apt /etc/passwd",
                 "echo 'APT::Sandbox::User \"root\";'"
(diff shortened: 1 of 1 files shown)


lib/fakeroot.py
```diff
@@ -233,13 +237,49 @@
...
+                # failing with missing getopt in the fakeroot script.
+                "zypper refresh; zypper install -y fakeroot /usr/bin/getopt") ],
+     "cmds": ["zypper", "rpm"],
+     "each": ["fakeroot"] },
+
+   # If I use a recent image, it fails (on Debian 11) to find the
+   # packages mirror, which looks like
+   # https://gitlab.archlinux.org/archlinux/archlinux-docker/-/issues/56
+   # Tested with base-20210307.0.16708 on Debian 11.  Recent Arch might
+   # work under current Fedora.
```
reidpr, line 264: Should this comment refer to some Arch version instead of Debian 11?

reidpr (commented):
Fantastic @loveshack, thanks for the contribution!

I would like to also add tests for the new distros. If you want, you could put them in `test/force-auto.py.in`; it should be relatively straightforward short new classes near the end. We like to test the earliest distro version supported and then "latest" or similar rolling newest version.

Otherwise, we'd be happy to put those in. In that case, what versions of SUSE, Arch, and Alpine should we test?

loveshack:
You wrote: 

> I would like to also add tests for the new distros. If you want, you
> could put them in `test/force-auto.py.in`; it should be relatively
> straightforward short new classes near the end. We like to test the
> earliest distro version supported and then "latest" or similar rolling
> newest version.

I guess I can add tests, but not immediately.  I'm afraid I've no idea
about earliest versions for Alpine and Arch, but fakeroot has been in
SUSE for ages, and zypper hasn't changed, so it presumably could only
break in later versions.

> Otherwise, we'd be happy to put those in. In that case, what versions
> of SUSE, Arch, and Alpine should we test?

I used current opensuse/leap (but see above), current alpine, and arch
as below.

>> +   # If I use a recent image, it fails (on Debian 11) to find the
> +   # packages mirror, which looks like
> +   # https://gitlab.archlinux.org/archlinux/archlinux-docker/-/issues/56
> +   # Tested with base-20210307.0.16708 on Debian 11.  Recent Arch might
> +   # work under current Fedora.
>
> Should this comment refer to some Arch version instead of Debian 11?

It means building that version of Arch on Debian 11, i.e. that arbitrary
arch version is old enough not to require a newer kernel than Debian's
5.10 to support glibc.  But now I realize I can't reproduce the problem
I thought there was with Arch latest, at least natively and pulling from
docker hub; it does work OK with fakeroot.  I'll amend the comment.

loveshack:
I think this should have high priority rather than medium to allow use of Alma, at least. (If there's anything wrong with it, let me know.)
EL8 examples now normally fail.  CentOS Stream 8 should work, but the only images I know for it are the linuxcontainers ones.

loveshack:
> However:
>
> 4. Arch does not work; even with `--force` the problem package
> (`dbus-1`) fails with `fchownat()` errors. However, it's possible
> `--force` is not needed for Arch, because `pacman` printed errors but
> did not actually fail. I had to go out of my way to grep the logs to
> detect the problem.

Sorry, I can't now remember what I saw, but it did appear to work.  I
can have another look, but I'm not an Arch'er.  I've an idea a colleague
is, though -- I'll ask and maybe have another look.  I guess the test I
added isn't good enough.

> 5. Alpine also does not work; it still gets "failed to set ownership".
>
> If you have good ideas on how to fix Arch and Alpine, I'm all ears. If
> not, I suggest we move those into a future PR.

Definitely not Alpine, and I suspect Arch is more likely to be used.  A
"help wanted" issue?  I'll let you know if I get anywhere with Arch
soon.

By the way, I can't remember if I said, but I looked through the list of
linuxcontainers images for any that I thought might be relevant, and the
only other one I couldn't work out was gentoo (which is probably at most
marginal), but I didn't know what some of them were!

reidpr:
Great, thanks. I put Arch in issue #1295 and Alpine in issue #1296.

reidpr:
OK I think we have a merge candidate now. @loveshack does this work for you?

lib/fakeroot.py
```diff
@@ -224,22 +245,46 @@
...
-                "dnf install -y fakeroot") ],
-     "cmds": ["dnf", "rpm", "yum"],
+                # fakeroot seems to have a missing dependency, otherwise
+                # failing with missing getopt in the fakeroot script.
+                "zypper refresh; zypper install -y fakeroot /usr/bin/getopt") ],
+     "cmds": ["zypper", "rpm"],
      "each": ["fakeroot"] },
 
+   # no worky; see #1295
+   # [It seems to me (Dave Love) that it works sufficiently well, just
```
j-ogas, line 260:
```suggestion
   # It seems to me (Dave Love) that it works sufficiently well, just
```
Typo?

lib/fakeroot.py
```diff
@@ -224,22 +245,46 @@
...
+   # { "name": "Arch Linux",
+   #  "match": ("/etc/os-release", r"ID=arch"),  # /etc/arch-release empty
+   #  "init": [ ("command -v fakeroot > /dev/null",
+   #             "pacman -Syq --noconfirm glibc util-linux fakeroot") ],
+   #  "cmds": ["pacman"],
+   #  "each": ["fakeroot"] },
+
+   # no worky; see #1296
+   #"alpine":
+   #{ "name": "Alpine, any version",
```
j-ogas, line 282:
```suggestion
   # "alpine":
   # { "name": "Alpine, any version",
```

Be consistent with previous comment spacing.

j-ogas (changes requested):
See in-line comments. Aside from that LGTM.

lib/fakeroot.py
reidpr: I actually prefer the no-space version for commenting out code, because it provides a visual break from possibly preceding text comments.

lib/fakeroot.py
reidpr: There's a closing bracket too so I think this was intended.

j-ogas (approved):
LGTM.

PR #1281, opened 2022-02-15
handle absolute root directory and hard link targets

adrecord:


reidpr:
@adrecord Sorry about that — sounds like PR #1276 fixes your issue? I didn't mean to step on your toes; the solution for your issue just kind of emerged from fixing the other thing.

Issue #1268, opened 2022-01-14
document foreign architecture images


loveshack:
After responding to an assertion that you can't build container images for ppc64le on amd64, I looked for something about it in the docs.  I can't see anything about building images for foreign architectures, and --arch might mislead people to think it will just work generally.  Anyway, it does basically just work after installing (on Debian) the qemu-user-static package to support execution in the container.  (The -static version is necessary.)  Unfortunately that isn't in current RHEL/EPEL, but is in Fedora; I haven't checked elsewhere.

Related to that, I wonder if it's worth mentioning images.linuxcontainers.org as a source of much rootfs for when you can't pull something off docker hub (like --arch ppc64le CentOS in my case).

reidpr:
This sounds like a good FAQ, yes. Would you be interested in submitting a draft of your procedure as a PR? We don't know enough about this area to do so based on the notes above.

loveshack:
I'll try to concoct something when I have a chance.

I've since noticed a nuisance that if I use --arch to pull, say, a
debian:11 ppc64le image on x86_64 and then try to use `FROM debian:11`
for a native build, it picks up the foreign one.  I wonder if there's a
workaround for that.

The something else occurred to me in the spirit of unprivileged
operation, which we like:  You can't do a foreign build that isn't a
straight pull and unpack without an obviously privileged installation of
the kernel binfmt hooks to make the virtualization work transparently.
However, I think you could make dockerfiles work for foreign images
anyway if there was a hook for RUN (possibly elsewhere?) to allow the
command to be prefixed by the appropriate qemu-...-static command, like
PRoot's -q:

```
      -q command, --qemu=command
              Execute guest programs through QEMU as specified by command.

              Each time a guest program is going to be executed, PRoot inserts
              the QEMU user-mode command in  front  of  the  initial  request.
              That way, guest programs actually run on a virtual guest CPU em‐
              ulated by QEMU user-mode.
```

You can install the qemu-user-static binaries in $HOME somewhere for use
with a hook like that, if you've acquired them by unpacking a system
package or whatever, and you should then be all set without a
cooperative system manager.

Does that sound plausible?  I haven't looked at doing it.

reidpr:
> I'll try to concoct something when I have a chance.

Fantastic, thank you.

[...]

PR #1291, opened 2022-02-24
FAQ: add foreign-arch image building and linuxcontainers

loveshack:
Here's a go at addressing #1268, hoping it's useful.  (I wrote something more extensive for local consumption, but that's rather messed up by using examples with centos8 and commands which will stop working in 0.27.)

By the way, the doc copyright footer should presumably have 2021 changed to 2022 and probably say something like "Triad Security and contributors" to cover contributions,

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -953,62 +953,72 @@ One fix is to configure your :code:`.bashrc` or equivalent to:
      }
 
 
-.. _faq_foreign-images:
-
 How can I build images for a foreign architecture?
 --------------------------------------------------
 
 QEMU
 ~~~~
 
-Suppose you want to build Charliecloud containers on a system which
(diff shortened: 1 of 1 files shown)


reidpr:
@loveshack, this is fantastic. I did a (very light) copyedit and added an example. If it looks good to you, I'll merge.

heasterday (approved):

loveshack:
Sorry I didn't get round to looking at the amended version in time.  I
might wonder if Alpine is a good example in view of the missing fakeroot
support, but I guess it's small, and as you see fit.

[I found that showing simple Charliecloud solutions to things that
apparently couldn't be done -- originally building ppc64 images on the
system without privilege, and then on x86 -- just makes people look for
a way to do the same thing for Singularity.  PR...]

Issue #1305, opened 2022-03-16
ch-image pull erroneously using cached data


adrecord:
`ch-image pull` seems overlay aggressive in use of its download cache.

Below is a script that helps show the problem.  It uses docker to build a simple image that prints out a string, pushes the image to a container registry, then pulls and runs the image using both docker and charliecloud.  Note that you'll need to update `EXTERNAL_IMAGE` to point to an external registry tag where the container image can be pushed.  This script takes a single argument of the string to have the containerized script print.  To demonstrate the problem, run this script twice, passing in a different string to echo each time.  You'll notice the second time ch-run runs the original script, as `ch-image pull` has incorrectly used files from it's dlcache instead of pulling the new image.

```bash
#!/bin/bash

set -o xtrace

EXTERNAL_IMAGE=user/test:latest
MY_STRING=${1}

cat << EOF | docker build -t ${EXTERNAL_IMAGE} -f - . 
FROM alpine:latest
RUN printf '#!/bin/sh\necho "${MY_STRING}"\n' > /test.sh && chmod 755 /test.sh
EOF

docker push ${EXTERNAL_IMAGE}
docker pull ${EXTERNAL_IMAGE}
docker run --rm -it ${EXTERNAL_IMAGE} /test.sh
rm -rf ./test
ch-image pull ${EXTERNAL_IMAGE} ./test
ch-run --unset-env=* --set-env ./test -- /test.sh
```

I believe `ch-image pull` should never use the cache.

reidpr:
I think PR #1305 might address this.

I believe the right thing to do by default isn't re-download everything, but rather re-download just the manifest, and only if that indicates changes, re-download the config and layers.

j-ogas:
> I believe the right thing to do by default isn't re-download everything, but rather re-download just the manifest, and only if that indicates changes, re-download the config and layers.

If the manifest is automatically downloaded and checked for a hash difference, do we need `--no-cache` here? It seems to me that we can rely on the manifest to determine whether or not to use the dlcache files without user interaction. Thoughts?

reidpr:
At first glance that seems reasonable for normal operation, though for debugging I'd want to retain `--no-cache`.

adrecord:
> I believe the right thing to do by default isn't re-download everything, but rather re-download just the manifest, and only if that indicates changes, re-download the config and layers.

I agree, that makes more sense.

PR #1306, opened 2022-03-16
do not use cache on ch-image pull

adrecord:
PR for #1305


j-ogas:
Closing in favor of #1309.

PR #1309, opened 2022-03-16
ch-image: update pull behavior

j-ogas:
Addresses #1305.

This enables `ch-image` to catch and pull changes from images in external repositories.


lib/pull.py
```diff
@@ -139,9 +141,9 @@ def error_decode(self, data):
       return (code, msg)
 
    def fatman_load(self):
-      """Load the fat manifest JSON file, downloading it first if needed. If
-         the image has a fat manifest, populate self.architectures; this may
-         be an empty dictionary if no valid architectures were found.
+      """Download the fat manifest and load it. If the image has a fat manifest
+         populate self.architectures; this may be an empty dictionary if no
+         valid architectures were found.
```
reidpr, line 146: These two comments should explain why we always download it.

lib/pull.py
```diff
@@ -111,7 +111,9 @@ def download(self):
       # config
       ch.VERBOSE("config path: %s" % self.config_path)
       if (self.config_path is not None):
-         if (os.path.exists(self.config_path) and self.use_cache):
+         if (    os.path.exists(self.config_path)
+             and (ch.file_hash(self.config_path) == self.config_hash)
```
reidpr, line 115: This should use the hash in the filename rather than re-computing it. In fact, does this even need to be changed? By the time we call it, won't the config path be up to date with the new hash from the manifest?

reidpr (changes requested):
Couple suggestions in-line.

Is there a way to add a test that's not too onerous? Perhaps in `build/50_localregistry.bats`?

adrecord:
Any updates on this one?  This causes real problems when the pulled image can't be trusted as being correct.  I currently recommend people always use `--no-cache` to be sure.

reidpr:
Yes, I want to wait until the build cache PR lands because that will upend that part of the code. But this is one of the first things to merge after that. (Thanks for waiting on my reply; I've been on vacation.)

j-ogas:
OK, this is now behaving as intended. `ch-image pull` now always retrieves the image manifests (fat and standard) and will download any config and layers referenced in the manifest should they not already exist in storage.

I do wonder now if this behavior makes `--dlcache` obsolete and thus unnecessary. It was previously used to manually force an update by re-downloading everything. Now images in storage will always be up-to-date. Is there any value in forcing a re-download of everything? FWIW, removing `--dlcache` does make things much simpler, e.g., we could simply replace `--bucache` with `--cache`?

doc/ch-image_desc.rst
```diff
@@ -79,23 +79,21 @@ Common options placed before the sub-command:
...
 
   :code:`--no-cache`
-    Shorthand for :code:`--bucache=disabled --dlcache=write-only`. *Note:* If
-    you simply want to re-execute a Dockerfile in its entirety, use
-    :code:`--bucache=rebuild` instead, as that will avoid re-caching the base
-    image specified in :code:`FROM`.
+    Disable build cache (default).
+
+  :code:`--rebuild`
+    For cache misses for all operations except for :code:`FROM` instructions.
```
j-ogas, line 89:
```suggestion
    Do not reuse cached operations; execute all non :code:`FROM` instructions and add them to the cache.  
```

reidpr (approved):
LGTM. Nice work! If tests pass and my tidy looks OK to you, then I'll merge.

test/build/55_cache.bats
```diff
@@ -311,22 +311,156 @@ EOF
...
+    [[ $status -eq 0 ]]
+    diff -u - <(echo "$output" | treeonly) <<'EOF'
+*  (wc) RUN cat /worldchampion
+*  (localhost+5000%champ) PULL localhost:5000/champ
+| *  RUN cat /worldchampion
+| *  PULL localhost:5000/champ
+|/
+*  (HEAD -> root) ROOT
+EOF
 }
```
j-ogas, line 465:
```suggestion
#FIXME: for issue: 1359, add test here where they revert the image in the remote registry to a previous state; our next pull will hit, and so too should any subsequent previously cached instructions based on the FROM SID.```

j-ogas (commented):
A couple minor suggestions. Once resolved I'm happy with merging.

Issue #1318, opened 2022-03-22
ch-image:  Unable to pull from nVidia's NGC registry server


mej:
While attempting to download one of nVidia's "official" HPL-based HPC benchmarking containers, I found that `ch-image` was unable to correctly authenticate with the registry API.  From what I can tell, the NGC registry server is either using an entirely different authentication method than what `ch-image` supports or wanting to utilize a different auth mechanism without indicating that in a way `ch-image` currently recognizes.

The initial attempt:
```bash
> ch-image -vvv pull nvcr.io/nvidia/hpc-benchmarks:20.10-hpl .  
verbose level: 3
host architecture from uname: x86_64
host architecture for registry: amd64
found storage dir v2: /var/tmp/mej.ch
image_ref
  ir_hostport   nvcr.io
  ir_path       nvidia
  ir_name       hpc-benchmarks
  ir_tag        20.10-hpl

pulling image:    nvcr.io/nvidia/hpc-benchmarks:20.10-hpl
requesting arch:  amd64
destination:      .
downloading image: nvcr.io/nvidia/hpc-benchmarks:20.10-hpl
initializing session
auth: no authorization
GET: https://nvcr.io:443/v2/nvidia/hpc-benchmarks/manifests/20.10-hpl
response status: 401
HTTP 401 unauthorized
authorizing
WWW-Authenticate raw: Bearer realm="https://nvcr.io/proxy_auth",scope="repository:nvidia/hpc-benchmarks:pull,push"
WWW-Authenticate parsed: Bearer {'realm': 'https://nvcr.io/proxy_auth', 'scope': 'repository:nvidia/hpc-benchmarks:pull,push'}
authenticating using Bearer
requesting anonymous auth token
GET: https://nvcr.io/proxy_auth
response status: 200
received auth token: eyJraWQiOiJFUkNPOklCWFY6TjY2SDpO
retrying with auth: Bearer eyJraWQiOiJFUkNPOklCWFY6TjY2SDpO
GET: https://nvcr.io:443/v2/nvidia/hpc-benchmarks/manifests/20.10-hpl
response status: 401
b'<html>\r\n<head><title>401 Authorization Required</title></head>\r\n<body bgcolor="white">\r\n<center><h1>401 Authorization Required</h1></center>\r\n<hr><center>nginx/1.14.2</center>\r\n</body>\r\n</html>\r\n'
error: not in registry: nvcr.io:443/nvidia/hpc-benchmarks:20.10-hpl
```

[...]

PR #1324, opened 2022-03-27
retry with authenticated token if anon token is not authorized

adrecord:
PR for #1318

### DIFF
--- a/doc/ch-image_desc.rst
+++ b/doc/ch-image_desc.rst
@@ -78,6 +78,10 @@ Common options placed before the sub-command:
         architecture as a simple string and converts to/from the registry view
         transparently.
 
+  :code:`--auth MODE`
+    Set authentication mode to :code:`MODE`, which is one of **[FIXME]** or
+    :code:`auto` (the default). See section "Authentication" below.
+
   :code:`--no-cache`
     Download everything needed, ignoring the cache.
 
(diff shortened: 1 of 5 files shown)


reidpr:
Thank you Andy! We'll look at this as soon as we can. It's an issue that's bothering a number of people.

Please also consider a separate PR adding yourself to the contributors list in `README.md`. It is alphabetical by last name.

adrecord:
The errors in 40_pull.bats:`pull images that do not exist` are related to this PR, although I'm not sure that the code should do anything different.  What's happening is for the two images that that are invalid images (`doesnotexist:latest`, `charliecloud/doesnotexist:latest`), charliecloud obtains an anonymous token, then tries to pull the manifests with that anon token.  Charliecloud then tries to retrieve the manifest with that anon token and gets a 401, so it proceeds with what this PR introduced... it tries to get an authenticated token to try with.  In this case it prompts for a username/password which isn't going to work in a test script:

```
requesting anonymous auth token
GET: https://auth.docker.io/token
response status: 200
received auth token: Bearer (Anon: True) eyJhbGciOiJSUzI1NiIsInR5cCI6IkpX
retrying with auth: Bearer (Anon: True) eyJhbGciOiJSUzI1NiIsInR5cCI6IkpX
GET: https://registry-1.docker.io:443/v2/library/doesnotexist/manifests/latest
response status: 401
HTTP 401 unauthorized with anonymous token
authenticating using Bearer
forcing retrieval of authenticated token

Username:
```
  
I tried the code without this PR included and it's also receiving the 401 with the anon token, but it determines this to mean the image was not in the registry based on explicit logic documented in [charliecloud.py:fatman_to_file](https://github.com/hpc/charliecloud/blob/8c9accbd576aaeb612ce710c24c7af21bf096270/lib/charliecloud.py#L1357-L1360)

```
requesting anonymous auth token
GET: https://auth.docker.io/token
response status: 200
received auth token: eyJhbGciOiJSUzI1NiIsInR5cCI6IkpX
retrying with auth: Bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpX
GET: https://registry-1.docker.io:443/v2/library/doesnotexist/manifests/latest
response status: 401
b'{"errors":[{"code":"UNAUTHORIZED","message":"authentication required","detail":[{"Type":"repository","Class":"","Name":"library/doesnotexist","Action":"pull"}]}]}\n'
error: not in registry: registry-1.docker.io:443/library/doesnotexist:latest
```

The basic problem is that this 401 when using the anon token doesn't really tell you that the manifest doesn't exist.  It could just mean that you aren't authorized to see it.

Note that the two tests that are requesting a valid image, but for which the requested tag does not exist, do pass, as the manifest is retrieved.  Also, if I set `CH_IMAGE_USERNAME` and `CH_IMAGE_PASSWORD` to my dockerhub credentials, the the tests for the missing images also pass, after receiving the authenticated token via this PR's code.

A decent solution is to change the tests, probably passing in `CH_IMAGE_USERNAME` and `CH_IMAGE_PASSWORD` as some bogus values, then having the test expect 401's in the case of the 2 missing images.  (The test for the missing tags will remain unchanged.)  I'll put something together for this.

adrecord:
I changed the code to accept 401 and 403 from the credentialed token retrieval and have it bubble out as image not existing.  This seems reasonable because to get that far either retrieving the anonymous token failed or using the anonymous token to pull the manifest failed.  At this point, if authenticating fails trying to get an authenticated token the actual problem could still be that the image simply doesn't exist, or it does exist and it's protected.

For reference, here's Docker's behavior, in the case of no `.docker/config.json` file (anon), or there is one with correct credentials.  So it's not really distinguishing between no access versus missing image.

```
andy@host:~/charliecloud$ docker pull charliecloud/manifest:doesnotexist
Error response from daemon: pull access denied for charliecloud/manifest, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

In the case of a `.docker/config.json` with invalid credentials, docker reports this:

```
Error response from daemon: Head "https://registry-1.docker.io/v2/charliecloud/doesnotexist/manifests/latest": unauthorized: incorrect username or password
```

It would be nice to distinguish between these two cases the way docker does, but it's not straight forward because of the dynamic way charliecloud does authentication versus how docker knows in advance if it will be accessing a particular registry with credentials or anonymously (based on the contents of `.docker/config.json').  Docker knows going in if they're trying to access a registry anonymously, and if it's not working can report the 1st error mentioned above.  ch-image pull initially tries anonymously (even if the user provided creds in env vars), then only tries to get creds from user (via vars or prompting) if anon access fails.  If this credentialed access fails it could be that the user didn't even have an account on the registry, but simply wanted to access it anonymously.  We simply don't know, so it makes more sense for us to report the more general "unauthorized OR image not found" instead of the more specific "401 unauthorized".  If charliecloud is changed to preknow if we'll access a registry anonymous or credentialed, we could make the same distinction of errors that docker makes.

This PR is ready for review.

reidpr:
> it's not straight forward because of the dynamic way charliecloud does authentication versus how docker knows in advance if it will be accessing a particular registry with credentials or anonymously

Yeah this is a good point. Another implication is that you can't log into Docker Hub to get your own rate limit. Let me think on that a bit.

reidpr (commented):
I pushed some proposed documentation for a new authentication mode switch `--auth`. This does not match the current behavior of this PR, but I wonder if it would better support situations like not misinterpreting the nVidia NGC not-found responses while also not trying to upgrade to authenticated unnecessarily. (I think the PR as written will also prompt for username/password for public images that do not exist?)

@adrecord, what do you think?

adrecord:
@reidpr Here are my initial thoughts:

> I think the PR as written will also prompt for username/password for public images that do not exist?

This seems like an oxymoron.  If the image doesn't exist, it's not really a public image.  :)  The code can only know if the image is available anonymously *or* with the _specific_ credentials provided.  In staying with the original design of first trying anonymous access, you are right that this PR will request credentials if the image is not anonymously accessible.  I agree that this is an annoyance if the user intended _only_ anonymous access.

Note that even with an authenticated token, the registry returning 401 may still not mean that the image doesn't exist.  It can just mean that _this particular_ authenticated user can't access it.  The doc change in your last commit makes it sound like not being able to access the image with an authenticated token definitively means the image doesn't exist:

> Thus, an second authenticated request is needed, to either access the resource or determine that is really does not exist.

My inclination is to either keep the lazy method as is in this PR as it fixes the user issue in a reasonable way, or switch to *always* being very explicit about whether we'll be accessing the registry with an authenticated account or anonymously.  I think being explicit would simplify things from both coding perspective and usage.  Probably anon access should be the default, and a boolean `--auth` flag added that would cause the code to force authentication, via env vars or prompting.  I recognize that dropping the current `lazy` method would be a breaking change versus previous versions, although lazy mode isn't totally functional as is, as evidenced by this issue, so maybe that's not as big of a deal?

Other considerations that come to mind.. `ch-image build` with multi-stage Dockerfiles can reference images from different registries with diff creds, or anonymously.  I realize that command provides a `--password-many` option but any changes to authentication should consider the impact there.  If we're going to change authentication substantially we should maybe do it in light of needing different auth preferences/creds per registry?  This is something the current registry-agnostic userid/pw env vars don't currently support.  For example, I'm guessing there's currently no good way to do a non-interactive image build that accesses multiple registries with different creds.  If we're gonna rethink auth, it would be nice to do it with use cases like this in mind.

After typing all this, I personally think this PR should probably go as is to address the original issue at hand, and we should discuss broader changes to authenticated vs anon registry access, and multi-registry access separately.

reidpr:
OK, thank you, let me think about it. I moved the doc commit to PR #1329 because regardless, it doesn't seem to fit here.

reidpr:
Closing in favor of PR #1329; however, to be clear, many of the ideas in this PR made their way over there, so I want to emphasize that this contribution is very much appreciated.

PR #1354, opened 2022-05-26
pull: add destination reference

### LINKED ISSUE #1337: `ch-image pull`: allow specifying destination image
Follow up to issue #583 and PR #1255, which removed the second argument of `ch-image pull`. This issue is to restore the second argument, but make it the destination image reference, the same as `ch-image push`.

~~We could also add an argument `-o foo` to output the image to directory `foo` instead of container storage. The non-storage destination conflicted with the build cache somehow; I don't recall why or how bad it was.~~ [Split into issue #1353.]

We will also want to audit existing uses of `ch-image pull` in the test suite, to remove workarounds.

reidpr:
Closes #1337.

### DIFF
--- a/lib/build_cache.py
+++ b/lib/build_cache.py
@@ -230,6 +230,19 @@ class Enabled_Cache:
          ch.FATAL("can't create or delete temporary directory: %s: %s"
                   % (x.filename, x.strerror))
 
+   def branch_nocheckout(self, src_ref, dest):
+      """Create ready branch for Image_Ref src_ref pointing to dest, which can
+         be either an Image_Ref or a Git commit reference (as a string)."""
+      if (isinstance(dest, ch.Image_Ref)):
+         dest = self.branch_name_ready(dest)
+      # Some versions of Git won't let us update a branch that's already
+      # checked out, so detach that worktree if it exists.
+      src_img = ch.Image(src_ref)
+      if (src_img.unpack_exist_p):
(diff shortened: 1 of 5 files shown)


lib/build_cache.py
```diff
@@ -578,7 +593,7 @@ def tree_print(self):
       # See: https://git-scm.com/docs/git-log#_pretty_formats
       if (ch.verbose == 0):
          # ref names, subject (instruction)
-         fmt = "%C(auto)%d %Creset%<|(79,trunc)%s"
+         fmt = "%C(auto)%d %Creset%s"
```
j-ogas, line 596: If limiting the summary output is a testing issue then perhaps `--tree` could interpret multiple verbosity flags.

no verbosity = truncate
-v = full summary, no truncating
-vv = full debug output (what -v does now with `--tree`)

We can then update tests to pass -v with --tree to achieve readable default behavior and a way to test properly.

j-ogas (changes requested):
See multiple verbosity flag suggestion to preserve truncated commit summary and deal will the testing.

Aside from that, and assuming the tests pass, LGTM.

lib/build_cache.py
reidpr: Yeah you convinced me to keep the truncation. I'll fix it.

j-ogas (approved):
LGTM!

j-ogas (approved):

PR #1362, opened 2022-06-02
fakeroot: switch deb/ubuntu from pseudo to fakeroot

heasterday:
Resolves #1348 

This PR switches debian/ubuntu from using the psuedo package to fakeroot to resolve errors seen with Ubuntu 22.04. The PR also sets the env variable `FAKEROOTDONTTRYCHOWN` to `1` to tell fakeroot to not attempt to chown/fchown as this can fail with a `EINVAL` within a user namespace. Credit to Dave Dykstra for pointing us in the right direction on this.

See this PR for further details: https://salsa.debian.org/clint/fakeroot/-/merge_requests/4

PR #1363, opened 2022-06-06
enable Arch and Alpine fakeroot support

### LINKED ISSUE #1295: `ch-image --force`: support Arch

loveshack:
This gets the fakeroot support working using FAKEROOTDONTTRYCHOWN
(which I'll try to get documented) from #1362.  However, running
`ch-test build -b ch-image` fails with
```
...
error: scope '' invalid
found: /home/mdehsdl3/.local/share/doc/charliecloud/examples/serial/obspy/Dockerfile.skipped
found: /home/mdehsdl3/.local/share/doc/charliecloud/examples/chtest/Build

generate build_auto.bats ...
ok

executing build phase tests ...
Error: Duplicate test name(s) in file "/tmp/ch-test.tmp.mdehsdl3/build_auto.bats": test_build_hello test_builder_to_archive_hello
```
I haven't taken time to understand the test setup so I don't have a fix.  I have
tested the change separately, though.

Also there's a configure change to fix the test for FNM_EXTMATCH and
allow the build to work (on Debian 11) with current source.  Otherwise it fails
redefining FNM_EXTMATCH in ch_misc.c.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-20.04
     timeout-minutes: 90
     strategy:
-      fail-fast: true  # if any job fails, cancel the rest immediately
+      fail-fast: false  # Actions seems kind of flaky lately
       matrix:
         builder: [none, docker, ch-image]
         pack_fmt: [squash-mount, tar-unpack, squash-unpack]
--- a/configure.ac
+++ b/configure.ac
@@ -322,7 +322,8 @@ AC_MSG_RESULT($have_userns)
 AC_CHECK_DECL(FNM_EXTMATCH,
               [have_fnm_extmatch=yes],
               [have_fnm_extmatch=no],
-              [[#include <fnmatch.h>]])
+              [[#define _GNU_SOURCE
+#include <fnmatch.h>]])
 
 
 ### SquashFS #################################################################
--- a/configure.ac
+++ b/configure.ac
@@ -323,7 +323,7 @@ AC_CHECK_DECL(FNM_EXTMATCH,
               [have_fnm_extmatch=yes],
               [have_fnm_extmatch=no],
               [[#define _GNU_SOURCE
-#include <fnmatch.h>]])
+                #include <fnmatch.h>]])
 
 
 ### SquashFS #################################################################
--- a/examples/Dockerfile.nvidia
+++ b/examples/Dockerfile.nvidia
@@ -26,8 +26,8 @@ RUN apt-get update \
 WORKDIR /usr/local/src
 RUN wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/cuda-ubuntu2004.pin \
  && mv cuda-ubuntu2004.pin /etc/apt/preferences.d/cuda-repository-pin-600 \
- && wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/7fa2af80.pub \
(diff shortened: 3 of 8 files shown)


heasterday:
Hey Dave thanks for the contribution! For testing this change I would recommend something like the following:  
* To run just the archlinux fakeroot tests: `ch-test -s full -b ch-image --file=../test/build/61_force-auto.bats:archlinux`
* To run just the alpine fakeroot tests: `ch-test -s full -b ch-image --file=../test/build/61_force-auto.bats:alpine`

The issue with Alpine Linux does appear to be fixed but I was still seeing failures with Arch that appear to be resolved by removing `glibc` and `util-linux` from the fakeroot config. Is this consistent with what you are seeing?

reidpr:
@loveshack Do you want to do the testing fixes or should we?

reidpr:
> Also there's a configure change to fix the test for FNM_EXTMATCH and
> allow the build to work (on Debian 11) with current source. Otherwise it fails
> redefining FNM_EXTMATCH in ch_misc.c.

FWIW, it builds fine on my Debian 11.3 amd64 box.

reidpr:
> However, running
> `ch-test build -b ch-image` fails with

I wonder if it's an argument order problem. Does `ch-test -b ch-image build` work?

FWIW, under Python 3.9.2 your line works fine.

loveshack:
You wrote: 

> Hey Dave thanks for the contribution! For testing this change I would
> recommend something like the following:
> * To run just the archlinux fakeroot tests: `ch-test -s full -b
> ch-image --file=../test/build/61_force-auto.bats:archlinux`

All that does is skip >100 cases.

Was I doing something wrong just using the simple invocation?

> * To run just the alpine fakeroot tests: `ch-test -s full -b ch-image
> --file=../test/build/61_force-auto.bats:alpine`
> The issue with Alpine Linux does appear to be fixed but I was still
> seeing failures with Arch that appear to be resolved by removing
> `glibc` and `util-linux` from the fakeroot config. Is this consistent
> with what you are seeing?

Oh, I see, sorry.  However, the installation succeeds anyhow.  There are
other failures reported which might also seem worrying but aren't fatal.

If you remove the installation of glibc, you may end up with the
situation I saw, as in the comment above.  I don't know how
common/problematic that is, but it's quite obscure if you hit it.

Regardless, I think it would be useful for the fakeroot structure to
allow including one or more information messages, perhaps saying
something like "ignore X" or "try Z if Y fails".  Otherwise, perhaps the
expected error/warning messages from the installation could be filtered
out.  It's all a bit of a mess...

loveshack:
> @loveshack Do you want to do the testing fixes or should we?

Sorry, I don't know my way around to do it.

loveshack:
>> Also there's a configure change to fix the test for FNM_EXTMATCH and
>> allow the build to work (on Debian 11) with current source. Otherwise it fails
>> redefining FNM_EXTMATCH in ch_misc.c.
>
> FWIW, it builds fine on my Debian 11.3 amd64 box.

Oh...  I don't understand the difference.  The configure test as written
isn't correct anyway, is it?

reidpr:
OK, I pushed a commit to add the tests.

Hunter's suggestion to remove `glibc` and `util-linux` from Arch makes the tests pass. Based on the comment above, this does seem brittle. I agree it would be good if Charliecloud knew that the errors causing non-zero exit were spurious, but that seems very difficult.

heasterday (approved):

heasterday:
@reidpr This is good to merge. Please note that I updated the nvidia example for a new signing key.

PR #1366, opened 2022-06-10
fix tests on ARM

### LINKED ISSUE #1249: deal with registries that don't use ARM variants
Some registries don't offer variant information with ARM architectures. I've seen it with Google Container Registry and Red Hat. For example:

```
$ ch-image build -f examples/distroless/Dockerfile examples/distroless
inferred image name: distroless
  5 FROM gcr.io/distroless/python3
manifest list: using existing file
error: requested arch unavailable: arm64/v8
hint: available: amd64 arm64
```

In this case, I'm pretty sure `ch-image` should use the `arm64` image.

The fat manifest from GCR; note missing `variant` key:

```
{
  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
  "size": 918,
  "digest": "sha256:066953e1321d8e9acbe94023a14030bc47416a39edd4ab3c11e4b1d4da49089a",
  "platform": {
    "architecture": "arm64",
    "os": "linux"
  }
}
```

Compare to the corresponding excerpt from `centos:7` from Docker Hub (keys re-ordered to match):

```
{
  "mediaType": "application\/vnd.docker.distribution.manifest.v2+json",
  "size": 530
  "digest": "sha256:864a7acea4a5e8fa7a4d83720fbcbadbe38b183f46f3600e04a3f8c1d961ed87",
  "platform": {
    "architecture": "arm64",
    "os": "linux",
    "variant":"v8"
  },

},
```

There are a few issues disabled because of this, commented with the bug number.

### LINKED ISSUE #1364: fusermount3: failed to access mountpoint ...: Permission denied
I am seeing the following intermittently on GitHub Actions in `run/ch-run_join.bats`:

```
# ch-run[111951]: using default mount point: /var/tmp/runner.ch/mnt (ch_fuse.c:121)
# fusermount3: failed to access mountpoint /var/tmp/runner.ch/mnt: Permission denied
# ch-run[111951]: error: unknown FUSE error mounting SquashFS; try ch-run -vv? (ch_fuse.c:235)
```

The `fusermount3` error seems to come from [`fusermount.c`](https://github.com/libfuse/libfuse/blob/bdd2d4110fbc6d2059eb699efad2cca4a7eacccb/util/fusermount.c#L899), though there are four separate places where it's printed and I'm not sure which it is.

Restarting the run seems to fix things. I wonder if there is a race condition?

reidpr:
Closes #1249; closes #1364.

### DIFF
--- a/bin/ch_fuse.c
+++ b/bin/ch_fuse.c
@@ -226,12 +226,29 @@ void sq_mount(const char *img_path, char *mountpt)
    sq.ll = sqfs_ll_open(img_path, 0);
    Te (sq.ll != NULL, "can't open SquashFS: %s; try ch-run -vv?", img_path);
 
-   if (sqfs_ll_mount(sq.chan, sq.mountpt, &mount_args,
-                     &OPS, sizeof(OPS), sq.ll) != SQFS_OK) {
-      // We get back only SQFS_OK or SQFS_ERR, with no further detail. Looking
-      // at the source code [1], the latter says either fuse_session_new() or
-      // fuse_session_mount() failed, but we can't tell which.
-      // [1]: https://github.com/vasi/squashfuse/blob/74f4fe8/ll.c#L399
-      FATAL("unknown FUSE error mounting SquashFS; try ch-run -vv?");
-   }
+   // sqfs_ll_mount() is squirrely for a couple reasons:
+   //
+   //   1. Error reporting. We get back only SQFS_OK or SQFS_ERR, with no
+   //      further detail. Looking at the source code [1], the latter says
+   //      either fuse_session_new() or fuse_session_mount() failed, but we
+   //      can't tell which, or get any further information about what went
+   //      wrong. Hopefully fusermount3 also printed an error message.
+   //
+   //   2. Race condition. We have been seeing intermittent errors in the test
+   //      suite about permission denied accessing the mount point (issue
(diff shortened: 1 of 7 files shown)


bin/ch_fuse.c
```diff
@@ -226,12 +226,29 @@ void sq_mount(const char *img_path, char *mountpt)
...
+   //      wrong. Hopefully fusermount3 also printed an error message.
+   //
+   //   2. Race condition. We have been seeing intermittent errors in the test
+   //      suite about permission denied accessing the mount point (issue
+   //      #1364). I *think* this is because a previous mount on the same
+   //      location is not yet cleaned up. For this reason, we have a short
+   //      retry loop.
+   //
+   // [1]: https://github.com/vasi/squashfuse/blob/74f4fe8/ll.c#L399
+   for (int i = 5; true; i--)
```
mphinney1100, line 244: I wonder if this is the right way to go about the permission denied issue. When I had that issue the only thing that worked was using `fusermount -u`.  Waiting between mounting wasn't the solution for me but if this works great.

reidpr: I think that illustrates the difference between the situations: if the other mount is hung, then yes indeed waiting won't help. But if the other mount is in the process of being torn down, then waiting does seem to help.

That said, tests are failing, so I guess I have more work to do.

mphinney1100 (approved):

Issue #871, opened 2020-10-13
rename images in opaque storage


j-ogas:
It would be nice to change the name of images in opaque storage. Things to consider: 1) architecture hierarchy, and 2) changing the name of the corresponding image manifest and fat manifest?

j-ogas:
With the addition of the build cache, #1255, we can revisit this.

#1282 proposed simply renaming the image. This implied, at least to me, that the build cache would simply rename the branch. This would make all images referencing the previous image reference via `FROM` miss. This is still an option.

However the build cache also gives us the ability to implement a `tag`-like feature. The following.
```
$ ch-image list
foo
$ ch-image tag foo bar
[...]
```

The above would keep the worktree, `foo`, and add another worktree, `bar`, that is based on `foo`. This allows future references to foo to still hit.

PR #1375, opened 2022-06-17
ch-image: add tag subcommand.

j-ogas:
~~Addresses #873.~~
~~Possibly addresses #871.~~

### DIFF
--- a/bin/ch-image.py.in
+++ b/bin/ch-image.py.in
@@ -258,6 +258,14 @@ def main():
    sp = ap.add_parser("reset", "delete everything in ch-image builder storage")
    add_opts(sp, misc.reset, deps_check=True, stog_init=False)
 
+   # tag
+   sp = ap.add_parser("tag", "crate new reference from existing image in storage")
+   add_opts(sp, misc.tag, deps_check=True, stog_init=True)
+   sp.add_argument("source_ref", metavar="SOURCE_REF",
+                   help="existing image reference")
+   sp.add_argument("image_ref", metavar="IMAGE_REF",
+                   help="new image reference")
+
    # Parse it up!
    if (len(sys.argv) < 2):
        ap.print_help(file=sys.stderr)
--- a/bin/ch-image.py.in
+++ b/bin/ch-image.py.in
@@ -259,7 +259,7 @@ def main():
    add_opts(sp, misc.reset, deps_check=True, stog_init=False)
 
    # tag
-   sp = ap.add_parser("tag", "crate new reference from existing image in storage")
+   sp = ap.add_parser("tag", "create new reference from existing image in storage")
    add_opts(sp, misc.tag, deps_check=True, stog_init=True)
    sp.add_argument("source_ref", metavar="SOURCE_REF",
                    help="existing image reference")
--- a/doc/ch-image_desc.rst
+++ b/doc/ch-image_desc.rst
@@ -1019,9 +1019,9 @@ Delete all images and cache from ch-image builder storage.
 
    $ ch-image tag SOURCE_REF IMAGE_REF
 
-Add new image :code:`IMAGE_REF` from existing image :code:`SOURCE_REF` to
-storage and cache. If :code:`IMAGE_REF` already exists then it is replaced. This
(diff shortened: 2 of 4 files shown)


reidpr (commented):
I think this PR is on to something, but I'd like to think more carefully about how it should work.

First:

1. I don't care for the verb *tag*, because it's so thoroughly overloaded, and while the greater container ecosystem loves the term, we have avoided it.

2. It should work both with cache enabled and disabled.

I think there are at least three operations going on here, so let's work out which one(s) to support. I might not be getting the vocabulary right either.

1. *Alias*: Create a second name for an existing image. This is cheap both if cache is disabled (symlink in `img/`) and if it's enabled (symlink and also create a new Git branch). It does complicate the delete operation, and anything else which removes or replaces an image, which will need to learn how to deal with symlinks, and shuffle things around if the target is deleted.

2. *Rename*: Change the name of an image. This is again cheap both with cache disabled (move the directory in `img/`) and enabled (also create new Git branch and delete the old).

3. *Copy*: Create a second, independent image that is identical to an existing image. With cache enabled it can be either expensive (new Git branch, new worktree) or lazy-cheap (new Git branch only, new worktree created when needed); operations like *export* that assume a checked-out worktree may need to be updated. With cache disabled, it is expensive (copy a whole directory tree).

Thoughts?

j-ogas:
> I don't care for the verb tag, because it's so thoroughly overloaded, and while the greater container ecosystem loves the term, we have avoided it.

We do prefer _image reference_ when describing image names, i.e., `NAME[:TAG|@DIGEST]`; however, we do explicitly use "tag" when building, e.g., `ch-image build -t [...]`. I think it may be fair to say the word is overloaded but maybe not within the context of this specific command; `docker tag FOO BAR` is fairly intuitive and understandable.

I'm not opposed to calling it `fork`. I think `tag` is fine.

>    It should work both with cache enabled and disabled.

I'd like to move away from life without a cache and this is a step in that direction. FWIW, maintaining two sets of functionality is expensive in and of itself. 

> Alias: Create a second name for an existing image. This is cheap both if cache is disabled (symlink in img/) and if it's enabled (symlink and also create a new Git branch). It does complicate the delete operation, and anything else which removes or replaces an image, which will need to learn how to deal with symlinks, and shuffle things around if the target is deleted.

I, perhaps incorrectly, do not view this behavior as an _alias_. I see it as a _fork_. In the case of a storage without a cache this would be a recursive `copy`. I'd prefer that be its own unique command, e.g., `ch-image copy`, rather than overload this sub-command with fork-like-behavior for the cache and alias/copy behavior for no cache.

> Rename: Change the name of an image. This is again cheap both with cache disabled (move the directory in img/) and enabled (also create new Git branch and delete the old).

Although renaming is implicitly achieved this sub-command with an enabled cache, I think an explicit sub-command, e.g., `ch-image rename`, should take care of renaming for all cache modes.

TLDR: I'd like this sub-command to do one thing: create a "fork" of the cached image in storage with a new name. Other implied behavior this sub-command can achieve, e.g., copy, rename, alias, etc., should be handled explicitly with unique sub-commands for both storage modes.

reidpr:
OK, following up on our discussion, I've noodled on this a bit. I think there are multiple questions in play and we should consider them separately.

### Q1: What operations should we support?

I still see three:

1. Copy
   1. *Lazy copy.* Called "alias" above. Uses symlinks and Git branches to create a cheap copy in either mode, and if something happens so it shouldn't be a copy any more, we adjust things then.
   2. *Eager copy.* Called "copy" above. Makes a full (expensive) copy in `img`, but they are independent immediately and need no further adjustment.
2. *Move*. Called "rename" above. Cheap in either mode because it's just moving things.

### Q2: Should we support both cache modes, or just cache-enabled?

I agree it would be good if cache-enable was the only mode. But, IMO we don't yet have sufficient evidence on performance and the necessary Git is rather new. I'm also concerned about unknown-unknowns. So, I want to have cache-disabled available for a while, and in turn if a feature is reasonable to implement in both modes, we should do it that way.

### Q3: What should it be called?

I'm happy to work on this once we know what to do.

j-ogas:
Stale. May revisit in the future. Closing.

PR #1376, opened 2022-06-17
make MPI base images depend on stock AlmaLinux base

MFTabriz:


reidpr:
Hello @MFTabriz, thanks for the contribution!

`FROM almalinux8` is not actually a typo — we intended to build it on top of our `almalinux8` image which adds some stuff to the `almalinux:8` base.

That said, I am open to the argument that these are more useful examples if they use a stock base image. Do they build with your change? If not, what changes would make them build?

MFTabriz:
Hi @reidpr! Thanks for your quick response and clear explanation. I was not expecting to go through a custom base image to build charlicloud. I have built the docker images with the proposed changes myself but I have not tested the MPI functionality inside the containers yet. On Github CI the failed test is the same as this PR’s base branch (hpc:master).

I believe merging the customization script with charliecloud’s current dockerfiles would be beneficial to the users. If it’s not possible, I’d propose to fetch it through a public registry using its full URL.

I’ll be happy to make the required modifications in this PR as you see fit.

reidpr:
OK, so what I'm hearing is that (1) the two MPI example Dockerfiles build correctly as modified (i.e., base image is upstream `almalinux:8` rather than our `almalinux8`), but (2) whether they work is unknown.

My *guess* is that the build is the hard part, and if it builds it should work. I can kick off a test here that will exercise the MPI.

As for the permission denied CI error, I hope that PR #1366 will fix it.

Next steps I would propose are:

1. Wait for #1366 to be merged (should be soon).
2. Merge master into this branch.
3. Do an internal full-scope test to make sure MPI works.

I think all three of these are our responsibility, so I'd say for you, just wait for the outcome, hopefully within a few days, and we'll go from there.

MFTabriz:
@reidpr I re-based this branch on master.

reidpr:
Looks like `Dockerfile.openmpi` failed to build in our internal testing.

One quirk that may not be obvious is that both of the Dockerfiles modified by this PR are not exercised by the GitHub CI, which runs our test suite in “standard” scope, only in “full” scope. The full scope takes a while to run, and likely isn't necessary here.

Do the two modified Dockerfiles build for you with `ch-image`?

MFTabriz:
@reidpr Yes! I could successfully build them locally (in an unprivileged environment). Here’s my output of `./ch-image build --force` for the modified [Dockerfile.ompi](https://github.com/hpc/charliecloud/files/8986684/ompi.txt) and [Dockerfile.mpich](https://github.com/hpc/charliecloud/files/8986686/mpich.txt)

reidpr:
I think here is the mismatch. In your log (thanks BTW):

```
124 RUN ['/bin/sh', '-c', 'cd openmpi-${MPI_VERSION} && git apply dont-init-ucx-on-intel-cray.patch ...
```

However the Dockerfile at the tip of this branch reads (unless I've missed something):

```
RUN cd openmpi-${MPI_VERSION} \
 && patch -p1 dont-init-ucx-on-intel-cray.patch \
```

i.e., we replaced `git apply` with `patch` in another recent PR. And the failure on internal CI (which I can reproduce on my dev box) is:

```
# 125. RUN.F cd openmpi-${MPI_VERSION}  && patch -p1 dont-init-ucx-on-intel-cray.patch  && CFLAGS=-O3     CXXFLAGS=-O3     ./configure --prefix=/usr/local                 --sysconfdir=/mnt/0                 --with-slurm                 --with-pmi=/usr/local                 --with-pmix                 --with-ucx                 --disable-pty-support                 --enable-mca-no-build=btl-openib,plm-slurm  && make -j$(getconf _NPROCESSORS_ONLN) install  && rm -Rf ../openmpi-${MPI_VERSION}*
# /bin/sh: patch: command not found
# error: build failed: RUN command exited with 127
```

I believe adding `patch` after `librdmacm-devel` in the `dnf install` at the beginning will fix it. I can try this tomorrow.

reidpr:
Adding the `patch` install seems to fix it. Would you like us to make that change? If so, you'll need to not force-push any more, because that will confuse our repos.

MFTabriz:
I added the patch but I think the jobs got cancelled.

reidpr:
OK, I pushed a couple more changes. This branch now passes full-scope CI on x86-64. On ARM the images build correctly, but there was a spurious error before the examples phase, so I don't know the result there, though I suspect it's fine.

I'm becoming a little nervous about scope here. In addition to changing the base image of `Dockerfile.openmpi`, two more Dockerfiles needed fixing, and in ways that duplicate code in `Dockerfile.almalinux8`.

I am wondering if renaming our distro images could also resolve the confusion that led to this PR. E.g., if `Dockerfile.openmpi8` had `FROM almalinux:8-ch` or `FROM almalinux-charlie:8`, would that make it more clear they were chained?

reidpr:
Closing in favor of PR #1423.

Issue #641, opened 2020-02-12
docs: be consistent about long option argument separator


reidpr:
Currently, the man pages are inconsistent about whether to separate arguments from long options using a space (`--builder foo`) or an equals (`--builder=foo`). Many of the executables will accept both; some only spaces I think.

This issue is to make the docs consistent and figure out which executables don't accept both.

PR #1380, opened 2022-06-20
docs: consistent long arguments

j-ogas:
Addresses #641.

### DIFF
--- a/doc/ch-convert_desc.rst
+++ b/doc/ch-convert_desc.rst
@@ -31,7 +31,7 @@ producing the final format actually needed.
   :code:`-h`, :code:`--help`
     Print help and exit.
 
-  :code:`-i`, :code:`--in-fmt FMT`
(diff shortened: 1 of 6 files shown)


doc/ch-convert_desc.rst
```diff
@@ -31,7 +31,7 @@ producing the final format actually needed.
   :code:`-h`, :code:`--help`
     Print help and exit.
 
-  :code:`-i`, :code:`--in-fmt FMT`
+  :code:`-i FMT`, :code:`--in-fmt=FMT`, :code:`--in-fmt FMT`
```
reidpr, line 34: For documentation, I'd like to keep the short-hand that omits the argument for the short version; IMO that's sufficiently implied.

Similarly, I'd like to document either the equals or space, but not both, for conciseness. (My own quick impression is that I like the equals better, but I'm open to argument.) We could note somewhere that long arguments are documented one way but actually accept both.

reidpr (changes requested):

j-ogas:
Stale. Closing for now.

Issue #779, opened 2020-06-22
ch-grow: support ARG before FROM


reidpr:

reidpr:
Also trying to use a `variable` in `FROM` is a syntax error:

```
$ printf 'ARG OS\nFROM ${OS}\n' | ch-image build -t foo --build-arg OS=alpine:3.9 -f - /tmp
error: can't parse: -:2,6

FROM ${OS}
     ^
```

reidpr:
This [Github issue comment](https://github.com/moby/moby/issues/37345#issuecomment-400245466) says that global `ARG` (i.e., before `FROM`) applies to all stages, which complicates things.

reidpr:
Maybe the way to address this is to give `FROM` a way to specify arguments directly, e.g.:

```
```

PR #1395, opened 2022-06-29
ch-image build: support ARG before FROM

mphinney1100:
Addresses #779


lib/build.py
```diff
@@ -250,10 +250,16 @@ def __init__(self, tree):
...
+         if (k == "arg"):
+            try:
+               self.options[k] = self.options[k] + "=" + v
+            except KeyError:
+               self.options[k] = v
+         else:
+            if (k in self.options):
+               ch.FATAL("%3d %s: repeated option --%s"
+                        % (self.lineno, self.str_name, k))
+            self.options[k] = v
```
reidpr, line 262: I think this logic might be better in the grammar. It could grow two flavors of options: normal (what we had before) and key-value (or a better name?), that have a value that itself consists of a key/value pair. Then all this code has to do is pick out things from the parse.

lib/build.py
```diff
@@ -832,12 +838,21 @@ def __init__(self, *args):
...
+      image_ref = ch.Image_Ref(ch.tree_child(self.tree, "image_ref"))
+      arg_check = self.options.pop("arg", "False")
+      if (arg_check):
+         self.arg = arg_check.split("=")
+      image_ref.host = self.update_var(image_ref.host)
+      image_ref.port = self.update_var(image_ref.port)
+      image_ref.name = self.update_var(image_ref.name)
+      image_ref.tag = self.update_var(image_ref.tag)
+      image_ref.digest = self.update_var(image_ref.digest)
+      self.base_image = ch.Image(image_ref)
```
reidpr, line 855: 1. There's a choice here that we ought to make explicit: that variable substitution happens (1) before or (2) after the image ref is split into parts. I think Option 1 might be quite tricky to express in the grammar, though it's more expressive.
2. Is there a way to use the existing `variables_sub` method?
3. I think you can use a loop instead of several individual calls.

lib/build.py
```diff
@@ -918,14 +944,21 @@ def prepare(self, miss_ct):
...
       # Load metadata
       self.image.metadata_load(self.base_image)
+
+      # set --arg 
+      while (len(self.arg) > 1):
+         k = self.arg.pop(0)
+         v = self.arg.pop(0)
+         value = variables_sub(k, self.env_build)
+         ch.VERBOSE("setting %s to %s" % (k, v))
+         self.env_arg[k] = v
```
reidpr, line 954: We should comment why the variables are being substituted twice.

lib/build.py
```diff
@@ -918,14 +944,21 @@ def prepare(self, miss_ct):
...
+         ch.VERBOSE("setting %s to %s" % (k, v))
+         self.env_arg[k] = v 
       # Done.
       return int(self.miss)  # will still miss in disabled mode
 
    def execute(self):
       # Everything happens in prepare().
       pass
 
-
```
reidpr, line 928: Should leave this blank line.

lib/charliecloud.py
```diff
@@ -140,10 +140,10 @@ class Download_Mode(enum.Enum):
...
 ir_tag: ":" IR_TAG
 ir_digest: "@sha256:" HEX_STRING
-IR_HOST: /[A-Za-z0-9_.-]+/
-IR_PORT: /[0-9]+/
-IR_PATH_COMPONENT: /[a-z0-9_.-]+/
-IR_TAG: /[A-Za-z0-9_.-]+/
+IR_HOST: /[$A-Za-z0-9_.-]+/
+IR_PORT: /[$0-9]+/
+IR_PATH_COMPONENT: /[$a-z0-9_.-]+/
+IR_TAG: /[$A-Za-z0-9_.-]+/
```
reidpr, line 146: Suggest putting the `$` with the special characters at the end, so it doesn't look like a `$A` variable substitution. (Note the hyphen has to remain last so it's not misinterpreted as a range.)

test/build/50_dockerfile.bats
```diff
@@ -736,6 +746,36 @@ EOF
     else
         [[ $status -eq 0 ]]
     fi
+
+    # --arg not used in from
+    run ch-image build -v -t tmpimg -f - . <<'EOF'
+FROM --arg=foo=bar 00_tiny
```
reidpr, line 752: This disagrees with the comment?

reidpr (changes requested):
Great work! Some suggestions in-line.

A general thing to add: We should document how `FROM --arg` works. Maybe it goes in the compatibility with other interpreters section?

lib/build.py
```diff
@@ -254,6 +254,18 @@ def __init__(self, tree):
...
                      % (self.lineno, self.str_name, k))
          self.options[k] = v
+
+      for st in ch.tree_children(tree, "option_keypair"):
+         k = ch.tree_terminal(st, "OPTION_KEY")
+         s = ch.tree_terminal(st, "OPTION_VAR")
+         v = ch.tree_terminal(st, "OPTION_VALUE")
+         # assuming all key pair options allow multiple options
+         ch.INFO("setting %s to %s" % (s, v))
+         try:
```
reidpr, line 264: I'm not personally a fan of the Python ask-forgiveness approach, because it seems hard to reason about whether the `KeyError` came from the dictionary lookup I was intending. So, in this case I'd do `if (k in options):`.

However: I think here you can do a one-liner:

```
self.options.setdefault(k, {})[s] = v
```

lib/build.py
```diff
@@ -832,12 +844,14 @@ def __init__(self, *args):
 class I_from_(Instruction):
 
    __slots__ = ("alias",
+                "arg",
```
reidpr, line 847: Style: I recommend `args` (plural) because it's a collection.

lib/build.py
```diff
@@ -832,12 +844,14 @@ def __init__(self, *args):
...
                 "base_image")
 
    def __init__(self, *args):
       super().__init__(*args)
-      self.base_image = ch.Image(ch.Image_Ref(ch.tree_child(self.tree,
-                                                            "image_ref")))
+      self.arg = self.options.pop("arg", None)
```
reidpr, line 852: Give a default of the empty dictionary and you won't need `if (... not None)` below.

lib/build.py
```diff
@@ -918,6 +932,13 @@ def prepare(self, miss_ct):
          ch.WARNING("base image also exists non-cached; using cache")
       # Load metadata
       self.image.metadata_load(self.base_image)
+
+      # set --arg
+      if (self.arg is not None):
+         for var in self.arg:
+            val = self.arg.get(var)
+            ch.VERBOSE("setting %s to %s" % (var, val))
+            self.env_arg[var] = val
```
reidpr, line 941: I think you can replace this with `self.env_arg.update(...)`.

lib/charliecloud.py
```diff
@@ -190,8 +190,10 @@ class Download_Mode(enum.Enum):
 /// Common ///
 
 option: "--" OPTION_KEY "=" OPTION_VALUE
+option_keypair: "--" OPTION_KEY "=" OPTION_VAR "=" OPTION_VALUE
```
reidpr, line 193:
```suggestion
option_keypair: "--" OPTION_KEY "=" OPTION_KEY "=" OPTION_VALUE
```

lib/charliecloud.py
```diff
@@ -190,8 +190,10 @@ class Download_Mode(enum.Enum):
...
 OPTION_KEY: /[a-z]+/
-OPTION_VALUE: /[^ \t\n]+/
+OPTION_VALUE: /[^= \t\n]+/
```
reidpr, line 195: I think you can actually leave `OPTION_VALUE` unchanged.

lib/charliecloud.py
```diff
@@ -849,7 +851,8 @@ class Image_Ref:
                 "path",
                 "name",
                 "tag",
-                "digest")
+                "digest",
+                "dict")
```
reidpr, line 855: Use a more specific name than `dict`.

lib/charliecloud.py
```diff
@@ -894,6 +898,11 @@ def parse(class_, s):
                                    parser="earley", propagate_positions=True)
       s = s.replace("%", "/").replace("+", ":")
       hint="https://hpc.github.io/charliecloud/faq.html#how-do-i-specify-an-image-reference"
+      while (s.find("$") > -1): #variable substitution
+         start = s.find("$")
+         end = s.find(":", start)
+         str = s[start, end - 1]
+         s.replace(str, self.dict.get(str[1: -1]))
```
reidpr, line 905: I think you can use the existing `variables_sub()` here? You might need to move it to `charliecloud.py` and update existing callers.

lib/charliecloud.py
```diff
@@ -979,21 +988,31 @@ def defaults_add(self):
...
 
+   def var_sub(self, var):
+      if (var is not None and var[0] == "$"):
+         key = var[1:len(var)]
+         pair = self.dict.get(key)
+         if (pair is not None):
+            return pair
+         else:
+            ch.FATAL("%s is not set" % key)
+      return var
```
reidpr, line 1015: Use existing `variables_sub()` here too?

reidpr (commented):
Nice! Some comments in-line.

I wonder if it's time to update the grammar to add `ARG` before `FROM`?

lib/charliecloud.py
mphinney1100: `variables_sub()` isn't able to find the values. Should I try to tweak `variables_sub()` and make it work for what I want it to do or make something new?

lib/charliecloud.py
mphinney1100: the `=` is so it is able to find multiple keypair options. When `OPTION_VALUE` is unchanged the 2nd/3rd/etc... options don't get parsed the correct way. Is there a different way to get that working to have `OPTION_VALUE` unchanged?

lib/charliecloud.py
reidpr: That's surprising to me. Yes, I think changing `variables_sub()` is the way to go.

lib/build.py
```diff
@@ -381,6 +396,7 @@ def init(self, parent):
          self.image = self.parent.image
          self.image_alias = self.parent.image_alias
          self.image_i = self.parent.image_i
+         self.argfrom = self.parent.argfrom
```
reidpr, line 399: Is it possible to make the argfrom an attribute of the Dockerfile? Does that make it a global?

lib/charliecloud.py
```diff
@@ -979,21 +987,26 @@ def defaults_add(self):
...
       if (    self.host is not None
           and "." not in self.host
           and self.port is None):
          self.path.insert(0, self.host)
          self.host = None
 
+   def var_sub(self, var):
+      if (var is not None and var[0] == "$"):
+         return variables_sub(var, self.variables)
+      return var
```
reidpr, line 1009: If this method isn't called anywhere else, it could be a function defined within `from_tree()`.

test/build/50_dockerfile.bats
```diff
@@ -267,15 +267,6 @@ EOF
...
 
-    # ARG before FROM
-    run ch-image build -t tmpimg -f - . <<'EOF'
-ARG foo=bar
-FROM 00_tiny
-EOF
-    echo "$output"
-    [[ $status -eq 0 ]]
-    [[ $output = *'warning: ARG before FROM not yet supported; see issue #779'* ]]
-
```
reidpr, line 278: Need a replacement `ARG` before `FROM` test that checks if it works.

doc/ch-image_desc.rst
```diff
@@ -546,6 +546,13 @@ environment:
 Note that :code:`ARG` and :code:`ENV` have different syntax despite very
 similar semantics.
 
+:code:`ARG` before :code:`FROM`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+:code:`ARG` before :code:`FROM` is not currently supported but the :code:`--arg`
+option in the :code:`FROM` command has the same functionality.
+
+Note :code:`--arg` can be used multiple times in a single :code:`FROM` line. 
+
```
reidpr, line 555: This should probably be removed, and `--arg` documented within `FROM`?

reidpr (changes requested):
Nice work! Some suggestions in-line.

reidpr (changes requested):
Great work per usual. There is one fairly significant bug that we discussed offline (and may even have been the result of one of my suggestions 😂). I did tidy; hopefully I didn't break anything too badly.

reidpr (changes requested):
Discussed offline.

reidpr (approved):
LGTM! If you're OK with my tidy and it passes CI I'll merge.

mphinney1100:
Looks good!!

PR #1399, opened 2022-07-06
add python requirements.txt

MFTabriz:
from: https://github.com/hpc/charliecloud/blob/2ac3705ba05173bf471557b7a1cfb931c3c0f636/configure.ac#L396


reidpr:
Thanks for the contribution!

> * Do you think we can add requirements.txt?

I'm fine with it. I would like to put it in `packaging` though, and add a comment explaining what is is.

It's not something we use, so it would need either (1) a comment saying it's untested and who, if anyone, is in charge of maintaining it, or (2) a test to make sure it stays in sync with `configure.ac`, which is the source of truth for version requirements. 

> * Do you know why CI tests are failing? The only "not ok" I can find in the logs is:

I'm guessing it's spurious. Problems with RPM repos are not uncommon.

MFTabriz:
@reidpr what do you think about this solution?

I wired up the contents of new file to `configure.ac` so it gets used (and tested) during the configure time. I also added some comments to the file explaining what it is and how users/developers can use it. As the `requirements.txt` file is traditionally kept at the root directory of a project, I could not find a good way to add this file to the packaging code list in `misc/loc`. Feel free to modify the PR as you wish.

MFTabriz:
by the way, do we really need to install `lark-parser` and `wheel` python packages [here](https://github.com/hpc/charliecloud/blob/1329106d4998ade8d19a4147793135632f1b754f/.github/workflows/main.yml#L146)? If so, we must define a minimum version for each and move them to requirements.txt as well.

reidpr:
> However, `lark-parser` is probably leftover from before we embedded it.

This turns out to be untrue; see (closed) PR #1409.

reidpr:
Is this PR ready for review again?

MFTabriz:
@reidpr Yes it's ready! Sorry for the delay. I guess at the end it boils down to a simple file in the packaging folder.

reidpr (approved):
Thank you @MFTabriz! I added a support note. If you are OK with it, I will merge.

MFTabriz:
@reidpr This looks great! I'll use it in our builds of the Charliecloud (and hopefully the others will too) to make sure it's always up to date.

PR #1408, opened 2022-07-09
implement `ch-image modify`

### LINKED ISSUE #1400: support interactive image build
Some users prefer working with containers interactively, using `docker -it` to build a container with commands issued interactively, then pushing that container to a registry for deployment in production.

It would be useful if one of two things were different about charliecloud:
1. `ch-run -i` functionality, where interactive changes to a container capture provenance in the build cache.
2.  Docker build-cache to Charliecloud build-cache translation. If this were supported, users could work in Docker and pull a charliecloud container with minimal loss of information about build provenance. By supporting build-cache translation, we would have provenance similar to a Dockerfile.

"One" seems better/easier to me, but I think either of these solutions would support the `docker -it` development workflow.

reidpr:
Closes #1400.


reidpr:
@qwofford, what do you think of this proposal?

Alternate names include *deform*, *edit*, and *mutate*.

It would be nice not to implement the save prompt, i.e., only have the behavior with `--yes` as currently written and save or not depending on shell exit status.

doc/ch-image_desc.rst
```diff
@@ -856,6 +856,80 @@ functionally identical when re-imported.
...
+edit the image interactively. It is similar to a :code:`RUN` instruction that
+starts an interactive shell. By default, ask the user whether to save changes
+when the shell exits.
+
+Options:
+
+  :code:`-m MSG`
+    Use :code:`MSG` to identify the edits to the build cache. That is, if you
+    run this command twice with the same :code:`TARGET`, the same :code:`-o
+    DEST`, and the same :code:`MSG`, the second session will overwrite the
```
qwofford, line 884: What is `DEST`, in this case?

qwofford:
I think this might be a slightly greedy hope...but in reference to the warning block of text here...it would be really nice (and unique among container runtimes) if we somehow stored a series of Dockerfile commands that represents the build cache too. Anything from a Dockerfile could be stored directly, and then anything from a `ch-image modify` would get pre-pended with `RUN`. That way you could build up a container interactively, then do something like a `ch-image to-dockerfile DEST` and snip out any commands that weren't ultimately important.

This could get us the interactive functionality we want without diverging too far from best practices at the end of the day.

reidpr:
> if we somehow stored a series of Dockerfile commands that represents the build cache too

This is a cool idea. I can think of a couple approaches:

1. Horrible, nearly useless shell: Charliecloud interactively collects a command line, then executes it. Simple to implement, but loses essentially all shell features invented since 1975.

2. Munge Bash history variables and commands: Run a normal Bash shell, which can capture its own command history. Set that up so we can translate that into a Dockerfile. This would only work with specifically supported shells. I don't know how hard it would be to do this.

lucaudill:
@reidpr I'm ready for you to have a look at my progress so far

lib/build.py
```diff
@@ -287,6 +305,129 @@ def unescape(sl):
...
+   # FIXME: This is super kludgey
+   cli.tag = cli.out
+
+   print(cli.image_ref)
+   ch.ILLERI(cli.c)
+   ch.ILLERI(type(cli.c))
+   commands = []
+   # “Flatten” commands array
+   for c in cli.c:
+      commands += c
```
reidpr, line 325: The need for this will go away if you remove `nargs="*"`.

lib/build.py
```diff
@@ -287,6 +305,129 @@ def unescape(sl):
...
+               from_
+                  image_ref
+                     IMAGE_REF <image_name>
+               run
+                  run_shell
+                     LINE_CHUNK <command_1>
+               [...]
+               run
+                  run_shell
+                     LINE_CHUNK <command_N>
```
reidpr, line 398: Probably this example can be omitted; the second example covers it IMO.

lib/image.py
```diff
@@ -947,3 +949,4 @@ def terminals_cat(self, tname):
       """Return the concatenated values of all child terminals named tname as
          a string, with no delimiters. If none, return the empty string."""
       return "".join(self.terminals(tname))
+
```
reidpr, line 952: I don't see any substantive changes in this file?

reidpr (commented):
Nice work! I think next steps are:

1. Write docs describing the behavior we've discussed and you've implemented, including particularly the different behaviors depending on where the input is coming from.
2. Make the cache behavior like import.

doc/ch-image.rst
```diff
@@ -2106,6 +2106,195 @@ in the remote registry, so we don’t upload it again.)
 Delete all images and cache from ch-image builder storage.
 
 
+:code:`shell`
```
reidpr, line 2109: So this proposes calling it `ch-image shell` rather than `ch-image modify`. I’m open to argument but this seemed more specific.

lib/build.py
```diff
@@ -274,6 +268,147 @@ def build_arg_get(arg):
 
 ## Functions ##
 
+def modify(cli_):
+   # In this file, “cli” is used as a global variable
+   global cli
+   cli = cli_
+
+   # This file assumes that global cli comes from the “build” function. If we
+   # don’t assign these values, it casues problems.
```
reidpr, line 277: Be more specific about “problems” and how the assigned values were selected. Could also just solve the problems rather than assigning dummy values.

lib/build.py
```diff
@@ -274,6 +268,147 @@ def build_arg_get(arg):
...
+   out_image = im.Image(im.Reference(cli.out_image))
+   if (not src_image.unpack_exist_p):
+      ch.FATAL("not in storage: %s" % src_image.ref)
+   if (cli.out_image == cli.image_ref):
+      ch.FATAL("output must be different from source image (%s)" % cli.image_ref)
+
+   # This kludge is necessary because cli is a global variable, with cli.tag
+   # assumed present elsewhere in the file. cli.tag represents the image being
+   # built, which in our case can either be the source image or the output image
+   # (if specified).
```
reidpr, line 297: This comment needs updating, no?

lib/build.py
```diff
@@ -274,6 +268,147 @@ def build_arg_get(arg):
...
+   # Second condition here is to ensure that “commands” does’t get overwritten
+   # in the case where “ch-image modify” isn’t called from a terminal session
+   # (e.g. in Github actions). I’m considering this a temporary fix, although I
+   # think a case could also be made for “-c” to have precedence over pipeline
+   # input.
+   if ((not sys.stdin.isatty()) and (commands == [])):
+      # Treat stdin as opaque blob and run that
+      commands = [sys.stdin.read()]
+   # Again, the fact that we have to check against both “[]” and “['']” is to
+   # get CI to work...
```
reidpr, line 313: One-element list containing the empty string seems very weird.

lib/build.py
```diff
@@ -274,6 +268,147 @@ def build_arg_get(arg):
...
+   if (commands not in [[],['']]):
+      tree = modify_tree_make(src_image.ref, commands)
+
+      # Count the number of stages (i.e., FROM instructions)
+      global image_ct
+      image_ct = sum(1 for i in tree.children_("from_"))
+
+      ml = traverse_parse_tree(tree)
+   else:
```
reidpr, line 322: Should label the branches on which of the three modes we’re in.

lib/build.py
```diff
@@ -274,6 +268,147 @@ def build_arg_get(arg):
...
+      fake_sid = uuid.uuid4()
+      out_image.unpack_clear()
+      out_image.copy_unpacked(src_image)
+      bu.cache.worktree_adopt(out_image, src_image.ref.for_path)
+      bu.cache.ready(out_image)
+      bu.cache.branch_nocheckout(src_image.ref, out_image.ref)
+      foo = subprocess.run([ch.CH_BIN + "/ch-run", "--unsafe", "-w",
+                            str(out_image.ref), "--", shell])
+      # FIXME: This causes issues when you change the value in ch_misc.h and
+      #        forget to change it here...
```
reidpr, line 335: What kind of issues?

lib/build.py
```diff
@@ -274,6 +268,147 @@ def build_arg_get(arg):
...
+      if (foo.returncode == 49):
```
reidpr, line 336: Avoid magic number 49.

lib/build.py
```diff
@@ -274,6 +268,147 @@ def build_arg_get(arg):
...
+   # has been given a “line” value will result in the attribute not being present,
+   # which causes an error when we try to access that attribute. Here we give the
+   # attribute a debug value of -1 to avoid said errors.
+   meta = lark.tree.Meta()
+   meta.line = -1
+   #df_children.append(im.Tree(lark.Token('RULE', 'from_'), [im.Tree(lark.Token('RULE', 'image_ref'),[lark.Token('IMAGE_REF', src_img.name)], meta)], meta))
+   df_children.append(im.Tree(lark.Token('RULE', 'from_'), [im.Tree(lark.Token('RULE', 'image_ref'),[lark.Token('IMAGE_REF', str(src_img))], meta)], meta))
+   for cmd in cmds:
+      df_children.append(im.Tree(lark.Token('RULE', 'run'), [im.Tree(lark.Token('RULE', 'run_shell'),[lark.Token('LINE_CHUNK', cmd)], meta)],meta))
+   return im.Tree(lark.Token('RULE', 'start'), [im.Tree(lark.Token('RULE','dockerfile'), df_children)], meta)
```
reidpr, line 383: Overlong lines.

test/run/ch-run_misc.bats
```diff
@@ -312,118 +316,127 @@ EOF
     # no argument to --bind
     run ch-run "$ch_timg" -b
     echo "$output"
+    echo "STATUS"
+    echo "$status"
+    echo "STATUS"
```
reidpr, line 321: leftover debug code?

bin/ch_misc.c
```diff
@@ -586,6 +586,16 @@ void msg(enum log_level level, const char *file, int line, int errno_,
    va_end(ap);
 }
 
+void msg_error(const char *file, int line, int errno_,
+               const char *fmt, ...)
+{
+   va_list ap;
+
+   va_start(ap, fmt);
+   msgv(LL_FATAL, file, line, errno_, fmt, ap);
```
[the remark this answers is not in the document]
lucaudill: Right now the only way this function is being used is as an alternative for `msg_fatal` in the case where you want to specify an exit status other that `ERR_CHRUN`. I'm a little hesitant to treat it differently than `msg_fatal`.

lib/build.py
```diff
@@ -274,6 +268,215 @@ def build_arg_get(arg):
...
+
+   # This kludge is necessary because cli is a global variable, with cli.tag
+   # assumed present elsewhere in the file. Here, cli.tag represents the
+   # destination image.
+   cli.tag = str(out_image)
+
+   stdin = sys.stdin.read()
+
+   # We check that stdin isn’t None to ensure that we don’t go down this code
+   # path by mistake (e.g. in CI, where stdin will never by a TTY).
```
reidpr, line 307: CI might be able to use `-i` to bypass the TTY test.

lib/build.py
```diff
@@ -274,6 +268,215 @@ def build_arg_get(arg):
...
+   if ((not sys.stdin.isatty()) and (commands == []) and (stdin != '')):
+      # https://stackoverflow.com/a/6482200
+
+      # We use “decode("utf-8")” here because stdout seems default to a bytes
+      # object, which is not a valid type for an argument for “Path”.
+      tmpfile = ch.Path(subprocess.run(["mktemp", "-d"],capture_output=True).stdout.decode("utf-8"))
```
reidpr, line 313: There’s a Python module for this; you don’t need a subprocess.

lib/build.py
```diff
@@ -274,6 +268,215 @@ def build_arg_get(arg):
...
+      with open(tmpfile, "w") as outfile:
+         outfile.write(stdin)
+      # By default, the file is seemingly created with its execute bit
+      # unassigned. This is problematic for the RUN instruction.
```
reidpr, line 317: The script doesn’t actually have to be executable if it’s a file argument to the shell.

lib/build.py
```diff
@@ -274,6 +268,215 @@ def build_arg_get(arg):
...
+   if ((commands != []) or (cli.script is not None)):
+      # non-interactive case (“-c” or script)
+      if (cli.script is not None):
+         tree = modify_tree_make_script(src_image.ref, cli.script)
+      else:
+         tree = modify_tree_make(src_image.ref, commands)
+
+      # Count the number of stages (i.e., FROM instructions)
+      global image_ct
+      image_ct = sum(1 for i in tree.children_("from_"))
```
reidpr, line 335: Can we let `main()` do this? It’s kind of hairy code to duplicate.

lib/build.py
```diff
@@ -274,6 +268,215 @@ def build_arg_get(arg):
...
+      ch.VERBOSE("using SID %s" % fake_sid)
+      # FIXME: metadata history stuff? See misc.import_.
+      if (out_image.metadata["history"] == []):
+         out_image.metadata["history"].append({ "empty_layer": False,
+                                                "command":     "ch-image import"})
+      out_image.metadata_save()
+      bu.cache.commit(out_image.unpack_path, fake_sid, "MODIFY interactive", [])
+
+def modify_tree_make(src_img, cmds):
+   """Function that manually constructs a parse tree corresponding to a set of
```
reidpr, line 364: Mood:

```suggestion
   """Construct a parse tree corresponding to a set of
```

lib/build.py
```diff
@@ -274,6 +268,215 @@ def build_arg_get(arg):
...
+   # Metadata attribute. We use this attribute in the “_pretty” method for our
+   # “Tree” class. Constructing a tree without specifying a “Meta” instance that
+   # has been given a “line” value will result in the attribute not being present,
+   # which causes an error when we try to access that attribute. Here we give the
+   # attribute a debug value of -1 to avoid said errors.
+   meta = lark.tree.Meta()
+   meta.line = -1
+   df_children.append(im.Tree(lark.Token('RULE', 'from_'),
+                      [im.Tree(lark.Token('RULE', 'image_ref'),
+                      [lark.Token('IMAGE_REF', str(src_img))], meta)], meta))
```
reidpr, line 397: Indent to reflect structure, e.g. something like:

```suggestion
   df_children.append(im.Tree(lark.Token('RULE', 'from_'),
                      [im.Tree(lark.Token('RULE', 'image_ref'),
                        [lark.Token('IMAGE_REF', str(src_img))],
                        meta)
                      ], meta))
```

lib/build.py
```diff
@@ -274,6 +268,215 @@ def build_arg_get(arg):
...
+      df_children.append(im.Tree(lark.Token('RULE', 'shell'),
+                         [lark.Token('STRING_QUOTED', '"%s"' % cli.shell),
+                         lark.Token('STRING_QUOTED', '"-c"')],meta))
+   for cmd in cmds:
+      df_children.append(im.Tree(lark.Token('RULE', 'run'),
+                         [im.Tree(lark.Token('RULE', 'run_shell'),
+                         [lark.Token('LINE_CHUNK', cmd)], meta)],meta))
+   return im.Tree(lark.Token('RULE', 'start'), [im.Tree(lark.Token('RULE','dockerfile'), df_children)], meta)
+
+# FIXME: Combine with “modify_tree_make”?
```
reidpr, line 408: Yeah probably.

lib/build.py
```diff
@@ -274,6 +268,215 @@ def build_arg_get(arg):
...
+# Tree.iter_trees_topdown(), which *is* documented to be in-order [3].
+#
+# This change seems to have been made in 0.8.6 (see PR #761); before then,
+# visit() was in order. Therefore, we call that instead, if visit_topdown() is
+# not present, to improve compatibility (see issue #792).
+#
+# [1]: https://lark-parser.readthedocs.io/en/latest/visitors/#visitors
+# [2]: https://github.com/lark-parser/lark/blob/445c8d4/lark/visitors.py#L211
+# [3]: https://lark-parser.readthedocs.io/en/latest/classes/#tree
+def traverse_parse_tree(tree):
```
reidpr, line 468: Most significant first?

```suggestion
def parse_tree_traverse(tree):
```

lib/pull.py
```diff
@@ -106,7 +106,7 @@ def download(self):
                # currently, this error is only raised if we’ve downloaded the
                # skinny manifest.
                have_skinny = True
-               if (ch.arch == "amd64"):
+               if (ch.arch in ["amd64", "aarch64"]):
```
reidpr, line 109: I think we talked about this? The issue was that your ARM VM was set up to silently execute x86-64 code via QEMU.

reidpr (commented):
Solid. In particular the clear, well-written comments continue.

I think my main criticism is that the logic of selecting the mode is mixed in with carrying out that mode. Is there a way to just (1) figure out the mode, then (2) do the setup (build tree or run interactive session, then (3) call `build.main()` or `misc.import()`? Possibly in a new `modify.py`?

Also I suspect you should validate (a) cache behavior and (b) parse trees.

test/build/50_dockerfile.bats
```diff
@@ -1160,7 +1160,8 @@ EOF
     [[ $status -ne 0 ]]
     [[    $output = *'error: no context because '?'-'?' given'* \
        || $output = *'COPY failed: file not found in build context or'* \
-       || $output = *'no such file or directory'* ]]
+       || $output = *'no such file or directory'* \
+       || $output = *'failed to compute cache key: failed to calculate checksum of ref'* ]]
```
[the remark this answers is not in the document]
lucaudill: No. I believe we had an offline discussion about adding this error to possible outputs here.

lib/build.py
lucaudill: This comment is stale, we discussed this offline at some point and I believe we decided against it. I'd be happy to discuss further if you feel differently now.

lib/modify.py
```diff
@@ -0,0 +1,225 @@
...
+      if (cli.interactive):
+         ch.FATAL("incompatible opts: “-c”, “-i”")
+      if (cli.script is not None):
+         ch.FATAL("script mode incompatible with command mode")
+      mode = Modify_Mode.COMMAND_SEQ
+   elif (cli.script is not None):
+      if (cli.interactive):
+         ch.FATAL("script mode incompatible with interactive mode")
+      mode = Modify_Mode.SCRIPT
+   elif (sys.stdin.isatty() or (cli.interactive)):
```
reidpr, line 68:
```suggestion
   elif (sys.stdin.isatty() or cli.interactive):
```

lib/modify.py
```diff
@@ -0,0 +1,225 @@
...
+   # has been given a “line” value will result in the attribute not being present,
+   # which causes an error when we try to access that attribute. Here we give the
+   # attribute a debug value of -1 to avoid said errors.
+   meta = lark.tree.Meta()
+   meta.line = -1
+   df_children.append(im.Tree(lark.Token('RULE', 'from_'),
+                      [im.Tree(lark.Token('RULE', 'image_ref'),
+                        [lark.Token('IMAGE_REF', str(src_img))],
+                        meta)
+                      ], meta))
```
reidpr, line 165: Helper function might be clearer here?

reidpr (commented):
Great work; thanks Lucas. A few nits but I’m not sure it’s worth holding up merge over those?

PR #1409, opened 2022-07-11
rm unneeded CI install of Lark

reidpr:


reidpr:
Oh wait we do use this Lark for the external Lark test; never mind. 🙄

PR #1410, opened 2022-07-12
kill Git garbage collection on bucache before deleting

### LINKED ISSUE #1406: cache reset while garbage collection is running crashes and corrupts storage directory
(we believe) After building an image there is a race condition that results from the cleaning up of the git repos. This results in the cleanup failing. This then subsequently causes any attempts to build things to fail, as the images are not in the cache properly.

```
deleting build cache
error: can't recursively delete directory /var/tmp/bstormer.ch/bucache: gc.pid: No such file or directory
initializing empty build cache
 12. FROM ubuntu:20.04
manifest list: downloading: 100%
manifest: downloading: 100%
config: using existing file
layer 1/1: d7bfe07: using existing file
error: command failed with code 128: git checkout -B 'ubuntu+20.04#' root
 12. FROM ubuntu:20.04
manifest list: downloading: 100%
manifest: downloading: 100%
config: using existing file
layer 1/1: d7bfe07: using existing file
error: command failed with code 128: git checkout -B 'ubuntu+20.04#' root
```
The command `ch-image build-cache --reset && ch-image build --force -f Dockerfile.almalinux8` with things in the cache already should produce this error.

reidpr:
Closes #1406.

### DIFF
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -19,6 +19,7 @@ import pprint
 import re
 import shlex
 import shutil
+import signal
 import stat
 import subprocess
 import sys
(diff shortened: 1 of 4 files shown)


test/build/55_cache.bats
```diff
@@ -1094,3 +1094,42 @@ RUN true        # miss
...
+
+    # Build an image that’s going to be annoying to garbage collect, but not
+    # too annoying, so the test isn’t too long. Keep in mind this is probably
+    # happening on a tmpfs.
+    ch-image build -t tmpimg - <<'EOF'
+FROM alpine:3.9
+RUN for i in $(seq 0 1024); do \
+       dd if=/dev/urandom of=/$i bs=192K count=1 status=none; \
+    done
+#RUN ls -lh
```
j-ogas, line 1117:
```suggestion
```

lib/charliecloud.py
```diff
@@ -2524,6 +2525,29 @@ def json_from_file(path, msg):
...
+      FATAL("can’t signal PID %d with %d: %s" % (pid, sig, x.strerror))
+   for i in range(timeout*2):
+      try:
+         os.kill(pid, 0)  # no effect on process
+      except ProcessLookupError:  # done
+         return
+      except OSError as x:
+         FATAL("can’t signal PID %s with 0: %s" % (pid, x.strerror))
+      time.sleep(0.5)
+   FATAL("timeout of %ds exceeded trying to kill PID %d" % (timeout, pid))
```
j-ogas, line 2549: Perhaps include some kind of hint here?

j-ogas (changes requested):

j-ogas (approved):
LGTM.

PR #1415, opened 2022-07-22
add nested namespaces for non-setuid `fusermount3`

reidpr:
This PR lets `ch-run`'s internal SquashFS work with non-setuid `fusermount3`.

If that executable is installed setuid, suppress the setuid bit with `prctl(2)`, i.e., `ch-run` now *never* escalates during SquashFUSE mount.


reidpr:
This needs to be tested on CentOS 7.

mphinney1100 (approved):
I'm so surprise how minimal the changes needed were for non-setuid.

reidpr:
Confirmed this works on CentOS 7.

PR #1423, opened 2022-08-05
give some base images clearer names

reidpr:
Alternate take on image name confusion describe in PR #1376.


reidpr:
@MFTabriz, what is your opinion on this approach to address the concern in PR #1376?

MFTabriz:
@reidpr Yes! This makes it much more clear. But I’m still not sure why (for example) we can’t merge the [Dockerfile.almalinux_8ch](https://github.com/hpc/charliecloud/blob/744e47cd97d44a8e05212b3fac1c6cdbe1280807/examples/Dockerfile.almalinux_8ch) in the [Dockerfile.mpich](https://github.com/hpc/charliecloud/blob/744e47cd97d44a8e05212b3fac1c6cdbe1280807/examples/Dockerfile.mpich) and just rely on `almalinux:8` as base.

reidpr:
It's because using `almalinux:8` as base required multiple changes in downstream Dockerfiles, some of which repeated things already in `almalinux8`. I.e., it's more than a drop-in substitution.

Chaining Dockerfiles like that is a common approach, FWIW. It's not just us doing that.

PR #1428, opened 2022-08-16
ch-image list: add last change timestamp to output

### LINKED ISSUE #1413: ch-image list show most recent modification
It would be handy to see a timestamp that reflects the most recently pulled image along with the output of ch-image list. Something like:

```
$ ch-image list
myregistry.com/my_container:mytag   | Thu Jul 21 09:07:10 MDT 2022
```

rbberger:
Implements a change to resolve #1413. 

Not sure if this is enough. Right now it will only show the time when the unpack directory was modified. That should be correct after a `pull`. But, e.g. after `ch-image build`, it will show the beginning of the build, not when it was completed. An alternative would be to take the timestamp from `ch/metadata.json`.


reidpr:
Sounds good! Greatly appreciate the contribution. If you like, also please add yourself to the contributors list in `README.rst`.

rbberger:
@reidpr done

reidpr:
@rbberger Thanks for this and thanks for waiting. I made some style tweaks. If this looks OK to you and passes tests, I'll merge.

PR #1451, opened 2022-09-20
build cache: metadata instructions only commit what’s needed

reidpr:
This PR optimizes metadata-only instructions (`ARG`, `ENV`, and `SHELL`) to skip the recursive image walk, instead processing only the file(s) that actually changed.

There are two more instructions with similar optimization opportunities not included here:

1. `WORKDIR` modifies only `ch/metadata.json` and (maybe) the directory operand. However, dealing with the directory is trickier than I thought. There are multiple cases to deal with, including at least: (a) all directories in specified path created by the instruction, (b) some directories created, (c) no directories created, (d) path traverses one or more symlinks. So, I'm putting it off.

2. `COPY` also modifies a known set of files. I didn't look into this in detail, but the number of files is going to be qualitatively different than the one or two so far.

Because of these future possible changes, `commit_files` is an instance attribute rather than a class attribute.

### DIFF
--- a/lib/build.py
+++ b/lib/build.py
@@ -234,13 +234,14 @@ class Main_Loop(lark.Visitor):
 
 class Instruction(abc.ABC):
 
-   __slots__ = ("git_hash",     # Git commit where sid was found
+   __slots__ = ("commit_files",  # modified files; default: anything
+                "git_hash",      # Git commit where sid was found
                 "image",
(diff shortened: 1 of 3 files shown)


j-ogas (approved):
LGTM. Are we opening separate issues for optimization opportunities 1 and 2?

PR #1459, opened 2022-09-26
don’t re-checkout if WD is already the right commit

reidpr:
Turns out that with very large images, `git checkout` can take a few minutes, so don’t do it if we don’t have to.


j-ogas (commented):
What impact, if any, does this have on images that break mid build, e.g., force canceling during a RPM install? Does the current workflow of deleting the bonked image and re-executing build still fix it?

reidpr:
There should be none. Also, we now roll back the image when an instruction fails, so that workaround should no longer be needed.

Issue #1341, opened 2022-05-16
update “environment variables” sections


reidpr:
The various commands have grown a number of environment variables that aren't recorded in these man page sections. Audit and update.

PR #1466, opened 2022-10-11
add debug environment variable

### LINKED ISSUE #1458: add environment variable equivalent to `--debug`
We should add an environment variable equivalent to the `--degbug` flag for `ch-image` and set it in the test suite. This would be helpful for tracking down bugs.

lucaudill:
closes #1458

### DIFF
--- a/bin/ch-test
+++ b/bin/ch-test
@@ -13,6 +13,8 @@ else
     export LC_ALL=C
 fi
 
+# Set environment variable for ch-image --debug flag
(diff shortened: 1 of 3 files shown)


lucaudill:
closes #1458

reidpr (approved):
Great work thanks! Docs for the new variable are missing, but that's true for lots of variables (#1341) and I'd like to get this merged.

PR #1467, opened 2022-10-11
`ch-image list` hash

### LINKED ISSUE #1365: `ch-image`: list sha256 hash of container
I like that `ch-image list <container>` shows information about a container both on the remote and in the local CH_IMAGE_STORAGE. I would like to see sha256 info about the remote and local containers too. For example:
```
$ ch-image list myregistry.com/group/tag
details of image:    myregistry.com/group/tag:latest
sha256 remote:       abcdef01234
sha256 local:        43210fedcba
...
```

This feature would allow someone to easily capture the state of a containerized application as the container is deployed. This feature is important to establish application provenance at runtime.

Current practice would require reading a manifest JSON out of CH_IMAGE_STORAGE, and then locating the correct manifest hash in the list of manifest hashes that define the state of the container. This seems too hard...

lucaudill:
closes #1365

### DIFF
--- a/lib/misc.py
+++ b/lib/misc.py
@@ -138,7 +138,12 @@ def list_(cli):
          pullet.fatman_load()
          remote = "yes"
          arch_aware = "yes"
-         arch_avail = " ".join(sorted(pullet.architectures.keys()))
+         digests = pullet.digests
+         arch_keys = sorted(pullet.architectures.keys())
+         max_key = max(arch_keys)
+         arch_avail = arch_keys[0] + " (" + pullet.digests[arch_keys[0]][:11] + ")\n"
+         for i in arch_keys[1:]:
+            arch_avail += (" " * 21) + i + " (" + pullet.digests[i][:11] + ")\n"
       except ch.Image_Unavailable_Error:
          remote = "no (or you are not authorized)"
          arch_aware = "n/a"
--- a/lib/misc.py
+++ b/lib/misc.py
@@ -140,10 +140,12 @@ def list_(cli):
          arch_aware = "yes"
          digests = pullet.digests
          arch_keys = sorted(pullet.architectures.keys())
-         max_key = max(arch_keys)
-         arch_avail = arch_keys[0] + " (" + pullet.digests[arch_keys[0]][:11] + ")\n"
-         for i in arch_keys[1:]:
-            arch_avail += (" " * 21) + i + " (" + pullet.digests[i][:11] + ")\n"
+         fmt_space = len(max(arch_keys,key=len)) + 1
+         # avert your eyes
+         arch_avail = [(" "*21*(i != arch_keys[0])) + "%-*s" % (fmt_space, i)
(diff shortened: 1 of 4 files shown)


lib/misc.py
```diff
@@ -138,20 +138,33 @@ def list_(cli):
          pullet.fatman_load()
          remote = "yes"
          arch_aware = "yes"
-         arch_avail = " ".join(sorted(pullet.architectures.keys()))
+         digests = pullet.digests
```
reidpr, line 141: unused?

lib/misc.py
```diff
@@ -138,20 +138,33 @@ def list_(cli):
...
+         try:
+            fmt_space = len(max(arch_keys,key=len)) + 1
+            arch_avail = []
+            for key in arch_keys:
+               arch_avail.append("%-*s" % (fmt_space, key) + "("
+                                 + pullet.digests[key][:11] + ")")
+         except ValueError:
+            # handles case where arch_keys is empty, e.g.
+            # mcr.microsoft.com/windows:20H2.
+            arch_avail = ['']
```
reidpr, line 152: `[None]` might be more idiomatic?

lib/misc.py
```diff
@@ -138,20 +138,33 @@ def list_(cli):
...
-         arch_avail = "unknown"
+         arch_avail = ["unknown"]
       pullet.done()
       print("available remotely:  %s" % remote)
       print("remote arch-aware:   %s" % arch_aware)
       print("host architecture:   %s" % ch.arch_host)
-      print("archs available:     %s" % arch_avail)
+      print("archs available:     %s" % arch_avail[0])
+      for arch in arch_avail[1:]:
+         print((" " * 21) + arch)
```
reidpr, line 167: clever!

lib/pull.py
```diff
@@ -161,6 +168,7 @@ def fatman_load(self):
       if (str(self.src_ref) in manifests_internal):
          # cheat; internal manifest library matches every architecture
          self.architectures = ch.Arch_Dict({ ch.arch_host: None })
+         self.digests[ch.arch_host] = manifests_internal[self.src_ref.name]['digest']
```
reidpr, line 169: You know, I wonder if the place to put in `"no digest"` is actually here?

lib/pull.py
```diff
@@ -195,6 +203,7 @@ def fatman_load(self):
          if (arch in self.architectures):
             ch.FATAL("manifest list: duplicate architecture: %s" % arch)
          self.architectures[arch] = ch.digest_trim(digest)
+         self.digests[arch] = digest[7:]
```
reidpr, line 204: If the slice is to remove the `sha256:`, I recommend you split on `:`, for greater clarity as well as insulation against the hash algorithm name length changing.

reidpr (changes requested):
Fantastic! Please see tidy commit. Also some comments in-line.

PR #1468, opened 2022-10-13
`ch-image build` context error message

### LINKED ISSUE #1440: unhelpful error message raised by ch-image build
Passing `ch-image build` a Dockerfile instead of a context directory raises an unhelpful error message:

```
$ ch-image build --debug test/Dockerfile.00_tiny 
inferred image name: dockerfile.00_tiny
error: can't open for rt: test/Dockerfile.00_tiny/Dockerfile: Not a directory
hint: charliecloud.py:2616:ossafe, charliecloud.py:2607:open_, build.py:116:main, ch-image:281:main
```
Revise this so the message gives the user a better idea of what they're doing wrong. Alternately, we could have Charliecloud assume that the context directory is the same as the one the file is located in.

Equivalent error using Docker:

```
$ sudo docker build test/Dockerfile.00_tiny
unable to prepare context: context must be a directory: /home/charlie/charliecloud/test/Dockerfile.00_tiny
```

lucaudill:
closes #1440

### DIFF
--- a/lib/build.py
+++ b/lib/build.py
@@ -110,6 +110,8 @@ def main(cli_):
    # Read input file.
    if (cli.file == "-" or cli.context == "-"):
       text = ch.ossafe(sys.stdin.read, "can't read stdin")
+   elif (not os.path.isdir(cli.context)):
+      ch.CONTEXT("context must be a directory: %s" % cli.context)
    else:
       fp = ch.Path(cli.file).open_("rt")
       text = ch.ossafe(fp.read, "can't read: %s" % cli.file)
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -2460,6 +2460,11 @@ class Timer:
 
 ## Supporting functions ##
 
+def CONTEXT(msg, hint=None, **kwargs):
+   # Used when context is not a directory.
+   log(msg, hint, "1;31m", "error: ", **kwargs) # bold red
+   sys.exit(1)
+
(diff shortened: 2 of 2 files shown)


lib/build.py
```diff
@@ -110,6 +110,8 @@ def build_arg_get(arg):
    # Read input file.
    if (cli.file == "-" or cli.context == "-"):
       text = ch.ossafe(sys.stdin.read, "can't read stdin")
+   elif (not os.path.isdir(cli.context)):
+      ch.CONTEXT("context must be a directory: %s" % cli.context)
```
reidpr, line 114: Why not just `ch.FATAL`?

reidpr (changes requested):

lib/build.py
lucaudill: When the `$CH_IMAGE_DEBUG` environment variable is set, `ch.FATAL` gives a stack trace. I thought this might be distracting/confusing to the user:
```
$ bin/ch-image build test/Dockerfile.00_tiny 
inferred image name: dockerfile.00_tiny
error: context must be a directory: test/Dockerfile.00_tiny
hint: build.py:115:main, ch-image:289:main
```
thoughts?

lucaudill: To avoid defining a new function, I could add a keyword argument to `ch.FATAL` that suppresses the stack trace. Or we could just assume that it's not really that distracting

reidpr: I think you're right that it's distracting, but folks will have asked for that.

PR #1469, opened 2022-10-13
`ARG` before `FROM` test: check correct line

reidpr:
Overlooked conflict during merge of PR #1466 and then PR #1395.


lucaudill (approved):
looks good to me!

PR #1472, opened 2022-10-18
authenticate registry before packaging

### LINKED ISSUE #1426: Fail faster for registry push auth
Gathering and preparing a container for upload can take a very long time, it would be good if auth was checked before this happens.

lucaudill:
closes #1426

### DIFF
--- a/lib/push.py
+++ b/lib/push.py
@@ -97,6 +97,9 @@ class Image_Pusher:
          There is not currently any support for re-using any previously
          prepared files already in the upload cache, because we don't yet have
          a way to know if these have changed until they are already build."""
+      # Initializing an HTTP instance for the registry and doing a 'GET'
+      # request right out the gate ensures the user needs to authenticate
+      # before we prepare the image for upload (#1426).
       self.registry = rg.HTTP(self.dst_ref)
       self.registry.request("GET", self.registry._url_base)
       tars_uc = self.image.tarballs_write(ch.storage.upload_cache)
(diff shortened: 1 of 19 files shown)


lib/push.py
```diff
@@ -95,6 +97,8 @@ def prepare(self):
          There is not currently any support for re-using any previously
          prepared files already in the upload cache, because we don't yet have
          a way to know if these have changed until they are already build."""
+      self.registry = rg.HTTP(self.dst_ref)
+      self.registry.request("GET", self.registry._url_base)
```
reidpr, line 101: Add a comment explaining what we're doing here, likely with a bug number.

lib/registry.py
```diff
@@ -365,10 +365,13 @@ def headers_log(hs):
...
 
+   @property
+   def _url_base(self):
+      return "https://%s:%d/v2/" % (self.ref.host, self.ref.port)
+   
    def _url_of(self, type_, address):
       "Return an appropriate repository URL."
-      url_base = "https://%s:%d/v2" % (self.ref.host, self.ref.port)
-      return "/".join((url_base, self.ref.path_full, type_, address))
+      return self._url_base + "/".join((self.ref.path_full, type_, address))
```
reidpr, line 374: This will add a redundant slash, no?

reidpr (changes requested):
Great work! Nice small, elegant solution. Couple of minor tweaks requested in-line.

lib/registry.py
lucaudill: I removed the base url from the `join()` method in attempt to avoid a redundant slash. Here's an example of what's going on in line 374
```
>>> "foo/" + "/".join(["bar", "baz"])
'foo/bar/baz
```

reidpr: Oh sorry I'm just wrong. Never mind.

PR #1497, opened 2022-11-22
skip test if glibc not compatible

lucaudill:
closes $1430

### DIFF
--- a/test/run/ch-fromhost.bats
+++ b/test/run/ch-fromhost.bats
@@ -34,10 +34,23 @@ fromhost_ls () {
     find "$1" -xdev -name '*sotest*' -ls
 }
 
+glibc_version_ok () {
+    # Ensure host glibc and guest glibc are compatible (see #1430).
+    host=$(ldd --version | head -1 | grep -o '....$')
+    guest=$(ch-run "$1" -- ldd --version | head -1 | grep -o '....$')
+    if [ $(printf $host'\n2.34' | sort -V | head -1) = '2.34' ] \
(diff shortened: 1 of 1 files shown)


test/run/ch-fromhost.bats
```diff
@@ -34,11 +34,24 @@ fromhost_ls () {
     find "$1" -xdev -name '*sotest*' -ls
 }
 
+glibc_version_ok () {
+    # Ensure host glibc and guest glibc are compatible (see #1430).
+    host=$(ldd --version | head -1 | grep -o '....$')
```
reidpr, line 39: A few suggestions here:

1. Consider how to re-write this regex so it doesn't assume a specific number of characters in the version number, but rather the content.
2. You can do this with a single `sed` invocation, but `head` and `grep` isn't horrible.
3. If you were to stick with `grep`, use `grep -E` for extended regex syntax (and because `egrep` will make ShellCheck yell at you).

test/run/ch-fromhost.bats
```diff
@@ -34,11 +34,24 @@ fromhost_ls () {
...
+    fi
+}
+
 @test 'ch-fromhost (CentOS)' {
     scope standard
     prerequisites_ok almalinux_8ch
     img=${ch_imgdir}/almalinux_8ch
 
+    # check glibc version compatibility.
+    glibc_version_ok "$img"
```
reidpr, line 53: We don’t need this for the Debian test also?

reidpr (changes requested):
Nice work. Some good shell programming happening here. Suggestions in-line.

PR #1498, opened 2022-11-22
omit `RPATH` when not needed

reidpr:
Previously, `configure` built `ch-run` with an empty `RPATH` if `libsquashfuse` was found in a default location (e.g., `/usr/local/lib` or `/usr/lib`) and `--with-squashfuse` was either not specified or given with `yes`.

This generated a `scanelf(1)` warning, likely due to a long history of exploits related to `RPATH` and `RUNPATH`. We were unable to identify any specific vulnerability related to the empty `RPATH` on modern systems. Regardless, it’s a bug and we’re fixing it.

Thanks to @olifre for identifying the bug, and to @mej and @najones19746 for analyzing its impact.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -362,7 +362,10 @@ AS_IF([   test $want_libsquashfuse = yes \
        && test $have_libsquashfuse_ll = yes \
        && test $have_ll_h = yes],
       [have_libsquashfuse=yes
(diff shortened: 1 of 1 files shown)


reidpr:
@olifre, does this PR address your warning?

olifre:
@reidpr Thanks a lot! Yes, indeed it does. I tested applying the patch and rebuilding within Gentoo's package management framework, and the warning by `scanelf` is gone for good :+1: .

PR #1499, opened 2022-11-23
friendly Docker Hub rate limit info

### LINKED ISSUE #1390: log Docker Hub rate limits at INFO level
Currently, we log the [Docker Hub rate limits](https://docs.docker.com/docker-hub/download-rate-limit/) headers as raw headers at level VERBOSE.

It would be better if we interpreted the headers and printed a friendlier message at INFO. Perhaps it should include:

1. Maximum pulls per time interval, `Ratelimit-Limit`
2. Remaining pulls, `Ratelimit-remaining`
3. Source of limit. This is a different header, and can be an IP address, a UUID which implies you're logged in and it's based on that user, or possibly something else (e.g., if the global limit has been reached).

lucaudill:
closes #1390

### DIFF
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -2144,29 +2144,24 @@ class Registry_HTTP:
       except requests.exceptions.RequestException as x:
          FATAL("%s failed: %s" % (method, x))
       # Docker Hub info message Step 1: read headers
-      rl_info, remaining, src = '', '', ''
+      pulls, period, remaining, src, auth = [''] * 5
       if ("ratelimit-limit" in res.headers):
-         rl_info = res.headers["ratelimit-limit"].split(";w=")
+         pulls, period = res.headers["ratelimit-limit"].split(";w=")
+         period = str(int(period) // 3600) # convert seconds to hours
       if ("ratelimit-remaining" in res.headers):
          remaining = res.headers["ratelimit-remaining"].split(";w=")[0]
       if ("docker-ratelimit-source" in res.headers):
          src = res.headers["docker-ratelimit-source"]
-      # Step 2: Construct friendly Docker Hub ratelimit message
-      rl_msg = [] # rate limit message list
-      if (src != ''):
          if (re.search(r"[0-9.a-f:]+", src).group(0) == src): # IP address
-            rl_msg.append("Docker Hub Auth: False")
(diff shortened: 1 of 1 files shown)


lib/charliecloud.py
```diff
@@ -2143,6 +2143,24 @@ def request_raw(self, method, url, statuses, auth=None, **kwargs):
                   % (method, statuses, res.status_code, res.reason))
       except requests.exceptions.RequestException as x:
          FATAL("%s failed: %s" % (method, x))
+      # construct friendly Docker Hub info message
+      rl_msg = [] # rate limit message array
+      if ("ratelimit-limit" in res.headers):
+         rl_info = res.headers["ratelimit-limit"].split(";w=")
+         rl_msg.append("Rate Limit: %s pulls per %s hours" % (rl_info[0],
+                                                              int(rl_info[1]) // 3600))
```
reidpr, line 2151: Add a comment about what units are being used.

lib/charliecloud.py
```diff
@@ -2143,6 +2143,24 @@ def request_raw(self, method, url, statuses, auth=None, **kwargs):
...
+      if ("ratelimit-remaining" in res.headers):
+         remaining = res.headers["ratelimit-remaining"].split(";w=")[0]
+         rl_msg.append("Pulls Remaining: %s" % remaining)
+      if ("docker-ratelimit-source" in res.headers):
+         src = res.headers["docker-ratelimit-source"]
+         if (re.match(r"[0-9.a-f:]+", src).group(0) == src): # IP address
```
reidpr, line 2157: 1. The idiomatic way to do this comparison would be just whether the match object is None or not.
2. By convention, we use `re.search()` for everything, with `^` and/or `$` if needed.

reidpr (changes requested):
Good progress. A few suggestions:

1. Gather the data and build up the message separately. Then it will be more clear what the log message is (e.g., it will be more searchable).
2. Use regular expressions to pull information out of the header values. Then it will be more clear what format is expected.
3. Remove the special case for headers containing “ratelimit” below the changes. Once the friendly message is present, we don't need to log rate limit headers at higher priority.

reidpr (commented):
Good progress. I added a tidy; take a look. It is untested (and in fact CI fails 😂), so will need more attention.

Issue #536, opened 2019-09-05
update tutorial, add best practices


j-ogas:
Our documentation is written and organized for a time that has since passed. We've come a long a way, our documentation is due for a makeover.

Updating the documentation is coupled with: #499, #484, and #445.

Brofessional:
Also related #414 and #348

reidpr:
So looking over all the issues tagged `doc`, there are a lot of specific critiques and then this one, which basically says “do a revision pass on the documentation to update it”. Given that, I'd expect a PR that addresses this issue along with several others selected from the `doc` tag. Maybe Hunter you can propose a set of issues to go together in that PR?

PR #1502, opened 2022-12-01
tutorial update for github

mphinney1100:
addresses #536

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -20,6 +20,7 @@
 
 EXTRA_DIST = \
 _loc.rst \
+best_practices.rst \
 bugs.rst \
 charliecloud.rst \
 ch-checkns.rst \
--- a/doc/best_practices.rst
+++ b/doc/best_practices.rst
@@ -1,4 +1,4 @@
-Best Practices
+Best practices
 **************
 
 Other best practices information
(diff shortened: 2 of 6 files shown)


reidpr (commented):
Great progress. I pushed a tidy commit. General comments:

1. I didn't worry too much about formatting, but there's a fair amount of stuff that still needs to be brought over from the Word doc.
2. In particular, some of the examples need to be updated to use the `$` vs. `>` prompt convention.
3. Lots of stuff likes to convert “--” (two hyphens) to “—”. This gotcha came up a few times.

reidpr (commented):
Nice! I pushed a tidy commit. Couple notes:

1. I wonder if if we should restore the notion of some enclosing directory that holds all the tutorial stuff (`$CHUTORIAL`?). If so, this needs another edit pass; if not, it needs another edit pass to remove a few lines I added with it.

2. Once we think it's ready for a dry run, let's have @lucaudill test it.

reidpr (changes requested):
I do want to keep the "MPI Hello World" section. I think the main thing that's needed is put back the deleted section, and fill in section "Build base image".

reidpr (approved):
Thanks! I committed a couple tidy commits. If they look good to you and tests pass, I'll merge.

PR #1504, opened 2022-12-05
don’t look for line continuations when matching `_NEWLINES`

### LINKED ISSUE #1503: parsing continuation lines with leading whitespace is slow
This simple Dockerfile takes about 10 seconds to parse (way too long):

```
$ cat /tmp/Dockerfile
FROM debian:11.5

RUN echo $(  echo hello \
                          | cat )  # each additional space ~doubles parse time
$ time ch-image build --parse-only /tmp
inferred image name: tmp

real	0m10.450s
user	0m10.422s
sys	0m0.027s
```

The culprit seems to be related to line continuations where the continued line starts with whitespace.

I *suspect* we are triggering Earley’s [O(*n*³) behavior](https://lark-parser.readthedocs.io/en/latest/parsers.html). Our grammar is not LALR and I don’t know how hard it would be to fix that.

reidpr:
Closes #1503.


lucaudill (approved):
looks good to me

mphinney1100 (approved):

PR #1505, opened 2022-12-07
Run images by name

### LINKED ISSUE #1058: `ch-run`: run images by name
When working with `ch-image` to pull, build, and list images, it seems a bit odd that to run the images with `ch-run`, one needs the full path to the image folder. This isn't simply `$CH_IMAGE_STORAGE/$IMAGE:$TAG` but `$CH_IMAGE_STORAGE/img/$IMAGE:$TAG`, which requires a bit of awareness of the structure of `ch-image`'s storage.

Here are two suggestions:

- Provide a `path` subcommand to `ch-image` that takes the image name (as in `list`) and provides the fully-qualified path to the image for use with `ch-run`; e.g.,

      ch-run $(ch-image path foo:1.2.3) bash

- Provide a `run` subcommand to `ch-image` that delegates to `ch-run`; e.g,

      ch-image [image_args...] run [run_args...] foo:1.2.3 bash`.

lucaudill:
closes #1058

Here's the tea: we've noticed that some users (a.k.a. me) try to run images out of the storage directory, e.g. 
```
$ ch-run $CH_IMAGE_STORAGE/img/foo -- echo foo
```
In general we don't want people to do this, so this PR is gonna make it so `ch-run` yells at you if you try it. Instead, we'll allow users to run images directly from storage by calling them by name, e.g. 
```
$ ch-run foo -- echo foo
```


bin/ch_misc.c
```diff
@@ -278,6 +278,63 @@ struct env_var env_var_parse(const char *line, const char *path, size_t lineno)
...
+  T_ (1 <= asprintf(&path, "%s/%s", storage, name));
+   if(!path_exists(path, NULL, false)) // make sure the path we constructed is there
+   {
+      if(path_exists(storage, NULL, false)){
+         FATAL("%s not found in storage", name);
+      } else { // Storage dir doesn't exist
+         FATAL("storage directory not found: %s", storage);
+      }   
+   } else if(writable) {
+      FATAL("'-w' not allowed when running out of storage");
```
reidpr, line 317: Allow this with `--unsafe`? Could simplify the test suite by taking advantage of this capability, I suspect.

bin/ch_misc.c
```diff
@@ -579,6 +636,17 @@ char *realpath_safe(const char *path)
    return pathc;
 }
 
+/* Replace all instances of character 'old' in array 'str' with 'new'. */
```
reidpr, line 639: Also document memory allocation behavior. As written, this implies that we modify `str`.

doc/ch-run.rst
```diff
@@ -92,6 +93,9 @@ setuid or setcap helpers, even for mounting SquashFS images with FUSE.
     bind-mounted into it. If this is specified, no such temporary files are
     created and the image’s files are exposed.
 
+  :code:`-s`, :code:`--storage DIR`
+    Set the storage directory (see section 6.5 for important details).
```
reidpr, line 97: The man pages don’t have section numbers; also this refers to a section in a different man page.

lib/charliecloud.py
```diff
@@ -458,7 +458,7 @@ def bytes_hash(data):
 def ch_run_modify(img, args, env, workdir="/", binds=[], fail_ok=False):
    # Note: If you update these arguments, update the ch-image(1) man page too.
```
reidpr, line 459: Did you update the `ch-image(1)` man page?

reidpr (changes requested):
Great work! Several comments in-line.

General: Please follow the surrounding code’s style w.r.t. indentation, braces, and things like that. There may also be some comments in the contributor’s guide.

follow surrounding style

test/run/ch-run_misc.bats
```diff
@@ -33,6 +65,13 @@ EOF
     ch-run -w "$ch_timg" rm write
 }
 
+@test 'run image by name' {
```
reidpr, line 68: I think “`specify storage`” can likely be eliminated in favor of expanding this test to validate all three methods of finding storage directory location. You can use the verbose output to check where `ch-run` thinks the storage directory is.

test/run/ch-run_misc.bats
```diff
@@ -33,6 +65,13 @@ EOF
...
+    [[ $CH_TEST_BUILDER = ch-image ]] || skip 'ch-image only'
+    run ch-run 00_tiny -- echo foo
```
reidpr, line 70: Instead of `echo foo` and checking the output, just use `true` and check if it succeeds.

reidpr (commented):
Great work! See tidy (untested). LMK what questions you have. Hope your holiday is going well.

reidpr (approved):
Great work. Please check my tidy. If it looks good and tests pass, I'll merge.

Issue #1147, opened 2021-08-10
check for storage directory version file when appropriate


reidpr:
Currently we check for subdirectories `img` and `dlcache` to see if a storage directory seems valid (`charliecloud.Storage.valid_p()`). Once we're confident #962 has been fixed long enough, also check for the version file and maybe some other things.

Issue #1007, opened 2021-03-16
remove last $CH_GROW_STORAGE check


reidpr:
Eventually the last person will stop using `$CH_GROW_STORAGE`, at which point we can remove the check in `charliecloud.py`. When? 🤷‍♂️

reidpr:
This can probably done now that we have a 1-year limit via #1507.

PR #1507, opened 2022-12-12
support storage directories up to 1 year old

reidpr:
Previously, we did not commit to any particular duration for support of old storage directories. The disadvantage is that we're growing an increasing amount of compatibility code and testing. This documents a 1-year support period.

One potential problem is distribution-packaged versions. Currently these tend to lag rather a lot (e.g. Debian stable has 0.21, which is about two years old and has storage version 1). I’m not *too* worried about that since everything in storage can be rebuilt, but I would like to gather comment on it. If we did want to support e.g. Debian stable-to-stable upgrades, we’re looking at 3 years or more of support.

If this passes review, I’ll also create child bugs to retire storage versions 1–4, and update the release instructions to create new such bugs upon release of each relevant version.

### DIFF
--- a/doc/ch-image.rst
+++ b/doc/ch-image.rst
@@ -264,7 +264,7 @@ The storage directory format changes on no particular schedule.
 :code:`ch-image` is normally able to upgrade directories produced by a given
 Charliecloud version up to one year after that version's release. Upgrades
 outside this window and downgrades are not supported. In these cases,
-:code:`ch-image` will refuse to fun until you delete and re-initialize the
+:code:`ch-image` will refuse to run until you delete and re-initialize the
 storage directory with :code:`ch-image reset`.
 
 .. warning::
--- a/doc/ch-image.rst
+++ b/doc/ch-image.rst
@@ -225,8 +225,8 @@ Storage directory
 =================
 
 :code:`ch-image` maintains state using normal files and directories located in
-its *storage directory*; contents include temporary images used for building
-and various caches.
+its *storage directory*; contents include various caches and temporary images
+used for building.
 
 In descending order of priority, this directory is located at:
 
@@ -260,11 +260,12 @@ images runnable with :code:`ch-run`, this is not a supported use case. The
 supported workflow uses :code:`ch-convert` to obtain a packed image; see the
 tutorial for details.
 
-The storage directory format changes on no particular schedule. Often
-:code:`ch-image` is able to upgrade the directory; however, downgrading is not
-supported and sometimes upgrade is not possible. In these cases,
-:code:`ch-image` will refuse to run until you delete and re-initialize the
-directory with :code:`ch-image reset`.
+The storage directory format changes on no particular schedule.
+:code:`ch-image` is normally able to upgrade directories produced by a given
+Charliecloud version up to one year after that version's release. Upgrades
+outside this window and downgrades are not supported. In these cases,
+:code:`ch-image` will refuse to fun until you delete and re-initialize the
+storage directory with :code:`ch-image reset`.
 
 .. warning::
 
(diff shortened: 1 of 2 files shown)


reidpr:
@olifre @wiene, any opinions?

doc/ch-image.rst
```diff
@@ -260,11 +260,12 @@ images runnable with :code:`ch-run`, this is not a supported use case. The
...
-The storage directory format changes on no particular schedule. Often
-:code:`ch-image` is able to upgrade the directory; however, downgrading is not
-supported and sometimes upgrade is not possible. In these cases,
-:code:`ch-image` will refuse to run until you delete and re-initialize the
-directory with :code:`ch-image reset`.
+The storage directory format changes on no particular schedule.
+:code:`ch-image` is normally able to upgrade directories produced by a given
+Charliecloud version up to one year after that version's release. Upgrades
+outside this window and downgrades are not supported. In these cases,
+:code:`ch-image` will refuse to fun until you delete and re-initialize the
```
j-ogas, line 267:
```suggestion
:code:`ch-image` will refuse to run until you delete and re-initialize the
```

j-ogas (changes requested):
Thoughts about supporting the most recent X storage versions, regardless of time lapsed? For example we support storage version n, n-1, n - 2, etc., where n is the current storage version upstream?

olifre:
@reidpr Sorry for the delayed reply. I think for Gentoo the situation is easy, as I will always try to keep only approximately the last 2-3 releases in-tree, and Gentoo users are regularly updating due to it's rolling nature (for example, there are always two versions of KDE packages in-tree, a stable one and the next one for testing). 

For other distributions, I wonder if this is something they may want to handle in the upgrade path: If the packager knows that the Charliecloud version packaged in the old stable and the current stable version of the distribution are too far apart to support upgrading the storage format, maybe the packaging code can already clear the storage directories or advice the user during the system upgrade with a message to do so. 
But I think @wiene as the one packaging the Debian version may have a more elaborate opinion on this.

reidpr:
See also #1007 and #1147.

wiene:
I suppose that supporting an outdated storage directory format means the ability to upgrade to the most recent version. 

As you write, @reidpr, I guess that in many (or even all) cases users should be able to rebuild the contents of the storage directory without too much effort. Thus we are talking about a convenience feature here. And [the documentation](https://hpc.github.io/charliecloud/ch-image.html#storage-directory) even mentions that "sometimes upgrade is not possible".

On the other hand convenience is also a nice feature. ;-) Therefore one has to strike a balance between the technical burden to maintain the required compatibility code and the user convenience it brings about. I do not feel in a position to make any judgement concerning this point.

In order to offer a convenient upgrade path for users upgrading from one stable Debian release to the next one (anything else is not officially supported), a support period of three years would be fine. The [time between the start of freezes for subsequent releases is typically two years](https://wiki.debian.org/DebianReleases). To be on the safe side we should add the typical time between Charliecloud releases plus the delay caused by packaging the release for Debian and some extra safety margin to account for fluctuations, such that we end up with something like three years.

If you think the price for a three year support period is too high, that would also be fine in my opinion as long as Charliecloud issues a clear error/warning message instructing the user how to proceed.

doc/ch-image.rst
reidpr: aw man, you’re no fun

reidpr:
> Thoughts about supporting the most recent X storage versions regardless of time lapsed? For example we support storage version n, n-1, n - 2, etc., where n is the current storage version upstream?

Makes sense. However I’m skeptical because (1) storage version bumps really are rather erratic, e.g. we had something like 3–4 releases for storage v2 and then only one release for storage v4 and (2) I don’t want to expose the storage version to users.

We could instead say something like “*m* major versions”, which is easier to count but also less predictable, and I keep wanting to increase the release cadence, so maybe that will actually happen sometime.

j-ogas (approved):

Issue #1509, opened 2022-12-13
update code for ShellCheck 0.9.0


lucaudill:
ShellCheck [v0.9.0](https://github.com/koalaman/shellcheck/blob/v0.9.0/CHANGELOG.md) recently came out, and with it some new warnings that affect the test suite. In particular, [SC2317](https://www.shellcheck.net/wiki/SC2317) pops up everywhere now. As a temporary fix I'm adding it to the ignore list, but at some point we should go through each instance and fix it. 

(Warning: when I say SC2317 pops up everywhere, I mean EVERYWHERE.)

PR #1510, opened 2022-12-13
make tests pass under ShellCheck 0.9.0

lucaudill:
Makes the test suite pass. #1509 suggests a more permanent solution.


reidpr (approved):
Great work. I don't think this is kludgy at all. ShellCheck also passes under 0.8.0.

PR #1511, opened 2022-12-14
newer alpine

### LINKED ISSUE #1483: update to newer Alpine
Throughout the test suite and examples we use Alpine Linux version 3.9. As of this writing, this still works, but the oldest [supported image](https://hub.docker.com/_/alpine) is 3.13 and the newest is 3.16. We should probably update.

Note that there are different versions used here and there, and this is for a reason; we don't want to just update everything to 3.16.

Along these lines, the purpose of the `00_tiny` image was largely to prevent explicit `FROM alpine:3.9`, but that seems to have not have happened. It's still all over the place. Further, in some situations this is required, e.g. in the build cache tests we reset things over and over, so re-building a base image would be needlessly tedious. It also tends to cause annoyance when running individual tests that use it. Should we remove `00_tiny`? I see the image has grown one test fixture, but that could probably be put somewhere else.

lucaudill:
closes #1483


test/Dockerfile.00_tiny
```diff
@@ -2,7 +2,7 @@
 
 # ch-test-scope: quick
 
-FROM alpine:3.9
```
reidpr, line 5: I thought `00_tiny` was going away?

test/old-storage
```diff
@@ -33,7 +33,7 @@ certain properties:
        $ tar tf storage-0.26.x86_64.tar.gz | head -3
        ./
        ./dlcache/
-       ./dlcache/alpine:3.9.fat.json
+       ./dlcache/alpine:3.17.fat.json
```
reidpr, line 36: This actually will not change because it's unpacking an old tarball.

reidpr (changes requested):
Great, thanks. Couple comments in-line.

The one larger issue: I'd like to get rid of `Dockerfile.00_tiny` entirely. So please keep going a little bit on that part.

test/run/ch-convert.bats
```diff
@@ -326,49 +326,59 @@ test_from () {
...
     # ch-image
-    printf 'FROM alpine:3.9\n' | ch-image build -t tmpimg -f - "$BATS_TMPDIR"
+    printf 'FROM alpine:3.17\n' | ch-image build -t tmpimg -f - "$BATS_TMPDIR"
     run ch-convert --no-clobber -o ch-image "$BATS_TMPDIR" tmpimg
     echo "$output"
     [[ $status -eq 1 ]]
     [[ $output = *"error: exists in ch-image storage, not deleting per --no-clobber: tmpimg" ]]
 
+    # convert ch_timg into ch-image format
+    ch-image delete timg | 'true'
```
reidpr, line 336: lol this actually works but it's wrong. You want:

```suggestion
    ch-image delete timg || true
```

test/run/ch-convert.bats
```diff
@@ -326,49 +326,59 @@ test_from () {
...
+    if [[ $(file "$ch_timg") = *"symbolic link"* ]]; then
```
reidpr, line 337: I'd use `stat(3)` here instead. There are output format options to get just what you want. `file(3)` is too broad.

test/run/ch-convert.bats
```diff
@@ -326,49 +326,59 @@ test_from () {
...
+        # symlink to squash archive
+        ch-convert -i squash -o ch-image "$ch_timg" timg
+    else
+        # directory
+        ch-convert -i dir -o ch-image "$ch_timg" timg
```
reidpr, line 342: Not DRY — set a variable for the input format and have the `ch-convert` command just once.

test/run/ch-run_misc.bats
```diff
@@ -32,7 +32,6 @@ EOF
     ch-run -w "$ch_timg" rm write
 }
 
-
```
reidpr, line 35: You'll want to leave this whitespace in place.

test/build/50_dockerfile.bats
```diff
@@ -1218,8 +1218,8 @@ EOF
     [[ $CH_TEST_BUILDER == ch-image ]] || skip 'ch-image only'
 
     run ch-image build --no-cache -t foo - <<'EOF'
-FROM alpine:3.9
-FROM alpine:3.10
+FROM alpine:3.17
+FROM alpine:3.17
```
reidpr, line 1222: I don't remember why these two layers had different base images but we should probably keep them different.

reidpr (changes requested):
Looks great. Several minor nits to fix and then I'll merge.

test/run/ch-convert.bats
```diff
@@ -333,14 +333,15 @@ test_from () {
     [[ $output = *"error: exists in ch-image storage, not deleting per --no-clobber: tmpimg" ]]
 
     # convert ch_timg into ch-image format
-    ch-image delete timg | 'true'
-    if [[ $(file "$ch_timg") = *"symbolic link"* ]]; then
+    ch-image delete timg | true
+    if [[ $(stat "$ch_timg") = *"symbolic link"* ]]; then
```
reidpr, line 337: This can be tightened up with something like:

```suggestion
    if [[ $(stat -c %F "$ch_timg") = 'symbolic link' ]]; then
```

reidpr (changes requested):
Looks good; one more suggestion for a tighter test.

reidpr (approved):

PR #1514, opened 2022-12-20
ch-image build: support LABEL

### LINKED ISSUE #781: ch-image build: support LABEL

BenStormer:
Closes #781

### DIFF
--- a/doc/ch-image.rst
+++ b/doc/ch-image.rst
@@ -755,7 +755,9 @@ serves the same purpose as the :code:`ARG` instruction. It can be repeated.
 ~~~~~~~~~~~~~
 
 The :code:`LABEL` instruction accepts :code:`key=value` pairs to
-add metadata for an image. Unlike Docker, multiline values are not supported. Can be repeated.
+add metadata for an image. Unlike Docker, multiline values are not supported;
+see issue `#1512 <https://github.com/hpc/charliecloud/issues/1512>`_.
+Can be repeated.
 
 :code:`COPY`
 ~~~~~~~~~~~~
--- a/doc/ch-image.rst
+++ b/doc/ch-image.rst
@@ -751,7 +751,7 @@ The :code:`FROM` instruction accepts option :code:`--arg=NAME=VALUE`, which
 serves the same purpose as the :code:`ARG` instruction. It can be repeated.
 
 :code:`LABEL`
-~~~~~~~~~~~~
+~~~~~~~~~~~~~
 
 The :code:`LABEL` instruction accepts :code:`key=value` pairs to
 add metadata for an image. Unlike Docker, this metadata cannot be observed by
--- a/doc/ch-image.rst
+++ b/doc/ch-image.rst
@@ -750,6 +750,16 @@ in other :code:`ARG` before the first :code:`FROM`.
 The :code:`FROM` instruction accepts option :code:`--arg=NAME=VALUE`, which
 serves the same purpose as the :code:`ARG` instruction. It can be repeated.
 
+:code:`LABEL`
+~~~~~~~~~~~~
+
+The :code:`LABEL` instruction accepts :code:`key=value` pairs to
+add metadata for an image. Unlike Docker, this metadata cannot be observed by
+using a :code:`inspect` command since Charliecloud users can directly view the
+metadata within their image manifest. Currently, :code:`LABEL` does not allow
+for users to declare label values on multiple lines. See issue `#1512
+<https://github.com/hpc/charliecloud/issues/1512>`_.
+
 :code:`COPY`
 ~~~~~~~~~~~~
 
--- a/doc/ch-image.rst
+++ b/doc/ch-image.rst
@@ -754,11 +754,7 @@ serves the same purpose as the :code:`ARG` instruction. It can be repeated.
 ~~~~~~~~~~~~~
 
 The :code:`LABEL` instruction accepts :code:`key=value` pairs to
-add metadata for an image. Unlike Docker, this metadata cannot be observed by
-using a :code:`inspect` command since Charliecloud users can directly view the
-metadata within their image manifest. Currently, :code:`LABEL` does not allow
-for users to declare label values on multiple lines. See issue `#1512
-<https://github.com/hpc/charliecloud/issues/1512>`_.
+add metadata for an image. Unlike Docker, multiline values are not supported. Can be repeated.
 
 :code:`COPY`
 ~~~~~~~~~~~~
--- a/lib/build.py
+++ b/lib/build.py
@@ -322,24 +322,6 @@ class Instruction(abc.ABC):
          return self.image.metadata["env"]
 
    @property
-   def label_arg(self):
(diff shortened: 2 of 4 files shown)


doc/ch-image.rst
```diff
@@ -750,6 +750,16 @@ in other :code:`ARG` before the first :code:`FROM`.
 The :code:`FROM` instruction accepts option :code:`--arg=NAME=VALUE`, which
 serves the same purpose as the :code:`ARG` instruction. It can be repeated.
 
+:code:`LABEL`
+~~~~~~~~~~~~
```
j-ogas, line 754:
```suggestion
~~~~~~~~~~~~~
```

doc/ch-image.rst
```diff
@@ -750,6 +750,16 @@ in other :code:`ARG` before the first :code:`FROM`.
...
+
+The :code:`LABEL` instruction accepts :code:`key=value` pairs to
+add metadata for an image. Unlike Docker, this metadata cannot be observed by
+using a :code:`inspect` command since Charliecloud users can directly view the
+metadata within their image manifest. Currently, :code:`LABEL` does not allow
+for users to declare label values on multiple lines. See issue `#1512
+<https://github.com/hpc/charliecloud/issues/1512>`_.
```
j-ogas, line 761:
```suggestion
The :code:`LABEL` instruction accepts :code:`key=value` pairs to
add metadata for an image. Unlike Docker, multiline values are not supported. Can be repeated.
```

j-ogas: We should probably move the multiline issue to a comment somewhere, perhaps in the test file?

j-ogas (changes requested):
Nice, thanks! Initial feedback and requested changes focus on documentation.

We should discuss the parse test offline with @reidpr.

reidpr (commented):
Looks like you got good feedback from Jordan, so I’m going to hold off for the time being. LMK if that’s wrong.

doc/ch-image.rst
BenStormer: That sounds good to me, I'll move it there in a place that looks good.

reidpr (changes requested):
Looks like there's some CI failures — as Jordan noted, take care of those and then re-request review.

lib/build.py
```diff
@@ -450,6 +468,10 @@ def prepare(self, miss_ct):
          Typically, subclasses will set up enough state for self.sid_input to
          be valid, then call super().prepare().
 
+         WARNING: Instructions that modify image metadata (at this writing,
+         ARG ENV FROM LABEL SHELL WORKDIR) must do so here, not in execute(),
+         so that metadata is available to late instructions even on cache hit.
+
```
reidpr, line 474: Looks like there's a merge error here: This paragraph moved to item 3 below, so modify that instead.

lib/build.py
```diff
@@ -976,6 +999,53 @@ def __init__(self, *args):
...
+   @property
+   def str_(self):
+      return "%s='%s'" % (self.key, self.value)
+
+   def execute(self):
+      with (self.image.unpack_path // "/ch/metadata.json").open_("wt") \
+           as fp:
+         for (k, v) in self.label_label.items():
+            print("%s=%s" % (k, v), file=fp)
```
reidpr, line 1019: This class doesn't need `execute()`; see point 3 in the comment for `prepare()` above.

lib/build.py
```diff
@@ -321,6 +321,24 @@ def env_env(self):
...
+   def label_build(self):
+      return { **self.label_arg, **self.label_label }
+
+   @property
+   def label_label(self):
+      if (self.image is None):
+         assert False, "unimplemented" # return dict()
+      else:
+         return self.image.metadata["labels"]
+
```
reidpr, line 341: I think these are not needed, for two reasons: (1) there is no distinction between “labels to save in the container’s environment”, akin to `ENV`, and “labels to use in the build environment”, i.e., the union of `ARG` and `ENV`, and (2) because we don't actually *use* the labels for anything, I don't think we need any property for access within Charliecloud.

lib/build.py
```diff
@@ -976,6 +999,53 @@ def __init__(self, *args):
...
+   def execute(self):
+      with (self.image.unpack_path // "/ch/metadata.json").open_("wt") \
+           as fp:
+         for (k, v) in self.label_label.items():
+            print("%s=%s" % (k, v), file=fp)
+
+   def prepare(self, *args):
+      self.value = ch.variables_sub(unescape(self.value), self.label_build)
+      self.label_label[self.key] = self.value
+      return super().prepare(*args)
```
reidpr, line 1024: This seems muddled. I'd expect it to set the image metadata, which I think it does but it's very roundabout?

`self.label_build` should I think be `self.env_build` if that clarifies things.

reidpr (changes requested):
Good progress! Some comments in-line. Looking forward to the next iteration.

j-ogas (dismissed):
See @reidpr's feedback. Thanks!

lib/build.py
BenStormer: Good catch, thanks!

lib/build.py
BenStormer: Thanks, another good catch

lib/build.py
BenStormer: Ok, I think I got some changes figured out and it appears to pass CI. My only question is regarding env_build. I assume it is ok to use this variable even though it is a reference to an env property in the label class? It works, but may be a little indirect. I tried just removing that line of code and this line just helps the grammar interpret the value correctly, so maybe there is another way to do it. If you are happy with the current status though, I'm happy with it too.

doc/ch-image.rst
```diff
@@ -751,6 +751,12 @@ in other :code:`ARG` before the first :code:`FROM`.
 The :code:`FROM` instruction accepts option :code:`--arg=NAME=VALUE`, which
 serves the same purpose as the :code:`ARG` instruction. It can be repeated.
 
+:code:`LABEL`
+~~~~~~~~~~~~~
+
+The :code:`LABEL` instruction accepts :code:`key=value` pairs to
+add metadata for an image. Unlike Docker, multiline values are not supported. Can be repeated.
```
reidpr, line 758: 1. Add a link to the issue to add multiline support (``(`issue #xxxx <https:/...>`_)``).
2. Wrap paragraph to 80 characters.

lib/build.py
reidpr: Yes, you can reference `env_build`. It just means "environment to use during the build". Lots of instructions need that.

reidpr (changes requested):
Looks great Ben! One documentation fix and then if Jordan is happy I think it's good to merge.

lib/build.py
BenStormer: Great! Thanks

doc/ch-image.rst
BenStormer: Ok, should be all good now!

reidpr (approved):

PR #1517, opened 2022-12-22
refactor fatal error handling

### LINKED ISSUE #1486: `COPY` errors give an error message printed before the instruction
Many types of `COPY` errors result in an error message before the instruction is announced; e.g.:

```
$ printf 'FROM alpine:3.9\nCOPY /doesnotexist .\n' | ch-image build -t foo -f -
/tmp
  1* FROM alpine:3.9
error: can't copy: source file not found: /doesnotexist
```

which is misleading because it associates the error with the previous, probably non-`COPY` instruction.

This is complicated by the fact that we need to do a lot of the `COPY` before we know whether it's a hit or miss, which is part of the announcement.

reidpr:
closes #1486

### DIFF
--- a/bin/ch-image.py.in
+++ b/bin/ch-image.py.in
@@ -306,4 +306,8 @@ def main():
 ## Bootstrap ##
 
 if (__name__ == "__main__"):
-   main()
+   try:
+      main()
+   except ch.Fatal_Error as x:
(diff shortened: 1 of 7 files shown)


lucaudill (approved):
LGTM!

PR #1530, opened 2023-01-04
only pull skinny manifest once

### LINKED ISSUE #1101: ch-image: use the skinny manifest we got when requesting a fat one
When requesting a fat manifest, if there is not one, a skinny manifest is returned. However in this case, we immediately re-request the skinny one. We should instead use the skinny manifest we already got, possibly provided it's a v2 manifest.

lucaudill:
closes #1101

### DIFF
--- a/lib/pull.py
+++ b/lib/pull.py
@@ -91,6 +91,7 @@ class Image_Puller:
       "Download image metadata and layers and put them in the download cache."
       # Spec: https://docs.docker.com/registry/spec/manifest-v2-2/
       ch.VERBOSE("downloading image: %s" % self.image)
+      have_skinny = False
       try:
          # fat manifest
          if (ch.arch != "yolo"):
@@ -101,6 +102,9 @@ class Image_Puller:
                            ("available: %s"
                             % " ".join(sorted(self.architectures.keys()))))
             except ch.No_Fatman_Error:
+               # currently, this error is only raised if we've downloaded the skinny
+               # manifest.
+               have_skinny = True
                if (ch.arch == "amd64"):
                   # We're guessing that enough arch-unaware images are amd64 to
                   # barge ahead if requested architecture is amd64.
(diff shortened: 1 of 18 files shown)


lib/pull.py
```diff
@@ -173,9 +173,11 @@ def fatman_load(self):
                                    "manifest list: downloading")
       fm = self.fatman_path.json_from_file("fat manifest")
       if ("layers" in fm or "fsLayers" in fm):
-         # FIXME (issue #1101): If it's a v2 manifest we could use it instead
-         # of re-requesting later. Maybe we could here move/copy it over to
-         # the skinny manifest path.
+         if (not os.path.exists(str(self.manifest_path))):
+            # check for broken symlink
+            if (os.path.lexists(str(self.manifest_path))):
+               os.remove(str(self.manifest_path))
```
reidpr, line 179: 1. You don't actually have to worry about the dangling symlink case. Just use `lexists()` one time.
2. That said, is there a reason this isn't using the `Path` object methods?
3. Add a comment explaining this symlink trick.

test/build/55_cache.bats
```diff
@@ -363,7 +363,6 @@ EOF
     [[ $status -eq 0 ]]
     [[ $output = *'. FROM'* ]]
     [[ $output = *'manifest list: downloading'* ]]
-    [[ $output = *'manifest: downloading'* ]]
```
reidpr, line 366: This message should still be printed for arch-aware images.

reidpr (changes requested):
Nice work. See comments in-line.

lib/pull.py
lucaudill: 1. Would you be willing to say more about this? My testing made it seem like if there is a dangling symlink present you'll get a FileExistsError. I wanted to protect against that case, even if it's unlikely.

2. No particularly strong reason, I'll consider switching.

reidpr (approved):
LGTM, thanks! I'll merge once we get 0.31 out the door.

PR #1534, opened 2023-01-11
ch-test: update default image and pack defaults

### LINKED ISSUE #1218: Change default test location for ch-test
Currently, the default location for tests with ch-test is: `/var/tmp/img` for images and `/var/tmp/tar/` for tarballs. In any multi user system, it's possible for this to result in the directories clashing between multiple charliecloud users. Instead, I propose we make the default `/var/tmp/<username>/` to avoid that issue.

j-ogas:
Closes #1218.

### DIFF
--- a/bin/ch-test
+++ b/bin/ch-test
@@ -968,6 +968,12 @@ if [[ -z $phase ]]; then
 fi
 printf '\n'
 
+# FIXME: https://github.com/hpc/charliecloud/issues/1580
+echo  'NOTE: default image and pack directories changed to:'
+echo  '      CH_TEST_IMGDIR=/var/tmp/${USER}.img'
+echo  '      CH_TEST_TARDIR=/var/tmp/${USER}.pack'
+printf '\n'
+
 #    variable name    CLI              environment         default
 #                     desc. width  description
 vset CH_TEST_SCOPE    "$scope"         "$CH_TEST_SCOPE"    standard \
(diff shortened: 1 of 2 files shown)


reidpr (commented):
This looks good to me, but maybe we should offer some kind of warning for the old path, because those directories can be pretty large?

bin/ch-test
```diff
@@ -935,6 +935,13 @@ if [[ -z $phase ]]; then
 fi
 printf '\n'
 
+# FIXME: https://github.com/hpc/charliecloud/issues/1580
+# shellcheck disable=SC2016
+echo  'NOTE: default image and pack directories changed to:'
+echo  '      CH_TEST_IMGDIR=/var/tmp/${USER}.img'
+echo  '      CH_TEST_TARDIR=/var/tmp/${USER}.pack'
+printf '\n'
```
reidpr, line 943: Might be good to print this (1) to stderr and (2) only if the old defaults actually exist.

bin/ch-test
```diff
@@ -935,6 +935,13 @@ if [[ -z $phase ]]; then
 fi
 printf '\n'
 
+# FIXME: https://github.com/hpc/charliecloud/issues/1580
```
reidpr, line 938:
```suggestion
# See issue #1580
```

reidpr (changes requested):
Good progress. See notes in-line; also the tests are not passing.

PR #1536, opened 2023-01-11
get rid of `--no-home`

### LINKED ISSUE #1524: don’t accept `--no-home` any more
Now that this is the default, no need for the option.

lucaudill:
closes #1524


reidpr (approved):
Looks great to me, thanks. I'll merge once we release 0.31.

Issue #1379, opened 2022-06-20
`ch-ssh`: remove


j-ogas:
In 5 years I've never once used this; thus, I'd like to see it purged to help reduce clutter.

shanegoff:
#2 and #4 related to this

reidpr:
See discussion #1600. We drew attention to this in release notes and some direct solicitations. We got one vote and it was for “no, I don’t use this”, so there’s our answer I think.

PR #1537, opened 2023-01-12
remove ch-ssh

j-ogas:
Closes: #1379.
Moots: #2 and #4.

In case we do decide to trim out `ch-ssh`.


reidpr:
See also discussion #1600.

reidpr:
Looks like the discussion generated exactly one vote and it was for “no, I don’t use this”. So lets remove it.

doc/tutorial.rst
```diff
@@ -1031,46 +1032,9 @@ container. The simplest is to manually invoke :code:`ch-run` in the
...
-      otherwise you’ll get a shell that accepts commands but doesn’t print
-      prompts, among other other issues. (`Issue #2
-      <https://github.com/hpc/charliecloud/issues/2>`_.)
-
-A third approach may be to edit one’s shell initialization scripts to check
-the command line and :code:`exec(1)` :code:`ch-run` if appropriate. This is
-brittle but avoids wrapping :code:`ssh` or altering its command line.
+Another may be to edit one's shell initialization scripts to check the command
+line and :code:`exec(1)` :code:`ch-run` if appropriate. This is brittle but
+avoids wrapping :code:`ssh` or altering its command line.
```
j-ogas, line 1037: Would a comment, here or in FAQ, that points to a previous tag/commit ch-ssh file  as an example be helpful?

Something like: "Another may be to implement a ssh wrapper, see since removed (link) as an example."?

PR #1538, opened 2023-01-15
configure: Remove bashisms, fix dash compatibility.

olifre:
`+=` and `|&` are not supported by systems using dash as shell, replace them by more clumsy but more portable syntax. 

See https://bugs.gentoo.org/890873 for an example report from a system using `dash` as `/bin/sh`.


reidpr:
Thank you @olifre! Sorry to miss this earlier.

olifre:
Many thanks for the quick review and merge, @reidpr ! :smile:

PR #1550, opened 2023-01-28
fix `realpath(3)` error handling

olifre:
glibc's realpath may set errno also on success. Since POSIX specifies errno should only be checked after realpath if the return value is not NULL, this should be checked first. 

See for example:
 https://bugzilla.redhat.com/show_bug.cgi?id=1916968

An example is the following program:
```
#include <errno.h>
#include <stdio.h>
#include <stdlib.h>

int main() {
  char *pathc = realpath("/home/", NULL);
  printf("%s\n", pathc);
  printf("%d\n", errno);
  return 0;
}
```
On my Gentoo system using `glibc-2.36`, it echoes upon executing:
```
/home
22
```

### DIFF
--- a/bin/ch_misc.c
+++ b/bin/ch_misc.c
@@ -589,7 +589,7 @@ char *realpath_(const char *path, bool fail_ok)
    errno = 0;
    pathc = realpath(path, NULL);
 
-   if (errno != 0) {
+   if (pathc == NULL && errno != 0) {
       if (fail_ok) {
          T_ (pathc = strdup(path));
       } else {
--- a/bin/ch_misc.c
+++ b/bin/ch_misc.c
@@ -586,10 +586,9 @@ char *realpath_(const char *path, bool fail_ok)
    if (path == NULL)
       return NULL;
 
-   errno = 0;
    pathc = realpath(path, NULL);
 
-   if (pathc == NULL && errno != 0) {
+   if (pathc == NULL) {
       if (fail_ok) {
          T_ (pathc = strdup(path));
       } else {


reidpr:
Ugh, sorry about that. I just changed it *on purpose* because I thought I'd read the man page more carefully and now it turns out I read it *less* carefully. 😂

I want to remove the `errno` setting too. I can probably produce an updated patch today.

reidpr:
@olifre Do you think this is survivable or does 0.32 need to come out ASAP? It did pass all our testing, including on an internal system where `/tmp` and `/var/tmp` are symlinks.

olifre:
I also had to read the man page thrice, I did first believe my system was broken ;-). 

> @olifre Do you think this is survivable or does 0.32 need to come out ASAP? 

That's a good question. I know that Gentoo is affected with any stable `glibc` flavour there, but I have patched it if people use `charliecloud` via the package manager. 
I also learnt from @wiene that Debian testing / Bookworm is affected. Testing on Bullseye, I do not see this issue. 

So the impact depends on the bleeding-edge-ness of users' systems. I believe most systems Charliecloud is used on are not bleeding edge (especially HPCs), so I am not sure how large the real-life impact will be.

reidpr:
@olifre What do you think about my further commit? If it looks OK to you, passes tests, and gets internal approval too, I’ll merge, maybe tomorrow.

olifre:
@reidpr That looks good, indeed, I think we don't need `errno` at all, but the return value should be sufficient to check for an error condition if I understand the man page correctly. :+1:

reidpr (approved):

PR #1559, opened 2023-02-14
add phase to `ch-test` that builds images

lucaudill:


bin/ch-test
```diff
@@ -81,6 +81,14 @@ EOF
 
 ### Functions
 
+build_images () {
```
reidpr, line 84: This should be called `test_build_images` I think.

bin/ch-test
```diff
@@ -466,7 +471,7 @@ test_make_auto () {
             exit 0
         fi
 
-        if [[ $mode == build ]]; then
+        if [[ $mode == build || $mode == build-images ]]; then
```
reidpr, line 474: I think you can do something like:

```suggestion
        if [[ $mode == build* ]]; then
```

if you want.

bin/ch-test
```diff
@@ -788,7 +793,7 @@ fi
 while [[ $# -gt 0 ]]; do
     opt=$1; shift
     case $opt in
-    all|build|clean|examples|mk-perm-dirs|rm-perm-dirs|run)
+    all|build|clean|examples|mk-perm-dirs|build-images|rm-perm-dirs|run)
```
reidpr, line 796:
```suggestion
    all|build|build-images|clean|examples|mk-perm-dirs|rm-perm-dirs|run)
```

bin/ch-test
```diff
@@ -55,7 +55,7 @@ Usage:
 
   $ $(basename "$0") [PHASE] [--scope SCOPE] [--pack-fmt FMT] [OPTIONS]
 
-Valid phases: build, run, examples, all, mk-perm-dirs, clean, rm-perm-dirs
+Valid phases: build, build-images, run, examples, all, mk-perm-dirs, clean, rm-perm-dirs
```
reidpr, line 58: Is this longer than 80 characters now?

reidpr (changes requested):
Nice tidy little change. Some minor things to fix in-line.

reidpr (approved):

PR #1575, opened 2023-02-24
use root:root when appending to Docker tarballs

### LINKED ISSUE #1573: can’t `ch-convert` from Docker if UID > 2²¹
Hello,
we are quite large university using Charliecloud for a while in our HPC center. Every person on the university (staff, student) receives unique random 10-digit number. Recently we have started using this unique number as UID for any unix user (we have used some other mapping previously). And here comes the problem.
There is a function `dockman_to_tar` in the `ch-convert` script. In this function we call `docker export` command, which unfortunately exports the docker image in old tar format with limitations for UID<2^21. Following `tar rf` command then fails with error:
> tar: value 88889462 out of uid_t range 0..2097151
> tar: Exiting with failure status due to previous errors

We have made some changes in the script, although it is not the best solution, it takes some time. Moreover, we should test UID<2^21 etc. It extracts the old tar from docker to a temporary directory and then recreate it using tar command. The following diff is valid for version 0.31.
[ch-convert-diff.txt](https://github.com/hpc/charliecloud/files/10825786/ch-convert-diff.txt)
Best regards,
Jakub

reidpr:
Closes #1573.

### DIFF
--- a/bin/ch-convert
+++ b/bin/ch-convert
@@ -377,7 +377,10 @@ dockman_to_tar () {
     INFO 'adding environment ...'
     "$1" inspect "$2" \
             --format='{{range .Config.Env}}{{println .}}{{end}}' > "$tmpenv"
-    tar rf "$tmptar" -b1 -P --xform="s|${tmpenv}|ch/environment|" "$tmpenv"
+    # The tar flavor Docker gives us does not support UIDs or GIDs greater
+    # than 2**21, so use 0/0 rather than what’s on the filesystem. See #1573.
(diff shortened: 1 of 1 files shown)


reidpr:
@yaghobvtm, can you please test this PR and see if it addresses your problem? It should be faster than the patch you sent.

yaghobvtm:
Yes, I can confirm, the solution is working for users with UID>2^21.

mphinney1100 (approved):
LGTM

Issue #1571, opened 2023-02-21
multistage build: cannot reference alias from previous stage(s)


j-ogas:
Reproducer:
```docker
FROM alpine as a
RUN echo foo

FROM a as b
RUN echo bar

FROM b
RUN echo qux
```

Expected behavior: image builds without issue.

Actual behavior:
```
$ ch-image build -t multi .
  1. FROM alpine AS a
[...]
  4. FROM a AS b
error: unauthorized or not in registry: registry-1.docker.io:443/library/a:latest
hint: if your registry needs authentication, use --auth
```

It appears to fail for one or more of the following reasons: (1) the tag has `:latest` appended by default, (2) ch-image does store images in storage with their alias, i.e., there is no `a`, `b`, or `c` in storage; rather there is `multi_stage0`, etc., and (3) ch-image doesn't check if the FROM argument matches an previous stage image alias.

reidpr:
Closed by PR #1581.

PR #1581, opened 2023-02-28
`ch-image build`: fix multistage alias referencing.

j-ogas:
Addresses #1571.

### DIFF
--- a/lib/build.py
+++ b/lib/build.py
@@ -1105,12 +1105,13 @@ class I_from_(Instruction):
       if (str(self.image.ref) == str(self.base_image.ref)):
          ch.FATAL("output image ref same as FROM: %s" % self.base_image.ref)
       # Close previous stage if needed.
-      if (self.image_i > 0 and hasattr("self", "parent")):
+      ch.WARNING("image_unpack: %s" % self.image.unpack_path)
+      if (self.image_i > 0):
          # We need to check out the previous stage (a) to read its metadata
          # and (b) in case there's a COPY later. This will still be fast most
          # of the time since the correct branch is likely to be checked out
          # already.
-         self.parent.checkout()
+         self.parent.checkout(self.parent.base_image)
          self.parent.ready()
       # At this point any meaningful parent of FROM, e.g., previous stage, has
       # been closed; thus, act as own parent.
--- a/lib/build.py
+++ b/lib/build.py
@@ -1105,11 +1105,11 @@ class I_from_(Instruction):
       if (str(self.image.ref) == str(self.base_image.ref)):
          ch.FATAL("output image ref same as FROM: %s" % self.base_image.ref)
       # Close previous stage if needed.
-      if (miss_ct == 0 and self.image_i > 0):
-         # While there haven't been any misses so far, we do need to check out
-         # the previous stage (a) to read its metadata and (b) in case there's
-         # a COPY later. This will still be fast most of the time since the
-         # correct branch is likely to be checked out already.
+      if (self.image_i > 0):
+         # We need to check out the previous stage (a) to read its metadata
+         # and (b) in case there's a COPY later. This will still be fast most
+         # of the time since the correct branch is likely to be checked out
+         # already.
          self.parent.checkout()
          self.parent.ready()
       # At this point any meaningful parent of FROM, e.g., previous stage, has
--- a/lib/build.py
+++ b/lib/build.py
@@ -1047,12 +1047,12 @@ class I_from_(Instruction):
 
    @property
    def str_(self):
-      if (hasattr(self, "base_image")):
-         base_image = str(self.base_image.ref)
+      if (hasattr(self, "image") and hasattr(self.image, "alias")):
+         base_text = self.image.alias
       else:
          # Initialization failed, but we want to print *something*.
          base_image = self.base_text
-      return base_image + ((" AS " + self.alias) if self.alias else "")
+      return base_text + ((" AS " + self.alias) if self.alias else "")
 
    def checkout_for_build(self):
       assert (isinstance(bu.cache, bu.Disabled_Cache))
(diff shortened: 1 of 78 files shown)


j-ogas:
Observations:

1. There is a distinction between `tag`, `base_text`, and `alias` in the FROM instruction's prepare method.
    -  `tag`, local variable, is the actual image reference to pull or locate in storage, e.g., `alpine:latest`, `alpline_stage%s`.
    -  `base_text`, instruction attribute, is the grammar terminals of the left side of a FROM instruction, e.g., `foo`, from `FROM foo`.
    - `alias`, instruction class attribute, is the grammar terminal of the right side of a FROM instruction, e.g., `as bar` from `FROM foo as bar`
 2. To fix the current issue each FROM instruction needs to check if the image reference generated from it's `base_text` is in the global dict `images`, as an alias. If found, use the existing image reference instead, otherwise proceed normally.
 3. the `the FROM instruction's `str` method, called by `announce_maybe()`, needs to be modified. Consider the following valid Dockerfile.
 
 ```
 ARG BASE=foo
 FROM $BASE as bar
 FROM bar as qux
 FROM bar
 ```
 From the observations above, this looks like the following.
 
FROM instruction: 1, multistage: 0, base_text: `$BASE`, instruction alias: `bar`, tag: `SOMETAG_stage%0`
FROM instruction: 2, multistage: 1, base_text: `bar`, instruction alias: `qux`, tag: `SOMETAG_stage%1`
FROM instruction: 3, multistage: 2, base_text: `bar`, instruction alias: `None`, tag: `SOMETAG`
 
 The FROM instructions announce and string methods can only access 'self.alias', or `self.base_image.ref`, which would produce the following output:
 
 ```
 FROM foo as bar
 FROM SOMETAG_stage%1 as qux
 FROM SOMETAG
 ```
 It should be as follows.
 ```
 FROM foo as bar    (resolve ARG variable, pull foo, store as _stage%0)
 FROM bar as qux    (report alias used, use _stage%0, store as _stage%1)
 FROM bar           (report alias used, use _stage%1, store as SOMETAG)
 ```

reidpr (changes requested):
Overall looks pretty solid. Let's add a test though.

reidpr (commented):
discussed offline

lib/build.py
```diff
@@ -1141,6 +1150,8 @@ class Run(Instruction):
 
    __slots__ = ("cmd")
 
+
+
```
j-ogas, line 1154:
```suggestion
```

lib/filesystem.py
```diff
@@ -430,6 +430,7 @@ def symlink(self, target, clobber=False):
          ch.FATAL("can’t symlink: %s -> %s: %s" % (self.name, target,
                                                    x.strerror))
 
+
```
j-ogas, line 433:
```suggestion
```

lib/build.py
```diff
@@ -1097,11 +1105,12 @@ def prepare(self, miss_ct):
       if (str(self.image.ref) == str(self.base_image.ref)):
          ch.FATAL("output image ref same as FROM: %s" % self.base_image.ref)
       # Close previous stage if needed.
-      if (miss_ct == 0 and self.image_i > 0):
-         # While there haven’t been any misses so far, we do need to check out
-         # the previous stage (a) to read its metadata and (b) in case there’s
-         # a COPY later. This will still be fast most of the time since the
-         # correct branch is likely to be checked out already.
+      if (        self.image_i > 0
+          and not isinstance(bu.cache, bu.Disabled_Cache)):
```
reidpr, line 1109: Need to figure out why `Disabled_Cache` fails rather than just excluding it.

test/build/50_dockerfile.bats
```diff
@@ -877,6 +877,39 @@ EOF
 }
 
 
+@test 'Dockerfile: FROM multistage alias' {
+    scope standard
+    run build_ --no-cache -t tmpimg -f - . <<'EOF'
+ARG BASEIMG=alpine:3.17
+FROM $BASEIMG as a
```
reidpr, line 884: Comment the reasoning to use `ARG` here.

test/build/50_dockerfile.bats
```diff
@@ -877,6 +877,39 @@ EOF
...
+    echo "$output"
+    [[ $status -eq 0 ]]
+    if [[ $CH_TEST_BUILDER = ch-image ]]; then
+        [[ $output = *"ARG BASEIMG='alpine:3.17'"* ]]
+        [[ $output = *'FROM alpine:3.17 AS a'* ]]
+        [[ $output = *'RUN echo foo'* ]]
+        [[ $output = *'FROM a AS b'* ]]
+        [[ $output = *'RUN echo bar'* ]]
+        [[ $output = *'FROM b'* ]]
+        [[ $output = *'RUN echo qux'* ]]
```
reidpr, line 901: Do we really need to validate all the instructions? If not, reduce to just the one you're concerned about; otherwise, comment to explain.

reidpr (changes requested):
Overall looks nice but I'm concerned about the exception for `Disabled_Cache`.

lib/build.py
j-ogas: I should note that this is different from, and doesn't interfere with copying `stage_0` to `stage_1`, which occurs later.

reidpr: Sounds like the logic within `Disabled_Cache` needs to be fixed somehow then?

j-ogas: > Sounds like the logic within Disabled_Cache needs to be fixed somehow then?

Hmm, perhaps. FWIW, I did consider it.

It appears to me that the issue is that the `FROM` instruction subclass needs to behave differently when the build cache is disabled. The build cache class is designed to be instruction agnostic, the methods behave a certain way based on the cache mode, not the calling Dockerfile instruction.

Recall that the FROM instruction calls it's parent instruction's checkout method; the parent could be any one of the other 16 instruction subclasses, some of which make no changes to the image directory at all. This is why FROM needs to know if the build cache is enabled or not, prior to calling a parent's checkout -- it only needs to when the cache is enabled.

If you have ideas, let me know. I spent quite a bit of time here but I certainly may have missed something.

reidpr: OK. Let me send it back to myself for review and I'll see if I can think of something.

reidpr (commented):
Discussed offline. Thanks for your hard work and patience on this subtle situation.

reidpr (approved):

PR #1584, opened 2023-03-06
update fedora build script for python 3

lucaudill:


j-ogas (changes requested):
Does `shlex` come standard with python3? If not, we should probably guard against the case where it is missing. Alternatively, we may want to update autoconf to only install this script if the requisites are met.

lucaudill:
> 

`shlex` is listed on the page for the [python standard library](https://docs.python.org/3/library/). From that page:

> [T]his library reference manual describes the standard library that is distributed with Python. It also describes some of the optional components that are commonly included in Python distributions.

I'm not sure if this means that `shlex` is _always_ bundled with python, or if it's just "commonly included"...

reidpr:
I'm not aware of any stdlib modules that don't come standard, unless for the wrong OS or something like that. Regardless we do use `shlex` elsewhere already.

reidpr (approved):

PR #1585, opened 2023-03-06
cache gc: use topological sort

reidpr:
This PR adds topological commit sort to the “`git rev-list`” call in garbage collection that enumerates the commits, which discards the very first commit, assuming this is the root commit.

Git only records timestamps with one-second granularity. Previously, if the cache (and its root commit) was created, then received its first real commit within the same second (e.g., the test suite makes commits really fast), those two commits appear in arbitrary order, making the above assumption a race condition.


lucaudill (approved):
LGTM!

PR #1596, opened 2023-03-13
drop support for some old storage

### LINKED ISSUE #1243: remove code to move pre-v0.26 default storage directory on 2023-01-20
With PR #1242 merged, we have code to automatically move the default storage directory from old to new path. When the old default is stale enough, remove this.

### LINKED ISSUE #1588: retire storage prior to v0.25 (v1) on 2022-09-17
Storage version 2 was introduced in version 0.25, released Sept. 17, 2021. Per #1507, this issue is to remove support for storage v1.

### LINKED ISSUE #1592: stop moving pre-0.26 storage in default location on 2023-01-20
Charliecloud 0.26, released Jan. 20, 2022, changed the default storage location from `/var/tmp/$USER/ch-image` to `/var/tmp/$USER.ch/ch-image`. Since then, we've been moving storage directories in the old default location to the new one. Per #1507, support for old storage ends after one year, so let's stop moving it.

reidpr:
Closes #1243. Closes #1588.


lucaudill (approved):
LGTM!

PR #1597, opened 2023-03-14
add Flux and ch-fromhost improvements

j-ogas:


bin/ch_core.c
```diff
@@ -68,6 +68,8 @@ struct bind BINDS_DEFAULT[] = {
    { "/var/opt/cray/alps",       "/var/opt/cray/alps",       BD_OPTIONAL },
    /* Cray Shasta/Slingshot bind-mounts. */
    { "/var/spool/slurmd",        "/var/spool/slurmd",        BD_OPTIONAL },
+   /* FLUX job information bind-mount. */
+   { "/var/tmp",                 "/var/tmp",                 BD_OPTIONAL },
```
reidpr, line 72: This seems too broad — it makes all of `/var/tmp` shared with the host and you can't turn it off. (Right?) Can we narrow it to just the actual Flux directory?

reidpr (changes requested):

bin/ch_core.c
j-ogas: It doesn't appear so. The Flux WLM makes use of a [URI scheme](https://flux-framework.readthedocs.io/projects/flux-core/en/latest/man1/flux-uri.html). The vadar & shared memory related job info are stored as follows.

Note that the hash-looking segments, with exception to the character `ƒ`, are made up. `${USER}` is the user who runs.
```
/var/tmp/${USER}/flux-p5Nca5/jobtmp-0-ƒ56rVV28X/vader_segment.[...]
```

I considered an optional bind-mount of just `/var/tmp/${USER}`, however, unless that user-specific path is created at buildtime, the optional bind wouldn't work and we have the same problem we have now: it won't work with Flux without user manipulation.

Any ideas?

reidpr: Can we bind the Flux directory to something predictable, then tell Flux where it is?

Do the Flux people have any advice?

I don't want to do `/var/tmp` by default as it really is way too broad.

reidpr (commented):
see thread above

bin/ch_core.c
j-ogas: > I don't want to do /var/tmp by default as it really is way too broad.

I disagree.

Unlike standard unix desktops or shared environments, HPC resources mount both `/var/tmp` and `/tmp` as tmpfs, and store temporary files only for the duration of the running job. Mounting /var/tmp by default isn't an issue from a HPC job perspective.

I can understand the concern when `/var/tmp` isn't tmpfs, on some shared unix environment, e.g., desktop or front-end, where files from other users are stored via sticky bit and world readable. Charliecloud is for HPC; `ch-run` shouldn't be more difficult to use with Flux because of shared environment concerns. 

FWIW, I think the Flux design decision to use node tmpfs to store node specific job information is very reasonable. Charliecloud should work with FLUX without needing to rely on a FAQ, or tribal knowledge of non-default binds.

bin/ch-fromhost
```diff
@@ -145,7 +145,7 @@ enqueue_file () {
     old_ifs="$IFS"
     IFS="$newline"
     d="${dest:-$2}"
-    debug "queueing files"
+    debug "enqueue file(s)"
```
reidpr, line 148: I’d leave this as “queueing files” because “enqueue” is a rather esoteric word and I’m not sure we should give it to users.

bin/ch-fromhost
```diff
@@ -337,7 +337,6 @@ if [ -n "$cray_fi_found" ]; then
             continue
         fi
         enqueue_file "$l"
-        [ ! -L "$l" ] || enqueue_file "$(readlink -f "$f")"
```
reidpr, line 340: Why was this line removed? Looks like some kind of symlink handling?

examples/Dockerfile.libfabric
```diff
@@ -143,9 +143,16 @@ RUN git clone https://github.com/openpmix/openpmix.git \
...
-ARG FLUX_VERSION=0.45.0
+#
 # Flux requires a number of additional packages. We install them here to
 # distinquish between libfabric provider and flux-pmi dependencies.
+#
+# Flux pmi library, libpmi.so, conflicts with SLURM pmi2. To use FLUX you must
+# set LD_PRELOAD, as follows, in the container at run time.
+#
+#ENV LD_PRELOAD=/usr/local/lib/flux/libpmi.so
+#
```
reidpr, line 154: In principle we could change the name of one or the other, or configure `/usr/local/lib/flux to be a higher priority than wherever Slurm’s library is.

bin/ch-fromhost
j-ogas: Aw, I like it. This is debug output, are you sure?

examples/Dockerfile.libfabric
j-ogas: Hmm, good point, I'll try that.

bin/ch-fromhost
reidpr: Never mind; I'm convinced.

examples/Dockerfile.libfabric
j-ogas: Hmm, I think the issue here was actually due to misconfiguring Openmpi by linking the wrong flux-pmi library. See new commit f493eab. I was incorrectly linking `-lflux-core`, instead of `-lpmi` as prescribed by the `flux-pmi.pc` file.

bin/ch-fromhost
j-ogas: > Does it rebuild the two symlinks in the container though?

It does not. FWIW, it does not appear to matter. See below.

```
$ ./bin/ch-fromhost -v -p /qux/libfabric.so /tmp/ompi/
[ debug ] enqueue file(s)
[ debug ]    shared library: /foo/bar/libdbz.so.9000
[...]
[ debug ] injecting into image: /tmp/ompi/
[ debug ]    /foo/bar/libdbz.9000 -> /usr/local/lib (inferred)
done
$ ./bin/ch-run /tmp/ompi/ -- ldd /usr/local/lib/libfabric.so
[...]
	libdbz.so.9000 => /usr/local/lib/libdbz.so.9000 (0x10101robotsolo-thehumansaredead)
[...]
```

The overwritten library in the container, `/usr/local/lib/libfabric.so`, still resolves the library dependency, and runs as expected on single and multiple nodes.

j-ogas: The same is true for the other libraries pulled in.

reidpr: Looks like `ldconfig` will make the appropriate symlinks, and we run that anyway.

PR #1602, opened 2023-03-17
upgrade v5 build cache correctly

reidpr:


lib/build_cache.py
```diff
@@ -744,11 +744,37 @@ def configure(self):
...
+      # to keep bad .gitignore from creeping back in. (I couldn’t figure out
+      # how to fix this for “filter-branch” either, which I didn’t want to use
+      # anyway for the reasons above.)
+      if (ch.storage.bucache_needs_ignore_upgrade.exists_()):
+         ch.INFO("upgrading build cache to v6+, some cached data may be lost",
+                 "see release notes for v0.32")
+         old = self.git(["fast-export", "--no-data", "--", "--all"]).stdout
+         new = re.sub(r"^(D|M [0-7]+ [0-9a-f]+) \.(git|weirdal_)ignore$",
+                      "#\g<0>", old, flags=re.MULTILINE)
+         #fs.Path("/tmp/new").file_write(new)
```
mphinney1100, line 774: do you want this comment?

reidpr: yeah, it was to show how to debug since it wasn't a standard method

reidpr:
Currently the test are failing; I'm pretty sure this is spurious. This branch does pass `old-storage` on my dev box.

lib/build_cache.py
```diff
@@ -744,11 +744,58 @@ def configure(self):
...
       #
-      # 1. Git has no default gitignore, but cancel any global gitignore rules
-      #    the user might have. https://stackoverflow.com/a/26681066
+      #   1. Git has no default gitignore, but cancel any global gitignore rules
+      #      the user might have. https://stackoverflow.com/a/26681066
       #
-      # 2. The oddly-named GIT_DIR.
+      #   2. The oddly-named GIT_DIR.
+      #
+      # It it easier to just write list we want every time, rather than trying
```
j-ogas, line 752:
```suggestion
      # It it easier to just write the list we want every time, rather than trying
```

j-ogas (changes requested):
See comments below. I'm assuming this upgrade will use existing tests? If not, perhaps we should add them?

reidpr:
> I'm assuming this upgrade will use existing tests?

Yes, that's a good point, and things are a little weird. The key test here is `test/old-storage`. So in another already-merged PR, we dropped support for v1 storage directories, but the internal CI has not yet been updated to handle that: it still runs `old-storage` on all archived storage directories we have. That is, once we merge this, the internal full-scope test that we normally run prior to a release cannot be run (until we fix it, which might be several days). However, the branch does pass `old-storage` on my development box.

j-ogas (approved):

PR #1617, opened 2023-03-29
ch-image: use previously prepared files

### LINKED ISSUE #1359: ch-image: use previously prepared config
Related to #1309.

`ch-image` always pushes a unique config; see `push.py` comment below.

> There is not currently any support for re-using any previously prepared files already in the upload cache, because we don't yet have a way to know if these have changed until they are already build.

This is no longer true with the introduction of the build cache. Update `push.py` to use cached config if the cache is enabled.

BenStormer:
Closes #1359


lib/push.py
```diff
@@ -82,19 +84,51 @@ def manifest_new(class_):
...
-
-         There is not currently any support for re-using any previously
-         prepared files already in the upload cache, because we don't yet have
-         a way to know if these have changed until they are already build."""
+     # # Delete the tarballs since we can't yet cache them.
+     # for (_, tar_c) in self.layers:
+     #    ch.VERBOSE("deleting tarball: %s" % tar_c)
+     #    tar_c.unlink_()
+
+   def get_prev_prepared(self):
```
j-ogas, line 92:
```suggestion
   def prepare_existing(self):
```

In general, we try to avoid starting methods with the verb `get`. This is because: 1) we typically order methods alphabetically, and 2) we try to maintain method locality for related methods.

In this case the verb 'prepare' is common for both methods, so we should name the function something like `prepare_existing`. This keeps the methods in alphabetical order.

lib/push.py
```diff
@@ -82,19 +84,51 @@ def manifest_new(class_):
...
+         if not."""
+      (sid, git_hash) = bu.cache.find_image(self.image)
+
+      conf_suffix = git_hash + ".config.json"
+      conf_path = ch.storage.upload_cache // conf_suffix
+      try:
+         json.load(open(conf_path))
+      except:
+         print("Previous config does not exist")
+         return None
```
j-ogas, line 104: Try to avoid using try catch blocks when possible. Here we can check if the file exists before attempting to load the json contents. See the module `os`; I believe there are function wrappers in one of the charliecloud library that wraps it, e.g., `ossafe()`.

Also, if there is an existing function that loads json file contents use that, rather than `json.load()`.

lib/push.py
```diff
@@ -110,10 +144,18 @@ def prepare(self):
...
                                      "size": size_c,
                                      "digest": "sha256:" + hash_c })
+        # # RENAME TAR WITH HASH, using Path rename function
+        # hash_str = str(hash_uc) + ".tar.gz"
+        # print("Original path:", path_c)
+        # target = Path(ch.storage.upload_cache // hash_str)
+        # print("target:", target)
+        # path_c.rename(hash_str)
+        # print(path_c)
+        # print("LOOK ABOVE")
```
j-ogas, line 157:
```suggestion
```

lib/push.py
```diff
@@ -149,6 +191,33 @@ def prepare(self):
...
+         manifest as a sequence of bytes).
+
+         There is not currently any support for re-using any previously
+         prepared files already in the upload cache, because we don't yet have
+         a way to know if these have changed until they are already build."""
+      # TODO: Check if build cache is even enabled first before below
+      # CHECK: use prev_prepeared to check if git-head.conf, etc. exists
+      prev_prepared = self.get_prev_prepared()
+      if prev_prepared is None:
+         prev_prepared = self.prepare_new()
```
j-ogas, line 218: I'm not sure this logic needs to be split into three methods. Consider putting all the logic here in `prepare()`.

j-ogas (changes requested):

j-ogas (changes requested):
Thanks! See comments below. One thing we will want to do, which I don't see, is to name the compressed tarballs as `${SHA256}.tar.gz`

lib/push.py
BenStormer: I put some more thought into the structure of this, and actually ended up going further in the opposite way. I agree it was a bit oddly structured how I had it, but rather than put all logic in ```prepare()``` I have split it up a bit further. It is now structured similar to ```push()```; meaning each of aspects of preparing to push (as I view it) are now independent functions. These aspects are:

1. check for existing config, manifest, and tar
    a. If not present, prepare new versions of these files
2. Store config bytes, manifest bytes, and layers

If this still seems worse than putting all logic in ```push()``` I am happy to change it back. This is the way that seemed slightly best to me, but I am curious what you think.

lib/push.py
```diff
@@ -80,21 +82,63 @@ def manifest_new(class_):
...
+      self.config = config_bytes
+      self.manifest = manifest_bytes
 
-   def prepare(self):
-      """Prepare self.image for pushing to self.dst_ref. Return tuple: (list
-         of gzipped layer tarball paths, config as a sequence of bytes,
-         manifest as a sequence of bytes).
+   def prepare_existing(self):
+      """Check self.image if an existing config, manifest, and tar exist.
+         Return tuple: (config, manifest, tar) if these files exist."""
```
j-ogas, line 113:
```suggestion
      """Attempt to reuse previously prepared self.image config, manifest,
          and layer files."""
```

lib/push.py
```diff
@@ -80,21 +82,63 @@ def manifest_new(class_):
...
+      (sid, git_hash) = bu.cache.find_image(self.image)
```
j-ogas, line 114:
```suggestion
      ch.DEBUG("checking for previously prepared files ")
      (sid, git_hash) = bu.cache.find_image(self.image)
      if ((sid, githash) == (None, None)):
         return
```
If the find image method misses we do not need to check for the existence of any files.

lib/push.py
```diff
@@ -80,21 +82,63 @@ def manifest_new(class_):
...
+      conf_suffix = git_hash + ".config.json"
+      conf_path = ch.storage.upload_cache // conf_suffix
+      if (conf_path.exists()):
+         config = conf_path.json_from_file("config")
+      else:
+         print("Previous config does not exist")
+         return
```
j-ogas, line 121:
```suggestion
      if (not conf_path.exists()):
         return
      config = conf_path.json_from_file("config")
 ```
 Use `not` to simplify.

File: lib/push.py [hunk 8]
```diff
@@ -80,21 +82,63 @@ def manifest_new(class_):
...
+         print("Previous manifest does not exist")
+         return
+      tar_c = str(self.image.ref) + ".tar.gz"
+      path_c = ch.storage.upload_cache // tar_c
+      if (path_c.exists()):
+         hash_c = path_c.file_hash()
+         tars_c = [(hash_c, path_c)]
+      else:
+         print("Previous tar does not exist")
+         return
```
j-ogas, line 136:
```suggestion
      if (not path_c.exists()):
         return
      hash_c = path_c.file_hash()
      tars_c = [(hash_c, path_c)]
```

lib/push.py
```diff
@@ -80,21 +82,63 @@ def manifest_new(class_):
...
+      return (config, manifest, tars_c)
```
j-ogas, line 137:
```suggestion
      self.prepare_upload(config, manifest, tars_c)
```
Instead of returning anything, we just set the class instance attributes.

lib/push.py
```diff
@@ -80,21 +82,63 @@ def manifest_new(class_):
...
 
-         There is not currently any support for re-using any previously
-         prepared files already in the upload cache, because we don't yet have
-         a way to know if these have changed until they are already build."""
+   def prepare_new(self):
+      """Prepare and return a new config, manifest, and tarball since they do
+         not all already exist."""
```
j-ogas, line 141:
```suggestion
      """Prepare new config, manifest, and layer files for pushing."""
```

lib/push.py
```diff
@@ -149,23 +193,31 @@ def prepare(self):
...
-      self.manifest = manifest_bytes
+      (sid, git_hash) = bu.cache.find_image(self.image)
+      ul_ref = str(ch.storage.upload_cache)
+      conf_path = ul_ref + "/" + git_hash + ".config.json"
+      man_path = ul_ref + "/" + git_hash + ".manifest.json"
+      with open(Path(conf_path), "w") as conf:
+         json.dump(config, conf)
+      with open(Path(man_path), "w") as man:
+         json.dump(manifest, man)
+      return (config, manifest, tars_c)
```
j-ogas, line 204:
```suggestion
      self.prepare_upload(config, manifest, tars_c)
```

lib/push.py
```diff
@@ -80,21 +82,63 @@ def manifest_new(class_):
...
+#      for (_, tar_c) in self.layers:
+#         ch.VERBOSE("deleting tarball: %s" % tar_c)
+#         tar_c.unlink_()
+
+   def pack_prepared_image(self, prepared):
+      """Store a list of gzipped layer tarball paths, config as a sequence of
+       bytes, and manifest as a sequence of bytes."""
+      config = prepared[0]
+      manifest = prepared[1]
+      tars_c = prepared[2]
```
j-ogas, line 97:
```suggestion
```
These variables are passed to the method.

j-ogas (changes requested):

lib/push.py
BenStormer: Thanks for the help! This new logic is a lot smarter and simpler.

doc/ch-image.rst
```diff
@@ -1234,6 +1239,26 @@ in the remote registry, so we don’t upload it again.)
    cleaning up
    done
 
+Same, except the local image has already been pushed to the destination in its
+current form. Charliecloud is able to expedite the pushing process by using
+previously prepared config, metadata, and tarballs from the previous push. If
+not already present, the :code:`--cache-upload` flag would store the generated
+files in the upload cache.
```
j-ogas, line 1246:
```suggestion
Same but cache the prepared upload files. Subsequent pushes use cached upload
files, expediting the push process. Useful when pushing a large image to multiple
repositories, or push debugging.
```

doc/ch-image.rst
```diff
@@ -1234,6 +1239,26 @@ in the remote registry, so we don’t upload it again.)
...
+   pushing image:   example.com:5000/foo/bar:latest
+   image path:      /var/tmp/image
+   starting upload
+   layer 1/1: 5ffaf42: checking if already in repository
+   layer 1/1: 5ffaf42: already present
+   config: 2966628: checking if already in repository
+   config: 2966628: already present
+   manifest: uploading
+   cleaning up
+   done
```
j-ogas, line 1260: This example should go directly under the first push example at line 1181.

doc/ch-image.rst
```diff
@@ -1257,6 +1282,22 @@ from the cache. Only available when the cache is enabled, and will not
 overwrite :code:`IMAGE_REF` if it exists.
 
 
+:code:`build-cache`
```
j-ogas, line 1285:
```suggestion
:code:`upload-cache`
```

doc/ch-image.rst
```diff
@@ -1257,6 +1282,22 @@ from the cache. Only available when the cache is enabled, and will not
...
+
+   $ ch-image [...] upload-cache [--reset]
+
+Print basic information about the upload cache.
+
+If the following option is given, do the corresponding operation before
+printing.
+
+  :code:`--reset`
+    Clear and re-initialize the upload cache.
```
j-ogas, line 1298:
```suggestion
    Delete all cached upload files
```

lib/push.py
```diff
@@ -37,16 +38,18 @@ def main(cli):
...
-   # Note; We use functions to create the blank config and manifest to to
+   # Note; We use functions to create the blank config and manifest to
    # avoid copy/deepcopy complexity from just copying a default dict.
 
-   __slots__ = ("config",    # sequence of bytes
-                "dst_ref",   # destination of upload
-                "image",     # Image object we are uploading
-                "layers",    # list of (digest, .tar.gz path), lowest first
-                "manifest")  # sequence of bytes
+   __slots__ = ("cache_upload",  # bool if upload cache is enabled
```
j-ogas, line 44: Adding an attribute, in this case a boolean, to augment behavior is certainly one approach. Consider class inheritance and how that may change your logic. For example:

```python
class Image_Pusher:
[...]
class Image_Pusher_Cache_Enabled (Image_Pusher):
```
`Image_Pusher_Cache_Enabled` can then augment methods of interest, e.g., `prepare(), `prepare_upload()`, `prepare_existing()`, etc. without relying on various `if self.cache_ulpoad` checks.

lib/misc.py
```diff
@@ -183,3 +183,24 @@ def undelete(cli):
    if (git_hash is None):
       ch.FATAL("image not in cache")
    bu.cache.checkout(img, git_hash, None)
+
+def upload_cache(cli):
+   #TODO: Make upload-cache an independent file and call functions from there
```
j-ogas, line 188: May as well go ahead and do that.

j-ogas (changes requested):
Excellent. This is starting to shape up nicely. See in-line suggestions and comments.

doc/ch-image.rst
BenStormer: Whoops, good catch. Thanks

lib/misc.py
BenStormer: This should be implemented and working now, however its in a pretty simply made file and I'm sure there is improvement on how to set up a file like this that could be made.

j-ogas (changes requested):
Discussed offline.

lib/push.py
```diff
@@ -28,7 +29,14 @@ def main(cli):
       ch.INFO("destination:     %s" % dst_ref)
    else:
       dst_ref = im.Reference(cli.source_ref)
-   up = Image_Pusher(image, dst_ref)
+   if (cli.cache_upload):
+      # If build cache is disabled, we can't know if the image has changed
+      # since the last push.
+      if (isinstance(bu.cache, bu.Disabled_Cache)):
+         ch.FATAL("can't use upload cache with build cache disabled")
```
j-ogas, line 36:
```suggestion
         ch.FATAL("cache disabled")
```

lib/push.py
```diff
@@ -37,14 +45,14 @@ def main(cli):
...
-   __slots__ = ("config",    # sequence of bytes
-                "dst_ref",   # destination of upload
-                "image",     # Image object we are uploading
-                "layers",    # list of (digest, .tar.gz path), lowest first
-                "manifest")  # sequence of bytes
+   __slots__ = ("config",        # sequence of bytes
+                "dst_ref",       # destination of upload
+                "image",         # Image object we are uploading
+                "layers",        # list of (digest, .tar.gz path), lowest first
+                "manifest")      # sequence of bytes
```
j-ogas, line 55:
```suggestion
   __slots__ = ("config",    # sequence of bytes
                "dst_ref",   # destination of upload
                "image",     # Image object we are uploading
                "layers",    # list of (digest, .tar.gz path), lowest first
                "manifest")  # sequence of bytes
```

lib/push.py
```diff
@@ -81,24 +89,75 @@ def manifest_new(class_):
                "weirdal": "yankovic" }
 
    def cleanup(self):
+      # Delete the config, manifest, and tarballs if the cache is disabled.
+      conf_suffix = str(self.image) + ".config.json"
+      conf_path = ch.storage.upload_cache // conf_suffix
+      man_suffix = str(self.image) + ".manifest.json"
+      man_path = ch.storage.upload_cache // man_suffix
+      if (not conf_path.exists() or not man_path.exists()):
+         return
```
j-ogas, line 98:
```suggestion
      # Delete uncached tarballs.
```
Push should behave exactly like it does now when not using `--cache-upload`. Other files are left alone and written over as needed.

lib/push.py
```diff
@@ -81,24 +89,75 @@ def manifest_new(class_):
...
       ch.INFO("cleaning up")
-      # Delete the tarballs since we can't yet cache them.
+      ch.VERBOSE("deleting config: %s" % conf_path)
+      conf_path.unlink()
+      ch.VERBOSE("deleting manifest: %s" % man_path)
+      man_path.unlink()
```
j-ogas, line 103:
```suggestion
```

j-ogas (changes requested):
See comments and discuss offline.

lib/push.py
```diff
@@ -176,3 +234,19 @@ def upload(self):
...
+   def rename_conf_and_man(self, git_hash):
+      conf_suffix = git_hash + ".config.json"
+      man_suffix = git_hash + ".manifest.json"
+      return (conf_suffix, man_suffix)
+
+   def rename_tar(self, git_hash, path_c):
+      tar_suffix = git_hash + ".tar.gz"
+      path_c.rename_(ch.storage.upload_cache // tar_suffix)
+      path_c = ch.storage.upload_cache // tar_suffix
+      return path_c
```
j-ogas, line 252:
```suggestion
```

See comment above.

lib/push.py
```diff
@@ -81,24 +89,75 @@ def manifest_new(class_):
...
-         manifest as a sequence of bytes).
+   def rename_conf_and_man(self, git_hash):
+      conf_suffix = str(self.image) + ".config.json"
+      man_suffix = str(self.image) + ".manifest.json"
+      return (conf_suffix, man_suffix)
+
+   def rename_tar(self, git_hash, path_c):
+      return path_c
+
+   def prepare_upload(self, config, manifest, tars_c):
```
j-ogas, line 116:
```suggestion
   def prepare_upload(self):
```

lib/push.py
```diff
@@ -81,24 +89,75 @@ def manifest_new(class_):
...
+      """Store a list of gzipped layer tarball paths, config as a sequence of
+       bytes, and manifest as a sequence of bytes."""
+      config_bytes = json.dumps(config, indent=2).encode("UTF-8")
+      config_hash = ch.bytes_hash(config_bytes)
+      manifest["config"]["size"] = len(config_bytes)
+      manifest["config"]["digest"] = "sha256:" + config_hash
+      ch.DEBUG("config: %s\n%s" % (config_hash, config_bytes.decode("UTF-8")))
+      manifest_bytes = json.dumps(manifest, indent=2).encode("UTF-8")
+      ch.DEBUG("manifest:\n%s" % manifest_bytes.decode("UTF-8"))
```
j-ogas, line 125: With the other suggested changes, we shouldn't need arguments passed to `prepare_upload()`. Here we just real the files from the `self.path_config`, `self.path_manfiest` etc.

j-ogas (changes requested):

lib/push.py
```diff
@@ -83,25 +106,113 @@ def manifest_new(class_):
...
       self.registry.request("GET", self.registry._url_base)
+
+      ch.VERBOSE('preparing push upload files')
+      config = None
+      manifest = None
+      layers = None
+
+      if (upload_cache is not None):
+         # Check for previously prepared; if they exist, use them.
+         config = self.json_from_path(self.path_config, 'config')
```
reidpr, line 182: What's the reasoning to re-parse the stored JSON rather than just sending the existing files?

lib/push.py
```diff
@@ -113,9 +224,11 @@ def prepare(self):
          hash_uc = path_uc.file_hash()
          config["rootfs"]["diff_ids"].append("sha256:" + hash_uc)
          size_uc = path_uc.file_size()
-         path_c = path_uc.file_gzip(["-9", "--no-name"])
+         path_c_ = path_uc.file_gzip(["-9", "--no-name"])
```
reidpr, line 227: I'm a little confused by `path_uc` (no underscore) and `path_uc_` (with underscore). Maybe use more verbose names?

test/build/50_localregistry.bats
```diff
@@ -40,6 +42,41 @@ EOF
...
+    echo "output"
+    [[ $status -eq 0 ]]
+    [[ $output = *'using previously prepared config'* ]]
+    [[ $output = *'using previously prepared manifest'* ]]
+    [[ $output = *'using previously prepared layer(s)'* ]]
+}
+
+
+@test "${tag}: upload cache error(s)" {
+    [[ $CH_IMAGE_CACHE == disabled ]] || skip 'build cache enabled'
```
reidpr, line 71: I'd gate this test with `[[ $CH_IMAGE_CACHE == enabled ]]`, then use `--no-cache` below. As is, I think it sets up a gotcha if there is more added to this function that doesn’t require the cache to be disabled.

reidpr (changes requested):
Nice, glad to see this feature moving along. Comments in-line.

doc/ch-image.rst
```diff
@@ -1286,6 +1294,23 @@ image must be named to match that remote reference.
...
+Same but cache the prepared upload files. Subsequent pushes use cached upload
+files, expediting the push process. Useful when pushing a large image to
+multiple repositories, or push debugging.
+
+::
+
+   $ ch-image push --ulcache example.com:5000/foo/bar:latest
+   pushing image:   example.com:5000/foo/bar:latest
+   starting upload
+   layer 1/1: 5ffaf42: checking if already in repository
```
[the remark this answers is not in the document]
j-ogas: The layer checking line is from a different segment of code. There is no added INFO output from this feature.

lib/push.py
j-ogas: I suppose there is no need to re-parse the stored config, however, the manifest needs to be read to get the sha256 digest for the tarball layers.

lib/image.py
```diff
@@ -435,18 +436,25 @@ def tarballs_write(self, tarball_dir):
...
       try:
          ch.INFO("layer 1/1: gathering")
          ch.VERBOSE("writing tarball: %s" % path)
-         fp = fs.TarFile.open(path, "w", format=tarfile.PAX_FORMAT)
-         unpack_path = self.unpack_path.resolve()  # aliases use symlinks
-         ch.VERBOSE("canonicalized unpack path: %s" % unpack_path)
-         fp.add_(unpack_path, arcname=".")
-         fp.close()
+         bytes_ = None
+         with io.BytesIO() as buf:
```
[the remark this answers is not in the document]
j-ogas: The idea was to calculate the uncompressed layer hash during the layer tarFile creation, rather than re-reading the file again later. TarFile is kinda hairy to me, unclear if I can read and write in chunks with it as the tar is being created.

j-ogas:
This branch is stale. There are merge conflicts on many of the modified files. Closing.

PR #1618, opened 2023-03-31
add bash completion to Charliecloud

lucaudill:
PR to add command line completion to Charliecloud for bash users. If successful, could be extended to other shells.

### DIFF
--- a/bin/ch-completion.bash
+++ b/bin/ch-completion.bash
@@ -5,6 +5,9 @@
 #
 # Resources for understanding this script:
 #
+#   * Everything bash:
+#     https://www.gnu.org/software/bash/manual/html_node/index.html
+#
 #   * Bash parameter expansion:
 #     https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html
 #
@@ -18,9 +21,54 @@
 #     https://unix.stackexchange.com/a/224564
 #
 #
-# SYNTAX GLOSSARY
+## SYNTAX GLOSSARY ##
+#
+# Source:
+# https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html
+#
+# ${array[i]}
+#   Gives the ith element of “array”. Note that bash arrays are indexed
+#   at zero, as all things should be.
+#
+# ${array[@]}
+#   Expands “array” to its member elements.
+#
+# ${#parameter}
+#   Gives the length of “parameter.” If “parameter” is a string, this
+#   expansion gives you the character length of the string. If “paramter”
+#   is an array subscripted by “@” or “*” (e.g. “foo[@]”), then the
+#   expansion gives you the number of elements in the array.
+#
+# ${parameter:offset:length}
+#   A.k.a. substring expansion. If “parameter” is a string, expand up to
+#   “length” characters, starting with the character at position “offset.”
+#   If “offset” is unspecified, start at the first character. If “parameter”
+#   is an array subscripted by “@” or “*,” (e.g. “foo[@]”) expand up to
+#   “length” elements, starting at the element at position “offset” (e.g.
+#   “${foo[offset]}”).
+#
+#   Example 1 (string):
+#   $ foo="foo_bar_baz"
+#   $ echo ${foo::7}
+#   foo_bar
+#   $ echo ${foo:1:7}
+#   oo_bar_
+#
+#   Example 2 (array):
+#   $ foo=("foo" "bar" "baz" "qux")
+#   $ echo ${foo[@]::3}
+#   foo bar baz
+#   $ echo ${foo[@]:1:3}
+#   bar baz qux
 #
 # FIXME: Add syntax glossary
+#
+
+# Possible extensions once this is merged:
+#   * Add completion support for non-bash shells (e.g. zsh and tchs).
+#     see https://github.com/git/git/tree/master/contrib/completion
+#   * Add support for mid-line completion (possibly using COMP_POINT,
+#     see https://devmanual.gentoo.org/tasks-reference/completion/index.html).
 #
 
 # FIXME: disable shellcheck SC2034 for this? (See base.sh)
@@ -62,8 +110,9 @@ _ch_image_completion () {
 
     # If "$cur" is non-empty, we want to ignore it as a potential subcommand
     # to avoid unwanted behavior. “${words[@]::${#words[@]}-1}” gives you all
-    # but the last element of the array “words,” so if "$cur" is non-empty,
-    # pass it to _ch_subcommand_get. Otherwise pass all elements of “words.”
+    # but the last element of the array “words” (see syntax glossary above) so 
+    # if "$cur" is non-empty, pass it to _ch_subcommand_get. Otherwise pass all
+    # elements of “words.”
     if [[ -n "$cur" ]]; then
         sub_cmd=$(_ch_subcommand_get "$_image_subcommands" "${words[@]::${#words[@]}-1}")
         strg_dir=$(_ch_find_storage "${words[@]::${#words[@]}-1}")
@@ -74,16 +123,13 @@ _ch_image_completion () {
     echo "sub command: $sub_cmd" >> /tmp/ch-completion.log
     echo "storage dir: $strg_dir" >> /tmp/ch-completion.log
     echo "len storage dir: ${#strg_dir}" >> /tmp/ch-completion.log
-    #if [[ -n "$strg_dir" ]]; then
-    #    strg_dir="-s $strg_dir"
-    #fi
 
     # Common opts that take args
     #
     case "$prev" in
     -a | --arch )
         # FIXME: Remove yolo?
-        # FIxME: Missing common architectures?
+        # FIXME: Missing common architectures?
         COMPREPLY=( $(compgen -W "host yolo 386 amd64 arm/v6 
                                   arm/v7 arm64/v8 ppc64le s390x" -- $cur) )
         return 0
--- a/bin/ch-completion.bash
+++ b/bin/ch-completion.bash
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+
+## ch-image ##
+
+# Subcommands for ch-image
+#
+__image_subcommands="build build-cache delete gestalt
+                     import list pull push reset undelete"
+
+__image_build_opts="-b --bing --build-arg -f --file --force
+                    --force-cmd -n --dry-run --parse-only -t --tag"
+
+__image_common_opts_noarg="--always-download --auth --cache --no-cache
+                           --no-lock --profile --rebuild --password-many
+                           --tls-no-verify -v --verbose"
+
+__image_common_opts_arg="-a --arch --cache-large -s --storage-dir"
+
+# Autocompletion function for ch-image
+#
+__ch-image_completion () {
+    local prev
+    local cur
+    local words
+    _get_comp_words_by_ref -n : cur prev words
+
+    local sub_command=$(__ch_subcommand_get "$__image_subcommands" "${words[@]}")
+
+    case "$prev" in
+    delete )
+        # FIXME: use filepath for ch-image or assume it's in PATH?
+        COMPREPLY=( $(compgen -W "$($SCRIPT_DIR/ch-image list)" -- $cur) )
+        __ltrim_colon_completions "$cur"
+        return 0
+        ;;
+    *ch-image )
+        COMPREPLY=( $(compgen -W "$__image_subcommands" -- $cur) )
+        return 0
+        ;;
+    esac
+}
+
+## Helper functions ##
+
+# Kludge up a way to look through array of words and determine the 
+# subcommand. Note that the double for loop doesn't take that much
+# time, since the charliecloud command line is relatively short.
+#
+# Usage:
+#   __ch_subcommand_get [subcommands] [words]
+#
+__ch_subcommand_get () {
+    local cmd subcmd
+    local cmds="$1"
+    shift 1
+    for word in "$@"
+    do
+        for cmd in $cmds
+        do
+            if [[ "$word" == "$cmd" ]]; then
+                subcmd="$cmd"
+            fi
+        done
+    done
+    echo "sub command: $subcmd" >> /tmp/ch-complete-sub_cmd.log
+    echo "$subcmd"
+}
+
+complete -F __ch-image_completion ch-image
--- a/bin/ch-completion.bash
+++ b/bin/ch-completion.bash
@@ -83,9 +83,6 @@ fi
 
 # Subcommands and options for ch-image
 #
-_image_subcommands="build build-cache delete gestalt
-                    import list pull push reset undelete"
-
 _image_build_opts="-b --bind --build-arg -f --file --force
                    --force-cmd -n --dry-run --parse-only -t --tag"
 
@@ -94,6 +91,9 @@ _image_common_opts="-a --arch --always-download --auth --cache
                     --no-cache --no-lock --profile --rebuild
                     --password-many -s --storage --tls-no-verify
                     -v --verbose --version"
+
+_image_subcommands="build build-cache delete gestalt
+                    import list pull push reset undelete"
 
 ## ch-image ##
 
@@ -120,9 +120,6 @@ _ch_image_completion () {
         sub_cmd=$(_ch_subcommand_get "$_image_subcommands" "${words[@]}")
         strg_dir=$(_ch_find_storage "${words[@]}")
     fi
-    echo "sub command: $sub_cmd" >> /tmp/ch-completion.log
-    echo "storage dir: $strg_dir" >> /tmp/ch-completion.log
-    echo "len storage dir: ${#strg_dir}" >> /tmp/ch-completion.log
 
     # Common opts that take args
     #
@@ -256,7 +253,27 @@ _ch_image_completion () {
     fi
 }
 
+## ch-run ##
+
+# Completion function for ch-run
+#
+_ch_run_completion () {
+    echo "does nothing"
+}
+
 ## Helper functions ##
+
+# Use this function to enable or disable completion after sourcing
+# this file. Useful if there are issues.
+# FIXME: Add other executables as their completion functions are
+#        implemented.
+ch-completion () {
+    if [[ "$1" == "disable" ]]; then
+        complete -r ch-image
+    elif [[ "$1" == "enable" ]]; then
+        complete -F _ch_image_completion ch-image
+    fi
+}
 
 # Figure out which storage directory to use (including cli-specified
 # storage).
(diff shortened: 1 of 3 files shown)


bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+        COMPREPLY=()
+        return 0
+        ;;
+    -s | --storage )
+        # This “if” helps avoid overzealous completion. E.g. if there's
+        # only one subdir of the current dir, this command completes to
+        # that dir even if "$cur" is empty. I didn't like that, hence the
+        # “if”.
+        if [[ -n "$cur" ]]; then
+            compopt -o nospace
```
reidpr, line 145: Does setting this option persist to future calls of `_ch_image_completion`? If so, should probably set default options at top of the function.

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+            ;;
+        *)
+            # Autocomplete to context directory, common opt, or build-specific opt
+            echo "compgen bit:" >> /tmp/ch-completion.log
+            # --force can take “fakeroot” or “seccomp” as an argument, or no
+            # argument at all. To account for this, we add those two arguments
+            # to the list of compgen suggestions, which allows compgen to autocomplete
+            # to “fakeroot,” “seccomp,” or anything that could logically follow
+            # “--force” with no argument.
+            if [[ "$prev" == "--force" ]]; then
```
reidpr, line 182: Does this deal with equals-separated options too? E.g. `--force=seccomp` is also valid.

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+                extras="$extras fakeroot seccomp"
```
reidpr, line 183:
```suggestion
                extras=+=' fakeroot seccomp'
```

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+            fi
+            COMPREPLY=( $(compgen -W "$_image_build_opts $extras"  -- $cur) )
+            # Completion for the context directory. Note that we put this under an
+            # “if” statement so that the “nospace” option isn't applied to all
```
reidpr, line 187: This is another example of a comment that should be more about the reasoning than the code. That is, we can see it's under an if statement by looking at the code; the comment should explain why.

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+            fi
+            ;;
+        esac
+        ;;
+    build-cache )
+        COMPREPLY=( $(compgen -W "--reset --gc --tree --dot" -- $cur) )
+        ;;
+    delete | list)
+        # FIXME: Is it janky to include “list” here? I'm leaning towards
+        #        “not that janky,” others may disagree.
```
reidpr, line 202: It's not janky but you should use fall-through rather than `if`.

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+            return 0
+        # The following check seems to fix a bug where the completion
+        # function initialzes an empty storage directory.
+        elif [[ -n "$(ls "$strg_dir"/img)" ]]; then
+            COMPREPLY=( $(compgen -W "$($CH_BIN/ch-image list -s $strg_dir) --image" -- "$cur") )
+            _ltrim_colon_completions "$cur"
+        fi
+        ;;
+    undelete )
+        # FIXME: Update with “ch-image undelete --list” once #1551 drops
```
reidpr, line 238: lol this is why you should gather this stuff from `ch-image` directly ;)

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+# this file. Useful if there are issues.
+# FIXME: Add other executables as their completion functions are
+#        implemented.
+ch-completion () {
+    if [[ "$1" == "disable" ]]; then
+        complete -r ch-image
+    elif [[ "$1" == "enable" ]]; then
+        complete -F _ch_image_completion ch-image
+    fi
+}
```
reidpr, line 276: I wonder if a simpler procedure might be:

1. Enable by sourcing the file (sourcing should be idempotent anyway).
2. Disable by calling some function (without an argument).

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+_ch_find_storage () {
+    if [[ -n "$(grep -Eo "(\-\-storage|[^\-]\-s)" <<< "$@")" ]]; then
+        # This sed only works as desired if “--storage” or “-s” are present
+        # in the command line...
+        sed -E 's/(.*)(--storage|[^-]-s)\ *([^ ]*)(.*$)/\3/g' <<< "$@"
+    elif [[ -n "$CH_IMAGE_STORAGE" ]]; then
+        echo "$CH_IMAGE_STORAGE"
+    else
+        echo "/var/tmp/$USER.ch/"
+    fi
```
reidpr, line 291: Probably better to `gestalt` this. There might even be an existing option.

Note: I recommend you `gestalt` everything you need once, when the script is sourced, to avoid lots of `exec(2)` while the completion function is doing its thing.

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+        complete -F _ch_image_completion ch-image
+    fi
+}
+
+# Figure out which storage directory to use (including cli-specified
+# storage).
+# FIXME: Can probably cook up a sed pattern that'll remove the need
+#        for the “if” statement.
+_ch_find_storage () {
+    if [[ -n "$(grep -Eo "(\-\-storage|[^\-]\-s)" <<< "$@")" ]]; then
```
reidpr, line 283: wow4

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+        complete -r ch-image
+    elif [[ "$1" == "enable" ]]; then
+        complete -F _ch_image_completion ch-image
+    fi
+}
+
+# Figure out which storage directory to use (including cli-specified
+# storage).
+# FIXME: Can probably cook up a sed pattern that'll remove the need
+#        for the “if” statement.
```
reidpr, line 281: I think this might be hard because you want the replacement to differ based on which part of the pattern matched. But, you can simplify the `grep` I think:

```
$ echo 'ch-image -s foo -a' | grep -Eq -- '\s(--storage|-\w*s)'; echo $?
0
$ echo 'ch-image -as foo -a' | grep -Eq -- '\s(--storage|-\w*s)'; echo $?
0
$ echo 'ch-image --storage foo -a' | grep -Eq -- '\s(--storage|-\w*s)'; echo $?
0
$ echo 'ch-image -b foo -a' | grep -Eq -- '\s(--storage|-\w*s)'; echo $?
1
```

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+    #local -n words=$1
+    local wrds=("$@")
+    local ct=1
+
+    #echo "len: ${#wrds[@]}"
+
+    while (($ct < ${#wrds[@]})); do
+        if [[ ( -f "${wrds[$ct]}" && ( "${wrds[$ct]}" == *".sqfs" || "${wrds[$ct]}" == *".tar."* || "${wrds[$ct]}" == *".tgz" ) ) || ( -d "${wrds[$ct]}" && "${wrds[$ct-1]}" != "--mount" &&
+                "${wrds[$ct-1]}" != "-m" && "${wrds[$ct-1]}" != "--bind" &&
+                "${wrds[$ct-1]}" != "-b" ) ]]; then
```
reidpr, line 309: This conditional is indeed quite rough. Maybe `case` could help? Or temporary variables instead of `${wrds[$ct]}` and `${wrds[$ct-1]}`?

In either case, I think you need a comment explaining what's going on here. It's something about files with certain extensions, or directories, unless the thing is preceded by `--bind` or `--mount`?

Or maybe you can leverage context better? There's only so many places in a `ch-run` command that there can be an image.

Indentation can help too (and you don't need quotes inside `[[ ]]`):
```suggestion
        if [[    (    -f ${wrds[$ct]} \
                   && (    ${wrds[$ct]} == *.sqfs \
                        || ${wrds[$ct]} == *.tar.? \
                        || ${wrds[$ct]} == *.tar.?? \
                        || ${wrds[$ct]} == *.tgz ) ) \
              || (    -d ${wrds[$ct]} \
                   && ${wrds[$ct-1]} != --mount \
                   && ${wrds[$ct-1]} != -m \
                   && ${wrds[$ct-1]} != --bind \
                   && ${wrds[$ct-1]} != -b ) ]]; then
```

There is not an agreed-on extension for SquashFS archives. So identifying those by extension is going to be very error-prone. Maybe `*.*sq*` (any extension containing `sq`)???

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+            echo "${wrds[$ct]}"
+        fi
+        # FIXME: Check for image refs
+        ((ct++))
+    done
+}
+
+# Kludge up a way to look through array of words and determine the
+# subcommand. Note that the double for loop doesn't take that much
+# time, since the Charliecloud command line is relatively short.
```
reidpr, line 319:
```suggestion
# Print the subcommand in an array of words; if there is not one,
# print FIXME. Note that the double for loop doesn't take that much
# time, since the Charliecloud command line is relatively short.
#
# FIXME: This is a kludge because FIXME
```

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+#   >> _ch_subcommand_get "build build-cache ... undelete" \
+#                         "ch-image --foo build ..."
+#      build
+_ch_subcommand_get () {
+    local cmd subcmd
+    local cmds="$1"
+    shift 1
+    for word in "$@"
+    do
+        echo "word: $word" >> /tmp/ch-completion.log
```
reidpr, line 334: Probably you want to gate the debug stuff on some environment variable, e.g. `CH_COMPLETION_DEBUG` (and just leave it in there forever).

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+        for cmd in $cmds
+        do
+            if [[ "$word" == "$cmd" ]]; then
+                subcmd="$cmd"
```
reidpr, line 338: You can return immediately once you know the subcommand, which will save a little time.

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+
+# Kludge up a way to look through array of words and determine the
+# subcommand. Note that the double for loop doesn't take that much
+# time, since the Charliecloud command line is relatively short.
+#
+# Usage:
+#   _ch_subcommand_get [subcommands] [words]
+#
+# Example:
+#   >> _ch_subcommand_get "build build-cache ... undelete" \
```
reidpr, line 325: I would set a global variable containing the subcommands and use that, rather than passing it as an argument.

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+    done
+    echo "$subcmd"
+}
+
+# Code that I shamelessly stole from StackOverflow
+#   (https://stackoverflow.com/a/40227233)
+# Returns filenames and directories, appending a slash to directory names.
+# This function takes option “-X,” a string of space-separated glob patterns
+# to be excluded from file completion using the compgen option of the same
+# name (see https://devdocs.io/bash/programmable-completion-builtins#index-compgen)
```
reidpr, line 350: Use imperative case.

As for “shamelessly stole”, I’d just put the link at the end of the comment.

bin/ch-completion.bash
```diff
@@ -0,0 +1,388 @@
...
+    # directories and the list of files respectively produced by
+    # compgen. The compgen statements also prepend (-P) a “^” and
+    # append (-S) a “$” to the file/dir names to avoid the case
+    # where a substring matching a dirname is erroniously removed
+    # from a filename by the inverted match. These delimiters are
+    # then removed by the “sed”. (See the StackOverflow post cited
+    # above for OP’s explanation of this code). The for loop iterates
+    # through exclusion patterns specified by the “-X” option. If
+    # “-X” isn't specified, the code in the loop executes once,
+    # with no patterns excluded (“-X ""”).
```
reidpr, line 376: Wrap comments to something like 78 or 79 characters. This is too narrow.

reidpr (changes requested):
Great work!

bin/ch-completion.bash
lucaudill: My testing suggests it doesn't persist

bin/ch-completion.bash
lucaudill: I worry about the user setting `CH_IMAGE_STORAGE` after sourcing the script, which would lead to suggestions from the wrong storage directory.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
+#!/bin/bash
+
+# shellcheck disable=SC2012
+# shellcheck disable=SC2207
```
reidpr, line 4: Comment what these are and why we're disabling them.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
+#!/bin/bash
```
reidpr, line 1: Not a script (it's intended to be sourced), so no shebang line.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+#   $ echo ${foo[@]:1:3}
+#   b c d
+#
+
+# Possible extensions once this is merged:
+#   * Add completion support for non-bash shells (e.g. zsh and tchs).
+#     see https://github.com/git/git/tree/master/contrib/completion
+#   * Add support for mid-line completion (possibly using COMP_POINT,
+#     see https://devmanual.gentoo.org/tasks-reference/completion/index.html).
+#
```
reidpr, line 77: Move these to issues.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+
+# Minimum supported bash version. According to
+# http://mywiki.wooledge.org/BashFAQ/061 and my own testing, negative array
+# indexing was introduced in this version. To simplify confusing syntax, we this
+# script is unsupported for bash < 4.2.0.
+bash_vmin="4.2.0"
+
+# Check bash version
+bash_v=$(bash --version | head -1 | grep -Eo "[0-9\.]{2,}[0-9]")
+if [[ $(printf "%s\n%s\n" "$bash_vmin" "$bash_v" | sort -V) < "$bash_vmin" ]]; then
```
reidpr, line 86: This should be comparing for string equality or inequality. I don't know how `<` works for strings in Bash, but it's definitely not comparing version numbers.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+    echo "ch-completion.bash: unsupported bash version ($bash_v < $bash_vmin)"
+    return
+fi
+
+# Debugging log
+if [[ -f "/tmp/ch-completion.log" && -n "$CH_COMPLETION_DEBUG" ]]; then
+    rm /tmp/ch-completion.log
+    printf "completion log\n\n" >> /tmp/ch-completion.log
```
reidpr, line 94: You probably want to always append so that people can compare results from different sourcings.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+    strg_dir=$(_ch_find_storage "${words[@]::${#words[@]}-1}")
+
+    # Populate debug log
+    if [[ -n "$CH_COMPLETION_DEBUG" ]]; then
+        # shellcheck disable=SC2129
+        echo "\$ ${words[*]}" >> /tmp/ch-completion.log
+        echo "  current: $cur" >> /tmp/ch-completion.log
+        echo "  previous: $prev" >> /tmp/ch-completion.log
+        echo "  sub command: $sub_cmd" >> /tmp/ch-completion.log
+    fi
```
reidpr, line 147: Probably best to encapsulate the `$CH_COMPLETION_DEBUG` test in a logging function.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+        ;;
+    esac
+
+    case "$sub_cmd" in
+    build )
+        # FIXME: opts with args that I need to revisit for this logic
+        #
+        #   * --build-arg KEY[=VALUE]
+        #   * --force-cmd=CMD,ARG1[,ARG2...]
+        #
```
reidpr, line 180: Make an issue for this too.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+        case "$prev" in
+        # Go through a list of potential subcommand-specific opts to see if
+        # “$cur” should be an argument. Otherwise, default to CONTEXT or any
+        # valid option (common or subcommand-specific).
+        -f|--file )
```
reidpr, line 185: I've not seen the space before the paren before. I suspect it's not idiomatic???

```suggestion
        -f|--file)
```

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+            compopt -o nospace
+            COMPREPLY=( $(_compgen_filepaths "$cur") )
+            return 0
+            ;;
+        -t )
+            # We can't autocomplete a tag, so we're not even gonna allow
```
reidpr, line 191: Use curly apostrophes.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+            return 0
+            ;;
+        *)
+            # Autocomplete to context directory, common opt, or build-specific opt
+            # --force can take “fakeroot” or “seccomp” as an argument, or no
+            # argument at all. To account for this, we add those two arguments
+            # to the list of compgen suggestions, which allows compgen to
+            # autocomplete to “fakeroot,” “seccomp,” or anything that could
+            # logically follow “--force” with no argument.
+            if [[ "$prev" == "--force" ]]; then
```
reidpr, line 200: No double quotes inside `[[ ]]` except for some special situations that ShellCheck will yell at you about.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+        ;;
+    delete | list )
+        if [[ "$sub_cmd" == "list" ]]; then
+            extras+="$extras -l --long"
+        fi
+        # The following check seems to protects against trying to ls a
+        # non-existent directory, and fixes a bug where the completion function
+        # initialzes an empty storage directory.
+        if [[ -f "$strg_dir" && -n "$(ls "$strg_dir/img")" ]]; then
+            COMPREPLY=( $(compgen -W "$(ls "$strg_dir/img" | sed 's/+/:/g' | sed 's/%/\//g') $extras" -- "$cur") )
```
reidpr, line 228: 1. https://unix.stackexchange.com/questions/128985
2. Use `sed -E`
3. You can give multiple expressions to one `sed`.
4. Suggest using a delimiter other than `/` to avoid the need for escaping, e.g. `s|%|/|g`.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+            compopt -o nospace
+        fi
+        ;;
+    push )
+        if [[ "$prev" == "--image" ]]; then
+            compopt -o nospace
+            COMPREPLY=( $(compgen -d -S / -- "$cur") )
+            return 0
+        # The following check seems to fix a bug where the completion function
+        # initialzes an empty storage directory.
```
reidpr, line 249: This is the second time this comment appears. I think I would instead put a general comment at the top saying you don't run `ch-image list` or any other commands that have side effects like initializing the storage directory.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+        elif [[ -n "$(ls "$strg_dir"/img)" ]]; then
+            COMPREPLY=( $(compgen -W "$(ls "$strg_dir/img" | sed 's/+/:/g' | sed 's/%/\//g') --image" -- "$cur") )
```
reidpr, line 251: Use a function rather than calling `sed` multiple times with the same arguments, to be DRY.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+        ;;
+    '' )
+        # Only autocomplete subcommands if there's no subcommand present.
+        COMPREPLY=( $(compgen -W "$_image_subcommands" -- "$cur") )
+        ;;
+    esac
+
+    # If we've made it this far, the last remaining option for completion is
+    # common opts. Note that we do the “-n” check to avoid being overzealous
+    # with our suggestions.
```
reidpr, line 267: Overzealous how?

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+}
+
+## Helper functions ##
+
+# Disable completion.
+ch-completion-disable () {
+    complete -r ch-image
+}
+
+# Figure out which storage directory to use (including cli-specified storage).
```
reidpr, line 289: Add comment to explain why this isn't done once when the script is sourced.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+        echo "$CH_IMAGE_STORAGE"
+    else
+        echo "/var/tmp/$USER.ch/"
+    fi
+}
+
+# Horrible, disgusting function to find an image or image ref in the ch-run
+# command line.
+#
+# NOT FINISHED, DON'T USE!!!
```
reidpr, line 304: Seems like it's not actually used, so remove.

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+# Usage: _ch_image_subcmd_get [words]
+#
+# Example:
+#   >> _ch_image_subcmd_get "ch-image [...] build [...]"
+#      build
+_ch_image_subcmd_get () {
+    local cmd
+    local subcmd=
+    for word in "$@"
+    do
```
reidpr, line 345: Style for our code:

```suggestion
    for word in "$@"; do
```

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+# the Charliecloud command line, even in the wost case, is relatively short.
+#
+# Usage: _ch_image_subcmd_get [words]
+#
+# Example:
+#   >> _ch_image_subcmd_get "ch-image [...] build [...]"
+#      build
+_ch_image_subcmd_get () {
+    local cmd
+    local subcmd=
```
reidpr, line 343:
```suggestion
    local subcmd
```

bin/ch-completion.bash
```diff
@@ -0,0 +1,403 @@
...
+    for word in "$@"
+    do
+        for cmd in $_image_subcommands
+        do
+            if [[ "$word" == "$cmd" ]]; then
+                subcmd="$cmd"
```
reidpr, line 349: Don't `cmd` and `subcmd` both refer to a subcommand? If so, the naming of `cmd` is confusing.

reidpr (changes requested):
Fantastic work! See comments in-line.

reidpr (approved):

PR #1619, opened 2023-04-03
packaging: add 0.32-2 fedora updates

j-ogas:
0.32 package updates are pending for f39, f37, el7, el8, and el9. See: https://bodhi.fedoraproject.org/users/jogas.

Changes include:
- version 0.32
- epel9 package
- tidy to el7 patch
- `no-rpath.patch`: remove squashfuse lib rpath. Fedora installs the squashfuse libraries in `/usr/lib`; their `check-rpath` linter errors stating that `/usr/lib` is a standard path and shouldn't be rpathed.
- new fedora/rhel conditionals for squashfuse auto-mounting
- new epel conditionals for git


j-ogas:
`0.36-2` pushed to F40, F39, and epel9. 

epel8 and epel7 have been unpushed, each requiring spec updates.

- epel7: we no longer support bats 0.4.0; remove `test` package. 
- epel8: the squashfuse version is 0.1.04; squashfuse auto-mount dependencies are unnecessary.

Both the above will be addressed in a subsequent release and push, e.g., `0.36-3`, based on the diff (`fedora/packaging/upstream.diff`)

PR #1623, opened 2023-04-04
Dockerfile.libfabric: don't build psm3 on aarch64

j-ogas:


examples/Dockerfile.libfabric
```diff
@@ -109,18 +109,24 @@ RUN dnf install -y --setopt=install_weak_deps=false \
...
+# PSM3 is our preferred provider for OPA, however, it requires libpsm2, which
+# is x86_64 only. Thus, the 'PSM_CONFIG' variable is used to avoid building PSM3
+# on aarch64 machines.
 ARG LIBFABRIC_VERSION=1.15.1
 RUN git clone --branch v${LIBFABRIC_VERSION} --depth 1 \
               https://github.com/ofiwg/libfabric/ \
  && cd libfabric \
  && ./autogen.sh \
+ &&    PSM_CONFIG='enable' && [[ "$(uname -m)" == 'x86_64' ]] \
+    || PSM_CONFIG='disable' \
```
reidpr, line 122: This is a bit confusing to me, and you can remove many of the quotes. How about:
```suggestion
 && if [[ $(uname -m) == x86_64 ]]; then \
        PSM_CONFIG=enable; \
    else \
        PSM_CONFIG=disable; \
    fi
```

reidpr (changes requested):
Good progress.

The internal CI failure looks spurious — I think you maybe said that already?

examples/Dockerfile.libfabric
j-ogas: That is pretty ugly lol. Let me try a two-liner?

reidpr (approved):

PR #1628, opened 2023-04-08
`Dockerfile.metadata`: `LABEL`: remove underscores

grosa1:
Hi!
The Dockerfile placed at "test/Dockerfile.metadata" contains the best practice violation [DL3048](https://github.com/hadolint/hadolint/wiki/DL3048) detected by the [hadolint](https://github.com/hadolint/hadolint) tool.

The smell DL3048 occurs when the pattern used for the label keys does not match the format recommended by official guidelines.
In this pull request, we propose a fix for that smell generated by our fixing tool. We have verified that the patch is correct before opening the pull request.
To fix this smell, specifically, the label keys are refactored to match the correct format.

This change is only aimed at fixing that specific smell. If the fix is not valid or useful, please briefly indicate the reason and suggestions for possible improvements.

Thanks in advance.

### DIFF
--- a/test/Dockerfile.metadata
+++ b/test/Dockerfile.metadata
@@ -23,8 +23,7 @@ ENTRYPOINT ["/bin/echo","foo"]
 ENV ch_foo=foo-ev ch_bar=bar-ev
 EXPOSE 867 5309/udp
 HEALTHCHECK --interval=60s --timeout=5s CMD ["/bin/true"]
-LABEL ch-foo=foo-label \
-    ch-bar=bar-label
+LABEL ch-foo=foo-label ch-bar=bar-label
 MAINTAINER charlie@example.com
 ONBUILD RUN echo hello
 RUN echo hello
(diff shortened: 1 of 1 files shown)


reidpr:
Hello @grosa1, thanks for the contribution. It looks like the actual problem is the underscores in `ch_foo` and `ch_bar`, not the line breaking? I think we can probably merge this if we keep it on one line.

In general, we're open to other style fixes, but it might be worth opening a discussion about what exactly to do. We've not seen hadolint before, and it's likely we don't agree on some of the recommendations. That said, I'm a big fan of linters (see `10_sanity.bats:'lint shell scripts'` for an example) and it seems this caught a real problem.

Also, we do have a lot of Dockerfiles that test edge cases, which I assume will cause a lot of false positives.

@BenStormer, I wonder if we should warn/error on labels that don't conform to the linked guidelines?

Thanks again!

BenStormer (approved):

PR #1631, opened 2023-04-13
Fix `LC_ALL=C` crash for older python versions

### LINKED ISSUE #1629: `ch-image --help` crashes with `LC_ALL=C` under Python 3.6
For example:

```
$ python3 --version
Python 3.6.8
$ LC_ALL=C bin/ch-image --help
  [...]
  File "/usr/lib64/python3.6/argparse.py", line 2372, in _print_message
    file.write(message)
UnicodeEncodeError: 'ascii' codec can't encode character '\u201c' in position 1792: ordinal not in range(128)
```

This was printed inside `argparse`. Messages we log don't crash but get escaped in an ugly way:

```
$ LC_ALL=en_US.UTF-8 bin/ch-image list
crash “me”
$ LC_ALL=C bin/ch-image list
crash \u201cme\u201d
```

It doesn't seem to happen under Python 3.9. I haven't tried to narrow down the version boundary.

lucaudill:
closes #1629

### DIFF
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -7,6 +7,7 @@ import datetime
 import enum
 import hashlib
 import io
+import locale
 import os
 import platform
 import pstats
@@ -712,12 +713,14 @@ def monkey_write_streams():
       f.write = write_monkey
    # Try to encode test string of problematic characters. If unsuccessful,
    # monkey patch them out.
-   try:
-      "“”’".encode(encoding=sys.stdout.encoding)
-      "“”’".encode(encoding=sys.stderr.encoding)
-   except UnicodeEncodeError:
-      monkey_write_insert(sys.stdout)
-      monkey_write_insert(sys.stderr)
+   for stream in sys.stdout, sys.stderr:
+      for encoding in stream.encoding, locale.getpreferredencoding().lower(), "ASCII":
+         try:
+               "“”’".encode(encoding=encoding)
+         except UnicodeEncodeError:
+            monkey_write_insert(sys.stdout)
+            monkey_write_insert(sys.stderr)
+            break
 
 def now_utc_iso8601():
    return datetime.datetime.utcnow().isoformat(timespec="seconds") + "Z"
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -710,7 +710,11 @@ def monkey_write_streams():
          text = text.replace("“", "\"").replace("”", "\"").replace("’", "'")
          write_orig(text)
       f.write = write_monkey
-   if (sys.version_info[:2] <= (3, 6)):
+   # Try to encode test string of problematic characters. If unsuccessful,
+   # monkey patch them out.
+   try:
+      "“”’".encode(encoding="ascii")
+   except UnicodeEncodeError:
       monkey_write_insert(sys.stdout)
       monkey_write_insert(sys.stderr)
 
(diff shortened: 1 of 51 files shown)


lib/charliecloud.py
```diff
@@ -701,6 +701,19 @@ def log(msg, hint, trace, color, prefix, end="\n"):
...
+      streams when running Python 3.6. (see #1629)."""
+   def monkey_write_insert(f):
+      write_orig = f.write
+      def write_monkey(text):
+         text = text.replace("“", "\"").replace("”", "\"").replace("’", "'")
+         write_orig(text)
+      f.write = write_monkey
+   if (sys.version_info[:2] <= (3, 6)):
+      monkey_write_insert(sys.stdout)
+      monkey_write_insert(sys.stderr)
```
reidpr, line 715: We'll want to also trigger the fix on the character set not being supported somehow; perhaps locale? terminal something? 3.11 seems to know that the terminal supports Unicode characters despite the locale being `C`.

It's probably possible to directly test if the characters can be encoded correctly. I assume lots of encodings have the curly quotes, not just UTF-8. You could maybe get the default encoding somehow, try to encode a test string, and install the workaround if that throws an exception. This would have the advantage of evaluating results rather than trying to guess what work would based on terminal/locale/etc.

reidpr (changes requested):
Good work. I think we want additional conditionals, though; see in-line.

This one also definitely needs passing internal CI.

reidpr (changes requested):
Good progress but one important gap. Also, please ensure successful internal CI before re-review.

lib/charliecloud.py
```diff
@@ -701,6 +701,24 @@ def log(msg, hint, trace, color, prefix, end="\n"):
...
+      write_orig = f.write
+      def write_monkey(text):
+         text = text.replace("“", "\"").replace("”", "\"").replace("’", "'")
+         write_orig(text)
+      f.write = write_monkey
+   # Try to encode test string of problematic characters. If unsuccessful,
+   # monkey patch them out.
+   try:
+      "“”’".encode(encoding=sys.stdout.encoding)
+      "“”’".encode(encoding=sys.stderr.encoding)
```
reidpr, line 717: I think it's possible for these `encoding` attributes to be None, in which case this code needs a fallback.

reidpr (changes requested):
Good progress! More special cases to deal with, though, unfortunately.

lib/charliecloud.py
```diff
@@ -701,6 +702,26 @@ def log(msg, hint, trace, color, prefix, end="\n"):
...
+   # Try to encode test string of problematic characters. If unsuccessful,
+   # monkey patch them out.
+   for stream in sys.stdout, sys.stderr:
+      for encoding in stream.encoding, locale.getpreferredencoding().lower(), "ASCII":
+         try:
+               "“”’".encode(encoding=encoding)
+         except UnicodeEncodeError:
+            monkey_write_insert(sys.stdout)
+            monkey_write_insert(sys.stderr)
+            break
```
reidpr, line 723: There's some subtle errors in the loop handling here, I think. I'd recommend something like (untested):

```suggestion
   for stream in sys.stdout, sys.stderr:
      for encoding in stream.encoding, locale.getpreferredencoding(), "ASCII":
         if (encoding is not None):
            try:
               "“”’".encode(encoding=encoding)
            except UnicodeEncodeError:
               monkey_write_insert(stream)
               break
```

reidpr (changes requested):
Getting closer!

reidpr (approved):

PR #1636, opened 2023-04-14
add `-q`/`--quiet` to `ch-image`

### LINKED ISSUE #1613: add `-q` / `--quiet`
As suggested in #1603: Add to all executables `-q` / `--quiet` options to reduce the log level to `WARNING`.

lucaudill:
closes #1613


doc/ch-image.rst
```diff
@@ -93,13 +96,52 @@ Common options placed before or after the sub-command:
...
+        - ❌
+        - ❌
+      * - Subprocess stderr
+        - ✅
+        - ✅
+        - ❌
+      * - Subprocess stdout
+        - ✅
+        - ❌
+        - ❌
```
reidpr, line 139: Couple of suggestions on this table:

1. Extend it to also include the default and the three `-v` levels (i.e., 7 columns total), and move it to its own section.
2. Order the rows from most suppressed to least suppressed.

lib/charliecloud.py
```diff
@@ -411,10 +411,12 @@ def DEBUG(msg, hint=None, **kwargs):
    if (verbose >= 2):
       log(msg, hint, None, "38;5;6m", "", **kwargs)  # dark cyan (same as 36m)
 
-def ERROR(msg, hint=None, trace=None, **kwargs):
-   log(msg, hint, trace, "1;31m", "error: ", **kwargs)  # bold red
+def ERROR(msg, hint=None, trace=None, fatal=False, **kwargs):
+   if (log_quiet < 3 or fatal):
```
reidpr, line 415: The `fatal` parameter seems awkward. What if we treat fatal errors and non-fatal errors (of which there are few) the same? Alternately, when we catch `Fatal_Error` we could just call `log()` manually.

lib/charliecloud.py
```diff
@@ -427,7 +429,8 @@ def FATAL(msg, hint=None, **kwargs):
 
 def INFO(msg, hint=None, **kwargs):
    "Note: Use print() for output; this function is for logging."
-   log(msg, hint, None, "33m", "", **kwargs)  # yellow
+   if (log_quiet == 0):
```
reidpr, line 432: I think things would be simpler if there was a single verbosity variable. Quiet could be negative values. Many programs call this “log level” ([e.g.](https://www.javatpoint.com/log4j-logging-levels)).

Also I don't recall whether there's an enum for it, but there probably should be, to avoid magic numbers. Then it would read something like:

```suggestion
   if (verbose >= LOG_INFO):
```

lib/charliecloud.py
```diff
@@ -481,6 +485,10 @@ def cmd(argv, fail_ok=False, **kwargs):
    """Run command using cmd_base(). If fail_ok, return the exit code whether
       or not the process succeeded; otherwise, return (zero) only if the
       process succeeded and exit with fatal error if it failed."""
+   if (log_quiet > 1):
+      kwargs["stdout"] = subprocess.DEVNULL
+   if (log_quiet == 3):
+      kwargs["stderr"] = subprocess.DEVNULL
```
reidpr, line 491: I think this will be mooted by other changes, but this will be more clear if the two conditions use the same comparison, e.g.:
```suggestion
   if (log_quiet > 1):
      kwargs["stdout"] = subprocess.DEVNULL
      if (log_quiet > 2):
         kwargs["stderr"] = subprocess.DEVNULL
```

lib/misc.py
```diff
@@ -86,6 +86,16 @@ def gestalt_python_path(cli):
...
 
+def gestalt_logging(cli):
+   ch.INFO("info")
+   ch.VERBOSE("verbose")
+   ch.DEBUG("debug")
+   ch.TRACE("trace")
+   ch.WARNING("warning")
+   ch.ERROR("error")
+   if (cli.fail):
+      ch.FATAL("the program failed inexplicably")
```
reidpr, line 97:
```suggestion
      ch.FATAL("tried to believe gecgecgec is music")
```

lib/charliecloud.py
```diff
@@ -599,10 +608,21 @@ def exit(code):
...
-   global log_festoon, log_fp, trace_fatal, verbose
+   global log_festoon, log_fp, log_quiet, trace_fatal, verbose
+   log_quiet = cli.quiet
    assert (0 <= cli.verbose <= 3)
    verbose = cli.verbose
    trace_fatal = (cli.debug or bool(os.environ.get("CH_IMAGE_DEBUG", False)))
+   if (log_quiet > 0):
+      fail_ct = 0
+      if (trace_fatal):
+         ERROR("“quiet” incompatible with “debug”")
```
reidpr, line 619: Might want to mention the environment variable here too since folks might not make the translation or remember they set it.

test/build/50_ch-image.bats
```diff
@@ -41,6 +41,53 @@ EOF
...
+    [[ $output = *'warning: warning'* ]]
+    [[ $output = *'error: error'* ]]
+
+    # quiet level 1
+    run ch-image gestalt -q logging
+    echo "$output"
+    [[ $status -eq 0 ]]
+    [[ $output != *"info"* ]]
+    [[ $output = *'warning: warning'* ]]
+    [[ $output = *'error: error'* ]]
```
reidpr, line 62: Also test logging with quiet level 2.

test/build/50_ch-image.bats
```diff
@@ -41,6 +41,53 @@ EOF
...
+
+    # quiet level 2
+    run ch-image build --force seccomp -t tmpimg -qq -f - . << 'EOF'
+FROM almalinux:8
+RUN dnf install -y --setopt=install_weak_deps=false openssh-clients \
+ && dnf clean all
```
reidpr, line 68: Use explicit redirection on toy commands like `echo` rather than a big messy command like `dnf`. E.g.:
```suggestion
FROM alpine:3.17
RUN echo 'this is stdout'
RUN echo 'this is stderr' 1>&2
```

lib/misc.py
```diff
@@ -86,6 +86,16 @@ def gestalt_python_path(cli):
 def gestalt_storage_path(cli):
    print(ch.storage.root)
 
+def gestalt_logging(cli):
+   ch.INFO("info")
+   ch.VERBOSE("verbose")
+   ch.DEBUG("debug")
+   ch.TRACE("trace")
+   ch.WARNING("warning")
+   ch.ERROR("error")
```
reidpr, line 95: Put these in order.

reidpr (changes requested):
Great work! See suggestions in-line.

doc/ch-image.rst
```diff
@@ -98,13 +101,15 @@ Common options placed before or after the sub-command:
     subprocesses. Profile data should still be written on fatal errors, but
     not if the program crashes.
 
+  :code:`-q, --quiet`
+    Be quieter; can be repeated. Incompatible :code:`-v` and suppresses
```
reidpr, line 105:
```suggestion
    Be quieter; can be repeated. Incompatible with :code:`-v` and suppresses
```

doc/ch-image.rst
```diff
@@ -98,13 +101,15 @@ Common options placed before or after the sub-command:
...
+    :code:`--debug` regardless of option order. See the :ref:`FAQ entry on
```
reidpr, line 106: I wonder if we should revisit the `--debug` incompatibility? It seems orthogonal now. 

Also `--debug` seems to be undocumented here.

doc/faq.rst
```diff
@@ -1098,6 +1094,153 @@ linuxcontainers.org uses the opposite order for “le” in the architecture nam
...
+     “fatal” and “error” levels, but this isn’t really meaningful to users.)
+
+  2. **Warning**. Unexpected condition the user needs to know about but that
+     should not stop the program. Examples: :code:`ch-run --mount` with a
+     directory image (which does not use a mount point), unsupported
+     Dockerfile instructions that are ignored.
+
+  3. **Info**. Chatter useful enough to be printed by default. Example:
+     progress messages during image download and unpacking. (:code:`ch-run` is
+     silent during normal operations and does not have any “info” logging)
```
reidpr, line 1116:
```suggestion
     silent during normal operations and does not have any “info” logging.)
```

doc/faq.rst
```diff
@@ -1098,6 +1094,153 @@ linuxcontainers.org uses the opposite order for “le” in the architecture nam
   $ wget https://uk.lxd.images.canonical.com/images/alpine/3.15/ppc64el/default/20220304_13:00/rootfs.tar.xz
   $ ch-image import rootfs.tar.xz ppc64le/alpine:3.17
 
+.. _faq_verbosity:
+
+How can I control Charliecloud’s quietness or verbosity?
```
reidpr, line 1099: Say something about Charliecloud's own stdout in this section?

lib/charliecloud.py
```diff
@@ -50,6 +51,24 @@ class Force_Mode(enum.Enum):
...
+# Log level
+@functools.total_ordering
+class Log_Level(enum.Enum):
+   TRACE = 3
+   DEBUG = 2
+   VERBOSE = 1
+   INFO = 0
+   QUIET_INFO = -1
+   QUIET_WARNING = -2
+   QUIET_STDERR = -3
```
reidpr, line 63: More idiomatic would be something like:

```suggestion
   WARNING = -1
   FATAL = -2
   FATAL_SILENCE = -3
```

lib/charliecloud.py
```diff
@@ -543,7 +568,8 @@ def cmd_stdout(argv, encoding="UTF-8", **kwargs):
       CompletedProcess object (its stdout is available in the “stdout”
       attribute). If logging is debug or higher, print stdout."""
    cp = cmd_base(argv, encoding=encoding, stdout=subprocess.PIPE, **kwargs)
-   if (verbose >= 2):  # debug or higher
+   # FIXME: Add a “log_quiet” check here!
```
reidpr, line 571: Is this comment left over?

lib/charliecloud.py
```diff
@@ -543,7 +568,8 @@ def cmd_stdout(argv, encoding="UTF-8", **kwargs):
...
+   if (log_level >= Log_Level.DEBUG):  # debug or higher
```
reidpr, line 572: Delete these comments since you put in a nice enum! ;)

lib/charliecloud.py
```diff
@@ -613,20 +639,32 @@ def exit(code):
...
    save_xattrs = (not cli.no_xattrs)
    trace_fatal = (cli.debug or bool(os.environ.get("CH_IMAGE_DEBUG", False)))
+   if (cli.quiet):
+      fail_ct = 0
+      if (trace_fatal):
+         ERROR("“quiet” incompatible with “debug” and “CH_IMAGE_DEBUG”")
+         fail_ct += 1
+      if (cli.verbose):
+         ERROR("“quiet” incompatible with “verbose”")
+         fail_ct += 1
```
reidpr, line 652: We could do it where every `-v` increases verbosity by 1 and every `-q` decreases it by 1. I wonder if that's too weird though.

lib/charliecloud.py
```diff
@@ -613,20 +639,32 @@ def exit(code):
...
+   global log_festoon, log_fp, log_level, trace_fatal, save_xattrs
    save_xattrs = (not cli.no_xattrs)
    trace_fatal = (cli.debug or bool(os.environ.get("CH_IMAGE_DEBUG", False)))
+   if (cli.quiet):
+      fail_ct = 0
+      if (trace_fatal):
+         ERROR("“quiet” incompatible with “debug” and “CH_IMAGE_DEBUG”")
+         fail_ct += 1
+      if (cli.verbose):
+         ERROR("“quiet” incompatible with “verbose”")
```
reidpr, line 651:
```suggestion
         ERROR("--quiet incompatible with --verbose")
```

lib/charliecloud.py
```diff
@@ -613,20 +639,32 @@ def exit(code):
...
    if ("CH_LOG_FESTOON" in os.environ):
       log_festoon = True
    file_ = os.getenv("CH_LOG_FILE")
    if (file_ is not None):
-      verbose = max(verbose, 1)
       log_fp = file_.open_("at")
    atexit.register(color_reset, log_fp)
    VERBOSE("version: %s" % version.VERSION)
-   VERBOSE("verbose level: %d" % verbose)
+   VERBOSE("verbose level: %d" % log_level.value)
```
reidpr, line 667: Would be better if the enum could report itself as a string too?

test/build/50_ch-image.bats
```diff
@@ -41,6 +41,52 @@ EOF
...
+    [[ $status -eq 0 ]]
+    [[ $output != *"info"* ]]
+    [[ $output = *'warning: warning'* ]]
+    [[ $output = *'error: error'* ]]
+
+    # quiet level 2
+    run ch-image build --force seccomp -t tmpimg -qq -f - . << 'EOF'
+FROM alpine:3.17
+RUN echo 'this is stdout'
+RUN echo 'this is stderr' 1>&2
```
reidpr, line 68: Would be good to test the output of these two `echo`s against the levels when they should and should not appear.

reidpr (changes requested):
Very nice! Several suggestions in-line.

reidpr (approved):

PR #1642, opened 2023-04-25
`debian:stretch` fix

lucaudill:
As indicated in [this thread](https://stackoverflow.com/questions/76094428/debian-stretch-repositories-404-not-found), `debian:stretch` has been moved to archive.debian.org (see: https://lists.debian.org/debian-devel-announce/2023/03/msg00006.html), which is causing our test suite to fail. [This comment](https://stackoverflow.com/a/76095392) suggests a fix that can be implemented inside of a Dockerfile, which is perfect for us. An alternative approach would be to update to a newer version of Debian, since [LTS for stretch has been dropped](https://wiki.debian.org/LTS) as of summer 2022.


reidpr (approved):
Fantastic, thanks for catching this!

PR #1643, opened 2023-04-26
tutorial: `ch-convert` instead of `ch-image convert`

jsta:
Solves an error when working through the tutorial:

> "ch-image: error: argument CMD: invalid choice: 'convert'"


reidpr:
Good catch, thank you! Merging.

PR #1654, opened 2023-05-17
`ch-image import`: fix missing metadata error

### LINKED ISSUE #1513: ch-image import: lack of history metadata leads to assertion error when pushing.
Related to #1215 and #1002.

Images in storage derived from `ch-image import` that do not have: a) any history entries in their metadata; or b) a history entry without `empty_value: false`, will trigger an assertion error or traceback due to:

```python
    assert(non_empty_winner is not None)
```
and
```python
if (   "empty_layer" not in hist[i].keys() [...]
```
 respectively.  This can be avoided if `ch-image import` is updated to add a history metadata when imported, e.g.,
 
 ```json
 {
    "history": [
      { "empty_layer": false,
        "command": "ch-image import foo bar"},
    ]
}
 ```

lucaudill:
closes #1513

### DIFF
--- a/lib/misc.py
+++ b/lib/misc.py
@@ -94,8 +94,6 @@ def import_(cli):
       # Un-tag previously deleted branch, if it exists.
       bu.cache.tag_delete(pathstr, fail_ok=True)
    dst = im.Image(im.Reference(cli.image_ref))
-   dst.metadata["history"] = [{ "empty_layer": False,
-                                "command":     "ch-image import"}]
    ch.INFO("importing:    %s" % cli.path)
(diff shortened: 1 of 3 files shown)


reidpr (changes requested):
Fix looks nice and tidy, but should we have a test too? @j-ogas has a reproducer in the bug, and it can probably be tightened up a bit.

lucaudill:
> #1513

Thanks to @j-ogas for the [tidy solution](https://github.com/hpc/charliecloud/issues/1513#issue-1503402662) btw

reidpr (approved):

Issue #1638, opened 2023-04-17
FROM ARG: all cache hits when value of ARG changes


j-ogas:
Reproduce:
```
$ ch-image reset
$ ch-image import alpine.tar.gz alpine
[...]
$ ch-image import debian.tar.gz debian
[...]
$ ch-image build -t foo -f - . <<EOF
arg base=alpine
from $base
run touch foo
EOF
[...]
$ ch-image build --build-arg base=debian -t foo -f - . <<EOF
arg base=alpine
from $base
run touch foo
EOF
  1  ARG base='debian'
  2* FROM debian
  3* run touch foo
[...]
copying image from cache ...
```

`run touch foo` should not hit when the base image is changed.

reidpr:
I suspect this is an error in how the state ID is computed for FROM. I know we take variables into account for things like `RUN`.

Note that we consider variable values for the cache when they are used, not when the are set like Docker. Hence no hit/miss at all for that `ARG`.

reidpr:
Closing as duplicate in favor of #1658, which has a better characterization of the problem.

PR #1658, opened 2023-05-24
use different state ID for different imported image references

reidpr:
Previously, images based on something `ch-image imported` would hit if they matched *any* imported image. E.g.:

```
#!/bin/bash

set -ex

ch-image reset

# must have been previously exported with ch-convert
ch-image import /var/tmp/alpine317.tar.gz alpine:3.17
ch-image import /var/tmp/alpine316.tar.gz alpine:3.16

ch-image build -t foo -f - . <<'EOF'
FROM alpine:3.17
RUN touch foo
EOF

# RUN should miss but it doesn't
ch-image build -t foo -f - . <<'EOF'
FROM alpine:3.16
RUN touch foo
EOF

ch-image build-cache --tree
```

(This was originally reported as issue #1638, but the problem is in the import, not `ARG`.)


doc/ch-image.rst
```diff
@@ -1121,10 +1121,12 @@ If the imported image contains Charliecloud metadata, that will be imported
...
 
-.. note::
+.. warning::
 
-   Every import creates a new cache entry, even if the file or directory has
-   already been imported.
+   Descendant images (i.e., :code:`FROM` the imported :code:`IMAGE_REF`) are
+   linked using :code:`IMAGE_REF` only. If a new image is imported under a new
+   :code:`IMAGE_REF`, all instructions descending from that :code:`IMAGE_REF`
+   will still hit, even if the new image is different.
```
j-ogas, line 1129: Perhaps we should add an issue to reference here for potential future ch-image import improvements?

j-ogas (approved):
LGTM.

PR #1356, opened 2022-05-27
delete RPM’s `__db.*`

reidpr:
Closes #1351.

PR #1662, opened 2023-05-26
fix several CI bugs

reidpr:
This PR fixes multiple bugs that prevent internal and external full-scope tests from passing.

1. Our old-storage storage directories are x86-64, but some versions contain an ARM `alpine:latest` in the cache left over from one test that gets caught by another. Remove it.

2. Fix some typos and word-wrapping in `Dockerfile.mpich`.

3. Don’t restore RPM databases if found in the cache (presumably left over from before PR #1356).

4. Fix indentation error in PR #1631 so it’s applied only when needed, instead of unconditionally.

5. Re-pull some base images (currently only `archlinux:latest`) that break downstream builds if out of date, fixing `old-storage` failures.

### DIFF
--- a/test/old-storage
+++ b/test/old-storage
@@ -100,7 +100,7 @@ for oldtar in $oldtars; do
 
     case ${storage#*-} in
         0.29|0.30|0.31)
-            INFO "working around issue #1662"
+            INFO "working around bug fixed by PR #1662"
             (cd "$storage"/bucache && git branch -D alpine+latest)
             ;;
     esac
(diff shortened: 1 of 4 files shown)


test/old-storage
```diff
@@ -98,8 +98,19 @@ for oldtar in $oldtars; do
     export CH_IMAGE_STORAGE=$storage
     INFO "unpacked: $(du -sh "$storage" | cut -f1) bytes in $(du --inodes -sh "$storage" | cut -f1) inodes"
 
+    case ${storage#*-} in
+        0.29|0.30|0.31)
+            INFO "working around issue #1662"
```
mphinney1100, line 103: isn't #1662 and PR and not an issue. (took me a while to figure that one out). Maybe rename to "working around issue in PR #1662"

PR #1666, opened 2023-06-05
fix mpich example typo

j-ogas:


reidpr:
Good catch. These are also fixed in PR #1662 merged moments ago, so tagging as duplicate.

PR #1677, opened 2023-06-08
reprint warnings before exit (`ch-run`, `ch-image`)

### LINKED ISSUE #1674: re-print warnings before exit
Warnings can get lost in the middle of all the other output. Let’s reprint them before `ch-run` or `ch-image` exits.

lucaudill:
closes #1674


bin/ch-run.c
```diff
@@ -73,6 +74,7 @@ const struct argp_option options[] = {
    { "unset-env",      -7, "GLOB", 0, "unset environment variable(s)" },
    { "verbose",       'v', 0,      0, "be more verbose (can be repeated)" },
    { "version",       'V', 0,      0, "print version and exit" },
+   { "warnings",      -16, "NUM",  0, "log NUM warnings and exit" },
```
reidpr, line 77: Is it possibly to bury this under some debug option? Or maybe we should worry about that when we get more similar stuff?

bin/ch-run.c
```diff
@@ -121,8 +126,14 @@ int main(int argc, char *argv[])
    int arg_next;
    char ** c_argv;
 
+   // initialze “warnings” buffer
```
reidpr, line 129: “initialize”

bin/ch_misc.c
```diff
@@ -41,6 +41,21 @@ char *host_tmp = NULL;
...
+   “warnings_append()” for more details. */
+char *warnings;
+
+/* Current byte offset from start of “warnings” buffer. This gives the address
+   where the next appended string will start. */
+size_t warnings_offset = 0;
+
+/* Size of “warnings” buffer, in bytes. We want this to be big enough that we
+   don’t need to worry about running out of room. */
+const size_t warnings_size = 4096;
```
reidpr, line 57: I think we usually do these constants as preprocessor `#define`s.

bin/ch_misc.c
```diff
@@ -41,6 +41,21 @@ char *host_tmp = NULL;
 /* Username of invoking users. Set during command line processing. */
 char *username = NULL;
 
+/* List of warnings to be re-printed on exit. This is a buffer of shared memory
+   allocated by mmap(2), structured as a sequence of
+   null-terminated character strings. Warnings that do not fit in this buffer
+   will be lost, though we allocate enough memory that this is unlikely. See
+   “warnings_append()” for more details. */
```
reidpr, line 48: Word-wrap looks weird here?

bin/ch_misc.c
```diff
@@ -450,18 +465,19 @@ noreturn void msg_fatal(const char *file, int line, int errno_,
 void msgv(enum log_level level, const char *file, int line, int errno_,
           const char *fmt, va_list ap)
 {
+   char *message = "";
```
reidpr, line 468: Initialize this to `NULL` (or not at all) so it doesn't look like you're trying to write to a static string.

bin/ch_misc.c
```diff
@@ -450,18 +465,19 @@ noreturn void msg_fatal(const char *file, int line, int errno_,
...
       return;
 
-   fprintf(stderr, "%s[%d]: ", program_invocation_short_name, getpid());
+   T_ (1 <= asprintf(&message, "%s[%d]: ", program_invocation_short_name, getpid()));
 
    // Prefix for the more urgent levels.
    switch (level) {
    case LL_FATAL:
-      fprintf(stderr, "error: ");  // "fatal" too morbid for users
+      T_ (1 <= asprintf(&message, "%s%s", message, "error: ")); // "fatal" too morbid for users
```
reidpr, line 477: I believe we have a `cat()` function to do this.

bin/ch_misc.c
```diff
@@ -471,12 +487,22 @@ void msgv(enum log_level level, const char *file, int line, int errno_,
    if (fmt == NULL)
       fmt = "please report this bug";
 
-   vfprintf(stderr, fmt, ap);
-   if (errno_)
-      fprintf(stderr, ": %s (%s:%d %d)\n",
-              strerror(errno_), file, line, errno_);
-   else
-      fprintf(stderr, " (%s:%d)\n", file, line);
+   char *ap_msg = "";
```
reidpr, line 490: I don't like to declare variables in the middle of a scope because they're valid for the entire scope, i.e., also valid in the code above before you get to the declaration.

bin/ch_misc.c
```diff
@@ -640,3 +666,37 @@ void version(void)
 {
    fprintf(stderr, "%s\n", VERSION);
 }
+
+/* Append null-terminated string “str” to the memory buffer “offset” bytes away
```
reidpr, line 670: s/away/after/

bin/ch_misc.c
```diff
@@ -640,3 +666,37 @@ void version(void)
...
+   from the address pointed to by “addr”. Buffer should be “size” bytes long.
```
reidpr, line 671: “Buffer length is ‘size’.”

bin/ch_misc.c
```diff
@@ -640,3 +666,37 @@ void version(void)
...
+   Return the number of bytes written. If there isn’t enough room for the
+   string, do nothing and return zero. */
+size_t warnings_append(char *addr, char *str, size_t size, size_t offset)
```
reidpr, line 674: This is pretty generic, yes? More about strings than warnings? If so name it accordingly; if not, explain why it's not generic.

test/build/50_dockerfile.bats
```diff
@@ -160,6 +160,7 @@ test 7 b
 grown in 16 instructions: tmpimg
 build slow? consider enabling the new build cache
 hint: https://hpc.github.io/charliecloud/command-usage.html#build-cache
+warning: not yet supported, ignored: issue #777: .dockerignore file
```
reidpr, line 163: I wonder if it's useful to give a line saying how many warnings are being re-printed?

bin/ch-run.c
lucaudill: I'm leaning towards waiting until we have more things to group under such an option, but I'd be willing to discuss further.

bin/ch_misc.c
lucaudill: good catch!

test/run/ch-run_misc.bats
```diff
@@ -1054,3 +1054,19 @@ EOF
...
+    [[ $status -eq 0 ]]
+    [[ $(echo "$output" | grep -Fc 'this is a warning!') -eq 0 ]]
+
+    run ch-run --warnings=1
+    [[ $status -eq 0 ]]
+    [[ $(echo "$output" | grep -Fc 'this is a warning!') -eq 2 ]]
+
+    # Warnings list is a statically sized memory buffer. Ensure it works as
+    # intended by printing more warnings than can be saved to this buffer and
+    # checking that the program doesn’t crash.
```
reidpr, line 1069: Also validate the number of warnings re-printed.

reidpr (changes requested):
Great work! See tidy commit, which builds but is mostly untested and will surely fail CI, so fixing that remains.

test/run/ch-run_misc.bats
```diff
@@ -1074,8 +1074,8 @@ EOF
     run ch-run --warnings=100
     echo "$output"
     [[ $status -eq 0 ]]
-    [[ $output == *'ch-run['*']: warning: reprinting first 52 warning(s)'* ]]
-    [[ $(echo "$output" | grep -Fc 'this is warning 52!') -eq 2 ]]
-    [[ $(echo "$output" | grep -Fc 'this is warning 53!') -eq 1 ]]
+    [[ $output == *'ch-run['*']: warning: reprinting first '*' warning(s)'* ]]
+    [[ $(echo "$output" | grep -Fc 'this is warning 1!') -eq 2 ]]
+    [[ $(echo "$output" | grep -Fc 'this is warning 100!') -eq 1 ]]
```
reidpr, line 1079: Clever!

reidpr (approved):

PR #1689, opened 2023-06-19
update tutorial and FAQ

lucaudill:
The tutorial has been stale since adding `seccomp` to `--force`. This PR addresses that.

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -99,6 +99,31 @@ two solutions:
    processes and one writes a file in the image that another is reading or
    writing).
 
+:code:`ch-image` fails with "argument --force: invalid choice"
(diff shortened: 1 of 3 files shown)


reidpr (approved):
Looks good; LMK if you like my tidy commit and if so I'll merge.

lucaudill:
@reidpr I'm tempted to put the admonition a little higher up, just so it's a little more visible.

reidpr:
> @reidpr I like your wording. I'm tempted to put the admonition a little higher up, just so it's a little more visible though.

Sure. My only reasoning was that then we had an admonition followed by the opening paragraph both starting "This section..." which seemed awkward.

PR #1694, opened 2023-07-05
clean up Python code

### LINKED ISSUE #1693: clean up Python code
While using VSCode, I saw that there have been a few persistent (but henceforth unnoticed) issues with our python code. In particular, I saw a number of imports across our python files that were not actually being used. There were also a couple instances where functions from `charliecloud.py` were being used in other files without a reference to their source (e.g. `func()` rather than `ch.func()`), which would throw an error. Apparently we just don't often get to that point in the code, and that's why it's been unnoticed for so long.

lucaudill:
closes #1693

### DIFF
--- a/lib/build.py
+++ b/lib/build.py
@@ -16,7 +16,6 @@ import build_cache as bu
 import filesystem as fs
 import force
 import image as im
-import pull
 
 
 ## Globals ##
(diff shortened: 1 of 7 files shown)


reidpr (changes requested):
Code looks great but can you add a bug description saying what the cleanup was about and why?

reidpr (approved):

PR #1699, opened 2023-07-18
Fix docker check if docker is a wrapper for podman

pagrubel:
This fixes the docker check if there is a docker wrapper for podman.

### DIFF
--- a/lib/base.sh
+++ b/lib/base.sh
@@ -144,11 +144,14 @@ vset () {
     fi
 }
 
-
-# Do we need sudo to run docker?
-# Skip if docker is a wrapper for podman
-if [[ $(docker --help 2> /dev/null) == *"podman"* ]]; then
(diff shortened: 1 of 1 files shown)


pagrubel:
What's the magic to get the tests to pass for this simple PR?

reidpr (commented):
Pat, thanks for the patch. I am sorry to let it sit so long. @mphinney1100 referred the brackets problem to me. Indeed, this caused the test failure. Changing to single brackets made the glob pattern not work any more, so I pushed an alternative. Also I moved the error message to later.

Can you please try the changed PR and let us know if it works for you?

reidpr:
This time the test failure is spurious (see PR #1707).

pagrubel:
Thank you for the fix. It does work on the system that we encountered problems on.

PR #1704, opened 2023-08-15
`squashfuse` update

### LINKED ISSUE #1696: SquashFUSE 0.3.0 is incompatible
[This commit](https://github.com/vasi/squashfuse/commit/53e1b97002c7bc13392a6b0121680d28a94a9715) to `squashfuse` adds a new subdirectory argument to the `sqfs_ll_open` function. This broke CI, since we use the development version of `squashfuse` there and we call `sqfs_ll_open` in `ch_fuse.c`. I'm writing a [temporary fix](https://github.com/hpc/charliecloud/pull/1697) that'll use the last commit before this change in CI, but we should really find a more permanent solution that'll allow us to keep using the most up-to-date version.

lucaudill:
closes #1696

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -154,9 +154,6 @@ jobs:
           cd /usr/local/src
           git clone https://github.com/vasi/squashfuse.git
           cd squashfuse
-          # checkout the last commit that we support. FIXME: find a more
-          # permanent solution (#1696).
-          git checkout 4465c39
           ./autogen.sh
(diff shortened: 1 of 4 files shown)


lucaudill:
The function signature changes that caused the bug were undone in [this commit](https://github.com/vasi/squashfuse/commit/c5f60ed29ba5a6f1ed7d7233f795ee9141e01bf1), so the bug is present in version 0.3.0, but all other releases should work normally. We should document this.

PR #1707, opened 2023-08-18
fix test suite for Docker BuildKit

reidpr:
As of version 23, Docker’s [legacy builder is deprecated](https://docs.docker.com/engine/deprecated/#legacy-builder-for-linux-images) and BuildKit is the default. This breaks a number of assumptions in the test suite, both functional and cosmetic. In recent weeks, the updated Docker has appeared in the GitHub Actions image.

This PR updates the test suite to pass again under Docker.

### DIFF
--- a/examples/copy/Dockerfile
+++ b/examples/copy/Dockerfile
@@ -164,14 +164,16 @@ COPY dirA dir08a
 # Dest:    new directory, one level, trailing slash
 COPY dirA dir08b/
 
-# Source:  one directory
-# Dest:    existing file, 2nd level
+# NOTE: Not currently tested (see #1707).
 #
-# Note: While this fails if the existing file is at the top level (which we
-# verify in test/build/50_dockerfile.bats), if the existing file is at the 2nd
-# level, it's overwritten by the directory.
-RUN touch dir08a/dirCb
-COPY dirCa dir08a
+# # Source:  one directory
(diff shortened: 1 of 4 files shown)


examples/copy/Dockerfile
```diff
@@ -269,37 +271,39 @@ RUN test $(stat -c '%A' dir19/dir19a2 | cut -c8-) != 'rwx' \
...
-        && ln -s file$i dir20/s_file$i \
-        && mkdir dir20/dir$i \
-        && echo dir$i/file_ > dir20/dir$i/file_ \
-        && ln -s dir$i dir20/s_dir$i; \
-    done \
- && ls -lR dir20
-
-# Copy in the new directory tree. In all of these cases, the source simply
-# overwrites the destination; symlinks are not followed.
+# NOTE: Not currently tested (see #1707).
```
lucaudill, line 274: You might want to add some extra context for this comment chunk, particularly the commented out Dockerfile instruction that's still included.

test/build/50_dockerfile.bats
```diff
@@ -590,6 +591,10 @@ EOF
     fi
     prerequisites_ok argenv
 
+    sed_ () {
```
lucaudill, line 594: I think explaining this function in a comment would be good since the regex is a bit intimidating.

PR #1709, opened 2023-08-28
bash completion tweak

lucaudill:
When trying to parse certain information out of the command line (e.g. image references or subcommands), we want to ignore the current word, because it might be incomplete, which would lead to erroneous completion behavior. When I originally wrote the script, I thought it was sufficient to ignore the last word in the command line, neglecting the fact that the last word in the command line isn't necessarily the current word. This PR fixes that, ensuring that the word being ignored is always the current word.


reidpr (approved):
Looks great, thanks for this.

Issue #1722, opened 2023-09-07
Arch Linux tests fail due to key problems


reidpr:
Sometime in early September one/some of the Arch Linux developer keys rolled over and the base images have not been updated. I believe it's supposed to automatically update the key, but that doesn't work. To reproduce:

```
$ cat /tmp/Dockerfile
FROM archlinux:latest
RUN pacman -Syq --noconfirm openssh
$ ch-image build -t tmpimg2 --rebuild -f /tmp/Dockerfile /tmp
[...]
  2. RUN pacman -Syq --noconfirm openssh
copying image from cache ...
[...]
checking keyring...
downloading required keys...
:: Import PGP key 9B7A287D9A2EC608, "David Runge <dvzrv@archlinux.org>"? [Y/n]
error: key "9B7A287D9A2EC608" could not be looked up remotely
error: required key missing from keyring
error: failed to commit transaction (unexpected error)
Errors occurred, no packages were upgraded.
something went wrong, rolling back ...
[...]
error: build failed: RUN command exited with 1
trace: build.py:1180:execute, build.py:250:__default__, visitors.py:241:_call_userfunc, visitors.py:269:visit_topdown, build.py:186:main, ch-image:314:main
```

For now, we'll just disable this test (see linked PR). This issue is to restore it when the problem is fixed; keep an eye on the [Docker Hub images](https://hub.docker.com/_/archlinux/tags) for updates.

reidpr:
I tried this but it didn't work: https://bbs.archlinux.org/viewtopic.php?pid=2119200

reidpr:
Problem seems to no longer happen.

PR #1723, opened 2023-09-07
`--force` tests: disable Arch Linux

reidpr:
Arch Linux `--force` tests are having problems; disable. See #1722 to restore when they fix it.


lucaudill (approved):
LGTM

PR #1724, opened 2023-09-08
rewrite embedded paths at install time

### LINKED ISSUE #683: don't hard-code path to Python and sh libraries

reidpr:
Closes #683.

This PR changes library paths in sh and Python code to be valid in the source directory and removes the `lib/charliecloud` symlink. It then re-writes these paths to be valid for the install location at install time.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -263,7 +263,15 @@ jobs:
           else
               LDD=ldd
           fi
-          ./configure --prefix=$ch_prefix/from-git
+          # Go wild with install directories to make sure it works (see #683).
+          # We do have a normal install below.
+          ./configure --prefix=$ch_prefix/from-git \
+                      --datarootdir=$ch_prefix/from-git_dataroot \
+                      --libexecdir=$ch_prefix/from-git_libexec \
+                      --libdir=$ch_prefix/from-git_lib \
+                      --mandir=$ch_prefix/from-git_man \
+                      --docdir=$ch_prefix/from-git_doc \
+                      --htmldir=$ch_prefix/from-git_html
           set -x
           fgrep 'documentation: yes' config.log
           [[ $CH_TEST_BUILDER = buildah* ]] && fgrep 'with Buildah: yes' config.log
@@ -344,7 +352,7 @@ jobs:
           mv charliecloud-*.tar.gz $ch_prefix
           cd $ch_prefix
           tar xf charliecloud-*.tar.gz
-          rm charliecloud-*.tar.gz  # else cd fails with "too many arguments"
+          rm charliecloud-*.tar.gz  # else cd fails with “too many arguments”
           cd charliecloud-*
           pwd
           # Configure and verify output.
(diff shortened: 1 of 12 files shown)


reidpr:
@ana @olifre @lnussbaum @wiene : I'd love opinions on how distributions feel about this approach.

olifre:
Since the critical part for distros is `valid for the install location at install time`, I don't see any problem with this approach. To confirm this, I also tried to run this branch through Gentoo's packaging and the package manager was happy (i.e. not automated flagging), so that seems good for Gentoo :+1: .

lucaudill (approved):
LGTM

wiene:
At present the situation for Debian is the following:
[A Debian specific patch](https://salsa.debian.org/hpc-team/charliecloud/-/blob/74b951b0e61f237ee0950494dc5e2833bb080671/debian/patches/adjust-test-suite-path.patch) is applied to modify the path to the tests in `ch-test` (`CHTEST_DIR`).

I glanced through the changes. Unless I overlooked something, it seems that a patch is still needed after  the modifications of PR have been applied. Removing the above mentioned patch and applying this PR, I get
```
CHTEST_DIR=/usr/libexec/installed-tests/charliecloud/test
```
in `ch-test`. This is closer than before, but not precisely what I am looking for:
```
CHTEST_DIR=/usr/libexec/installed-tests/charliecloud
```
Of course, it is nice to get rid of distribution specific patches. On the other hand, if this only affects Debian, it is also fine to keep the patch.

e4t:
> @ana @olifre @lnussbaum @wiene : I'd love opinions on how distributions feel about this approach.

@ana asked me to respond for her: 
SUSE/openSUSE uses `libexec` on the rolling release (openSUSE Tumbleweed) and any upcoming product (ALP etc), however, SLE and Leap still adhere to the somewhat older FHS 2.3 standard.
For this RPM-based Linux distros provide a set of RPM predefined macros  (`%_libexecdir`) which are set to the path elements of the underlying system. These will be specified during configure (using a predefined `%_configure` macro for convenience). Thus, allowing these paths / path elements to be configurable allows distributions to set them to their needs without extensive patching.

reidpr:
Thanks folks.

@wiene I believe the latest version of this PR will let you eliminate your patch.

lucaudill (approved):
LGTM

wiene:
> @wiene I believe the latest version of this PR will let you eliminate your patch.

I can confirm that the patch can be removed thanks to this PR. Many thanks!

PR #1727, opened 2023-09-12
convert SIGINT and SIGTERM to Fatal_Error

### LINKED ISSUE #1518: SIGINT and SIGTERM can leave the storage directory in invalid state
Currently, we do not catch `KeyboardInterrupt` anywhere, nor install any `SIGINT` handler ourselves. Thus, control-C exits `ch-image` immediately (with a traceback) during various phases can leave a broken storage directory. For example:

* Interrupting `RUN` prevents roll-back, leaving a broken image. Recovery can be done by deleting the image.
* Interrupting a download leaves the blob file on disk, and it will be credulously re-used the next time needed. Recovery here is rather tricky, as the right file must be identified and manually deleted.

reidpr:
Closes #1518.


lucaudill (approved):
LGTM

PR #1729, opened 2023-09-13
tmpfs compatibility patch for `xattr` tests

### LINKED ISSUE #1726: xattrs tests fail if `/tmp` is a tmpfs
If `/tmp` is a tmpfs (or some other filesystem that does not support xattrs), then various tests fail, for example:

```
 ✗ cache: restore ACLs, xattrs
[...]
     5. RUN setfattr -n user.foo -v bar /home/foo
   setfattr: /home/foo: Not supported
   something went wrong, rolling back ...
   HEAD is now at 9e9809a RUN touch /home/foo
   error: build failed: RUN command exited with 1
   trace: build.py:1180:execute, build.py:250:__default__, visitors.py:241:_call_userfunc, visitors.py:269:visit_topdown, build.py:186:main, ch-image:314:main
```

Fortunately, `/tmp` is not a tmpfs on GitHub Actions, so CI still passes. I don't think this necessarily needs to be fixed for 0.34.

I think the thing to do is simply identify the problem tests and skip them if *both* (1) `/tmp` does not support xattrs and (2) we're not on GitHub Actions (i.e., `$GITHUB_ACTIONS` is not set).

lucaudill:
closes #1726

### DIFF
--- a/test/build/55_cache.bats
+++ b/test/build/55_cache.bats
@@ -1406,9 +1406,9 @@ EOF
 @test "${tag}: restore ACLs, xattrs" {  # issue #1287
     # Check if test needs to be skipped
     touch "$BATS_TMPDIR/tmpfs_test"
-    run setfattr -n user.foo -v bar "$BATS_TMPDIR/tmpfs_test"
-    if [[ ("$status" != 0) && (-z $GITHUB_ACTIONS) ]]; then
-        skip 'xattrs unsupported by tmpfs'
+    if    ! setfattr -n user.foo -v bar "$BATS_TMPDIR/tmpfs_test" \
+       && [[ -z $GITHUB_ACTIONS ]]; then
+        skip "xattrs unsupported in ${BATS_TMPDIR}"
     fi
 
     # Build an image, then re-build from cache to test xattr/ACL cache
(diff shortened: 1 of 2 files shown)


test/build/55_cache.bats
```diff
@@ -1404,6 +1404,13 @@ EOF
 }
 
 @test "${tag}: restore ACLs, xattrs" {  # issue #1287
+    # Check if test needs to be skipped
+    touch "$BATS_TMPDIR/tmpfs_test"
+    run setfattr -n user.foo -v bar "$BATS_TMPDIR/tmpfs_test"
+    if [[ ("$status" != 0) && (-z $GITHUB_ACTIONS) ]]; then
+        skip 'xattrs unsupported by tmpfs'
+    fi
```
reidpr, line 1412: You can do this more simply and idiomatically, something like:

```suggestion
    if    ! setfattr -n user.foo -v bar "$BATS_TMPDIR/tmpfs_test" \
       && [[ -z $GITHUB_ACTIONS ]]; then
        skip "xattrs unsupported in ${BATS_TMPDIR}"
    fi
```

If you did keep `run`, don't forget to `echo "$output"`.

reidpr (changes requested):
Good! Suggestion for better code in-line.

reidpr (changes requested):
yah you need to update the other file too lol

lucaudill:
> yah you need to update the other file too lol

oops 😬

reidpr (approved):

PR #1735, opened 2023-09-29
`ch-convert`: add `--quiet`

lucaudill:
This PR adds a `--quiet` option to `ch-convert` to suppress output to stdout and stderr.

### DIFF
--- a/bin/ch-convert
+++ b/bin/ch-convert
@@ -865,6 +865,9 @@ done
 if [ "$#" -ne 2 ]; then
     usage
 fi
+if [ $verbose -gt 0 ] && [ $quiet -gt 0 ]; then
+    FATAL "incompatible options: --quiet, --verbose"
+fi
 if [ -n "$no_xattrs" ]; then
     tar_xattr_args=
     squash_xattr_arg=
--- a/bin/ch-convert
+++ b/bin/ch-convert
@@ -79,7 +79,7 @@ cv_chimage_tar () {
     tar_out_validate "$2"
     img=$(chimage_path "$1")
     INFO 'exporting ...'
-    tar_ "$img" | quiet_process gzip_ > "$2"
+    tar_ "$img" | gzip_ > "$2"
 }
 
 cv_dir_chimage () {
@@ -128,7 +128,7 @@ cv_dir_tar () {
     # Don't add essential files & directories because that will happen later
     # when converted to dir or squash.
     INFO 'packing ...'
-    tar_ "$1" | quiet_process gzip_ > "$2"
+    tar_ "$1" | gzip_ > "$2"
 }
 
 cv_docker_chimage () {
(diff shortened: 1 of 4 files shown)


reidpr (changes requested):
I think we want to follow the same scheme as `ch-image`, i.e. allow repetition of `-q` and follow the chart in the FAQ.

bin/ch-convert
```diff
@@ -51,7 +51,7 @@ cv_chimage_dir () {
     # completely half-assed performance test suggests that's about the same.
     dir_make "$2"
     # shellcheck disable=SC2086
-    tar_ "$img" | tar xf - $tar_xattr_args -pC "$2"
+    tar_ "$img" | quiet_process tar xf - $tar_xattr_args -pC "$2"
```
reidpr, line 54: Can you just call it `quiet`?

bin/ch-convert
```diff
@@ -86,7 +86,7 @@ cv_dir_chimage () {
     dir_in_validate "$1"
     chimage_out_validate "$2"
     INFO 'importing ...'
-    ch-image import "$1" "$2"  # FIXME: no progress meter
+    quiet_process ch-image import "$1" "$2"  # FIXME: no progress meter
```
reidpr, line 89: Since `ch-image` understands `-q` and `-v`, what about passing the arguments through?

bin/ch-convert
```diff
@@ -108,18 +108,18 @@ cv_dir_squash () {
...
     INFO 'packing ...'
-    touch "$pflist"
-    mount_points_ensure "$1" "$pflist"
+    quiet_process touch "$pflist"
+    quiet_process mount_points_ensure "$1" "$pflist"
     # Exclude build cache metadata. 64kiB block size based on Shane's
     # experiments.
     # shellcheck disable=SC2086
-    mksquashfs "$1" "$2" $squash_xattr_arg -b 65536 -noappend -all-root -pf "$pflist" \
+    quiet_process mksquashfs "$1" "$2" $squash_xattr_arg -b 65536 -noappend -all-root -pf "$pflist" \
```
reidpr, line 116: Didn't count but this line looks too long.

bin/ch-convert
```diff
@@ -108,18 +108,18 @@ cv_dir_squash () {
     squash_out_validate "$2"
     pflist=${tmpdir}/pseudofiles
     INFO 'packing ...'
-    touch "$pflist"
-    mount_points_ensure "$1" "$pflist"
+    quiet_process touch "$pflist"
+    quiet_process mount_points_ensure "$1" "$pflist"
```
reidpr, line 112: Quieting our own functions seems risky?

lib/base.sh
```diff
@@ -12,8 +12,11 @@ ch_lib=${ch_bin}/../lib
 # Verbosity level; works the same as the Python code.
 verbose=0
 
+# Quiet level
+quiet=0
+
```
reidpr, line 17: Don't we have a single verbose level in the Python, rather than separate quiet and verbose? That seems tidier.

reidpr (changes requested):
Nice work! Seems very close. Some comments in-line.

reidpr (approved):

PR #1739, opened 2023-10-17
set large file threshold to 4 MiB

reidpr:
This hasn’t caused much trouble and the build cache paper exercised it pretty thoroughly, so time IMO to promote it from experimental.


reidpr:
This surfaced some notable bugs so it’s not going into 0.35.

PR #1741, opened 2023-10-18
copy large files instead of hard linking

reidpr:
Closes #1740.

Turns out hard links are not copy-on-write like we (implicitly) assumed. 🙄 Therefore we must copy the large files back.

In a future PR I hope to have some performance improvements for this.

PR #1742, opened 2023-10-19
use kernel fast paths for file copying

reidpr:
`ch-image` does a lot of file copying. In particular, as of PR #1741, build cache large files that are stored OOB are copied instead of hard-linked.

The Linux kernel has two fast paths for copying regular files:

1. Make a *reflink* copy, where the destination file gets a new inode but shares all the data extents of the source file in a copy-on-write manner. Some filesystems support this.
 
2. Copy data in-kernel without passing through user-space. All filesystems support this.

Python ≥3.8 introduces `os.copy_file_range()` that wraps the underlying system call [`copy_file_range(2)`](https://man7.org/linux/man-pages/man2/copy_file_range.2.html), which does #1 if possible and silently falls back to #2 if not.

This PR implements a fast-path copy that calls `os.copy_file_range()` if available and the source and destination are on the same filesystem. This should mitigate the performance loss in #1741 to some degree.

### DIFF
--- a/doc/best_practices.rst
+++ b/doc/best_practices.rst
@@ -49,48 +49,97 @@ including Lustre.
 File copy performance
 ---------------------
 
-:code:`ch-image` does a lot of file copying. The bulk of this is copying
-images around in the storage directory. Importantly, this includes :ref:`large
-files <ch-image_bu-large>` stored by the build cache outside its Git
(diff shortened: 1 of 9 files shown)


shanegoff (approved):
Everything looks good to me.

PR #1775, opened 2023-11-21
enforce code order in Python

### LINKED ISSUE #1756: enforce order of Python code with `ast`
We have an informal order of things in the Python files specified (i.e., alphabetical, divided by sections), but it's not enforced. The [`ast`](https://docs.python.org/3/library/ast.html) module looks like it could help with that. It doesn't handle comments, but it does provide line and column numbers which can be used to identify where blank lines and comments would be.

reidpr:
Closes #1756.


lucaudill (approved):
LGTM

PR #1777, opened 2023-11-27
implement missed `Path.iterdir()`

reidpr:
PR #1745 missed implementing `Path.iterdir()`, which meant old storage directories couldn’t be upgraded.


lucaudill (approved):
LGTM

PR #1780, opened 2023-11-27
remove leftover `$quiet` from `base.sh`

### LINKED ISSUE #1778: leftover `$quiet`
`base.sh` has some `$quiet` left over from development on #1735 that need to be updated to `$log_level`. Possibly there are more `$quiet` elsewhere too; I didn’t look. Example failure:

```
$ ch-convert spack /var/tmp/spack
input:   ch-image  spack
output:  dir       /var/tmp/spack
exporting ...
/home/reidpr/charliecloud/bin/ch-convert: 216: [: Illegal number:
done
```

lucaudill:
closes #1778

### DIFF
--- a/lib/base.sh
+++ b/lib/base.sh
@@ -163,7 +163,7 @@ vset () {
     if [ -z "$value" ]; then
         value=no
     fi
-    if [ "$quiet" -eq 0 ]; then
+    if [ "$log_level" -ge 0 ]; then
         var_desc="$var_desc:"
         printf "%-*s %s (%s)\n" "$desc_width" "$var_desc" "$value" "$method"
     fi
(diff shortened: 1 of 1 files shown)


lib/base.sh
```diff
@@ -163,7 +163,7 @@ vset () {
     if [ -z "$value" ]; then
         value=no
     fi
-    if [ "$quiet" -eq 0 ]; then
+    if [ -z "$quiet" ]; then
```
reidpr, line 166: But isn’t `quiet` gone? I though this should be `log_level` too. Looks like it was causing problems in my PR #1775.

reidpr (changes requested):

lib/base.sh
lucaudill: This is actually a [different `quiet`](https://github.com/hpc/charliecloud/blob/37dd7ea8998c6672f653636f04e0704ca0d5d4a6/lib/base.sh#L147) that is only meaningful in the scope of the function `vset`. I missed this detail in my original PR and changed it to `log_level`, which I imagine caused (or would inevitably cause) some problems.

reidpr: Right, OK. I see the comment disagrees with the variable names, so something needs to be cleaned up still.

reidpr (approved):

Issue #1774, opened 2023-11-18
ch-image fails with fedora and libutempr 


kb1vc:
Apparently  #1136 was closed too soon

I’ve used fedora:35 as an example, but this fails with fedora:39 as well

Here’s the Dockerfile
```
FROM fedora:35
RUN dnf install -y git
```
The host was 
```
$ cat /etc/redhat-release 
Fedora release 38 (Thirty Eight)
```
The charliecloud build was from the git repo — v0.35
```
$ ch-run —version
0.35
```

The image build fails when I try
```
ch-image build .
```
At the end of the output: 
```
Failed:
  git-core-2.38.1-1.fc35.x86_64          libutempter-1.2.1-5.fc35.x86_64    openssh-8.7p1-3.fc35.x86_64   
  openssh-clients-8.7p1-3.fc35.x86_64    util-linux-2.37.4-1.fc35.x86_64   

Error: Transaction failed
error: build failed: --force may fix it
error: build failed: RUN command exited with 1
```

I get a less broken result with 
```
ch-image build —force .
```

But it all still fails with 
```
Installed:
  fakeroot-1.30.1-1.fc35.x86_64      fakeroot-libs-1.30.1-1.fc35.x86_64    libfdisk-2.37.4-1.fc35.x86_64 
  systemd-libs-249.13-6.fc35.x86_64  util-linux-core-2.37.4-1.fc35.x86_64 
Failed:
  libutempter-1.2.1-5.fc35.x86_64                      util-linux-2.37.4-1.fc35.x86_64                     

Error: Transaction failed
error: command failed with code 1: /usr/bin/ch-run
```

reidpr:
Thank you for the nice bug report!

`ch-image build --force` has been the default since 0.34. Also, the build is running `/usr/bin/ch-run` which is an unusual location for a `ch-run`  build from source.

I can’t reproduce with 0.35:

```
$ which ch-image
/home/reidpr/charliecloud2/bin/ch-image
$ ch-image --version
0.35
$ printf 'FROM fedora:35\nRUN dnf install -y git\n' | ch-image build -t foo -
[...]
--force=seccomp: modified 0 RUN instructions
grown in 2 instructions: foo
```

I wonder if there is a mixed installation here, and that’s the source of your problem?

kb1vc:
I will try this from a raw fedora 38 build in a VM.  

matt

kb1vc:
[...]

PR #1782, opened 2023-11-28
don’t follow symlinks when setting xattrs

reidpr:
The call to `os.setxattr` erroneously passed our `follow_symlinks` to `flags`, not `follow_symlinks`. Therefore, we always followed symlinks, and set `os.XATTR_CREATE` if our `follow_symlinks` was True. This caused failures with image `fedora:35` on a Fedora 38 host, which has xattrs on image `/etc/mtab`, which is a symlink to `../proc/self/mounts`, which is broken until containerized (see #1774).

We only ever call `Path.setxattr()` with `follow_symlinks=False`, so this PR removes the keyword argument and unconditionally does not follow symlinks.

### DIFF
--- a/lib/build_cache.py
+++ b/lib/build_cache.py
@@ -488,7 +488,7 @@ class File_Metadata:
       elif (not self.path.git_compatible_p):
          self.path_abs.git_escaped.rename(self.path_abs)
       for (xattr, val) in self.xattrs.items():
-         self.path_abs.setxattr(xattr, val, follow_symlinks=False)
+         self.path_abs.setxattr(xattr, val)
       # Recurse children.
       if (len(self.children) > 0):
          for child in self.children.values():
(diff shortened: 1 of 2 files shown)


reidpr:
@kb1vc, you could try this branch. It should fix your xattrs failure.

lucaudill (approved):
LGTM

kb1vc:
Reid, 

Thank you.  —no-xattrs did it.  

Now I’m peeling another layer of the onion that is not charliecloud’s fault.  (This is all in an effort to build openEMS with its visualization tools in a charliecloud image/container. The task is difficult even on a host system.)

And I’ll try the new branch in a day or so.  

Thank you.

matt

reidpr:
@kb1vc, feel free to wander over to the [Discussions](https://github.com/hpc/charliecloud/discussions) if you have any usage questions that you think we might be able to help with, even if not directly Charliecloud’s fault.

PR #1784, opened 2023-11-29
use `pkg-config` to find `libfuse3`

reidpr:


configure.ac
```diff
@@ -395,21 +401,32 @@ AC_CHECK_PROG([SQUASHFUSE], [squashfuse], [squashfuse])
...
+have_libsquashfuse_ll=n/a
+have_ll_h=n/a
+AS_IF([test $want_libsquashfuse = yes], [
+  # libfuse3. Must use pkg-config because as of version 0.5.0 SquashFUSE’s
+  # ll.h won’t build without an appropriate -I [1].
+  # [1]: https://github.com/vasi/squashfuse/commit/eca5764
+  PKG_CHECK_MODULES([fuse3], [fuse3], [
+    # libfuse3 found
+    have_libfuse3=yes
+    CFLAGS="$CFLAGS $fuse3_CFLAGS"
```
lucaudill, line 415: It looks like we don't define `fuse3_CFLAGS` ourselves. Does this variable come from somewhere else? If so, it might be worth writing a comment about it here so it's more clear what its value is.

lucaudill (approved):
LGTM

PR #1576, opened 2023-02-27
fix some Git problems

reidpr:
Closes #1285. Closes #1535. Closes #1566.

PR #1785, opened 2023-11-29
ch-convert: exclude correct Git metadata

reidpr:
PR #1576 moved Git metadata out of band but `ch-convert` still tries to exclude the in-band metadata. Fix.


lucaudill (approved):
LGTM

PR #1793, opened 2023-12-07
add `--write-fake` via unprivileged overlayfs

### LINKED ISSUE #96: Add support for bind mounts to directories not existing within a container on read-only FS
On the WLCG containers mailing list, somebody suggested the following magic to be able to add bind-mount points to a read-only container without using overlayfs. 
- Use a temporary local, read-writable directory.
- Create the missing mount point directories in there. 
- Recreate the top-level directories which are present in the container in there. 
- Create bind mounts for all top-level directories which are part of the read-only container to those directories.
- Finally, chroot to the local, read-writable directory instead of the "container root". 

Is there a logic error in this approach?

reidpr:
This pull request implements `ch-run --write-fake`, which overlayfs a writeable tmpfs atop a read-only image. This makes the image appear writeable when it really is not (all writes are discarded on container exit). Needs a recent kernel: upstream 5.11 for the feature to work at all, and upstream 6.6 to avoid some weirdness (distros vary).

Closes #96, but I'm tagging it as a stand-alone PR because I think its scope exceeds that of #96.

### DIFF
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -27,6 +27,9 @@ char *JOIN_CT_ENV[] =  { "OMPI_COMM_WORLD_LOCAL_SIZE",
 char *JOIN_TAG_ENV[] = { "SLURM_STEP_ID",
                          NULL };
 
+/* Default overlaid tmpfs size. */
+char *WRITE_FAKE_DEFAULT = "12%";
+
 
 /** Command line options **/
 
@@ -75,7 +78,9 @@ const struct argp_option options[] = {
    { "verbose",       'v', 0,      0, "be more verbose (can be repeated)" },
    { "version",       'V', 0,      0, "print version and exit" },
    { "warnings",      -16, "NUM",  0, "log NUM warnings and exit" },
-   { "write",         'w', 0,      0, "mount image read-write"},
+   { "write",         'w', 0,      0, "mount image read-write (avoid)"},
+   { "write-fake",    'W', "SIZE", OPTION_ARG_OPTIONAL,
+                           "overlay read-write tmpfs on top of image" },
    { 0 }
 };
 
@@ -155,6 +160,7 @@ int main(int argc, char *argv[])
                                .join_ct = 0,
                                .join_pid = 0,
                                .join_tag = NULL,
+                               .overlay_size = NULL,
                                .private_passwd = false,
                                .private_tmp = false,
                                .type = IMG_NONE,
@@ -173,7 +179,7 @@ int main(int argc, char *argv[])
       argp_help_fmt_set = true;
    else {
       argp_help_fmt_set = false;
-      Z_ (setenv("ARGP_HELP_FMT", "opt-doc-col=25,no-dup-args-note", 0));
+      Z_ (setenv("ARGP_HELP_FMT", "opt-doc-col=27,no-dup-args-note", 0));
    }
    Z_ (argp_parse(&argp, argc, argv, 0, &arg_next, &args));
    if (!argp_help_fmt_set)
@@ -457,6 +463,8 @@ static error_t parse_opt(int key, char *arg, struct argp_state *state)
       break;
    case -12: // --home
       Tf (args->c.host_home = getenv("HOME"), "--home failed: $HOME not set");
+      if (args->c.overlay_size == NULL)
+         args->c.overlay_size = WRITE_FAKE_DEFAULT;
       break;
    case -13: // --unsafe
       args->unsafe = true;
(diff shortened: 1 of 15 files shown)


reidpr:
> Gentoo ... explodes when the new `CH_OVERLAY_C` check is executed

This check is (at present) only used to fill in the report `configure` presents at its conclusion. It would be simple to add a configure flag that disables the check; the only impact would be the report is slightly less informative.

The reasoning to include it is that many folks have no idea whether their kernel supports these things (kernel version is not reliable due to distro backports), so let’s try it and tell them.

olifre:
> Is this a regression? That is, if we merge this PR with `/home` dealt with as-is, does it make things worse for you?

It is not, and it does not make things worse :wink:. My actual use case relies on a read-only FS which contains the unpacked container. Hence, with Charliecloud without this PR, I do (expectedly) get:
```
ch-run[1156]: error: can't mkdir: /cvmfs/container.physik.uni-bonn.de/Debian12/default/1701962369/home/olifre: Read-only file system (ch_misc.c:461 30)
```
Hence, it's certainly an improvement for me, and it also fixes https://github.com/hpc/charliecloud/issues/96 (but in fact, I currently have images with `755` for `/home` on the very same read-only FS).

reidpr:
>> Is this a regression? That is, if we merge this PR with `/home` dealt with as-is, does it make things worse for you?
> 
> It is not, and it does not make things worse :wink:. My actual use case relies on a read-only FS which contains the unpacked container.

One thing we could do is retain the existing behavior as a backup: overmount a new tmpfs on container `/home`, then mkdir the appropriate home directory in there. I'd expect this to work for your use case, so I'm a little surprised it doesn't.

olifre:
> > > Is this a regression? That is, if we merge this PR with `/home` dealt with as-is, does it make things worse for you?
> > 
> > 
> > It is not, and it does not make things worse 😉. My actual use case relies on a read-only FS which contains the unpacked container.
> 
> One thing we could do is retain the existing behavior as a backup: overmount a new tmpfs on container `/home`, then mkdir the appropriate home directory in there. I'd expect this to work for your use case, so I'm a little surprised it doesn't.

You are of course correct, my bad: I compared:
```
ch-run -w -b /home/olifre -vv  /cvmfs/container.physik.uni-bonn.de/singularity/Debian12/default/1701962369 bash
```
for both the previous version and with the unprivileged overlayfs patch. What I forgot was that `--home` has the special case using `/tmpfs`, mainly since I do not rely on this special case (in fact, I think it would be useful to have a workable solution for general overmounting for such images on read-only file systems). 

So you are completely correct, `--home` is broken in this case when using unprivileged overlayfs vs. the working solution via an overmounted `tmpfs` before.

reidpr:
@olifre, this is very close if you want to have another look.

olifre:
@reidpr Many thanks! I did several more tests using containers stored on a read-only FS. It works exceptionally well. Expectedly, `findmnt` looks quite interesting inside the containers :wink: . 

In some cases, I hit the "15 directories" limit, but I did not hit that for actual use cases (I'm not sure if a real use case exists in which someone wants to add, for example, a sub-directory to `/etc/` without actually rebuilding the container). At least from my point of view the limit seems reasonable, and the functionality is working very well! :+1: 

Thanks a lot (and Merry Christmas!).

reidpr:
@olifre, thinking about that limit, I tried a different approach that was more symlink-based (similar to your suggestion I believe) in commit 1b775bd. Will you try it and compare/contrast?

olifre:
@reidpr Many thanks! Yes, indeed this was a possibility I was thinking of. This approach works great, and the symlink ranch (in this way) seems much more lightweight than the overmount farm :wink: . It removes the limitation without the risk of getting an explosive number of bind mounts, which is great! 


I did not observe any issues with any use case I had at hand, but of course some differences are seen:
* Symlinks can naturally be deleted by the user in the container, as the links themselves don't inherit the read-only permissions of the directories from the underlying read-only image. 
* In the "ranched" directory, things are symlinks and not "directories" or "files" anymore, as expected. 

I don't have any use case at hand which has a problem with that, but wanted to mention it, since some tooling is sensitive to such things (e.g. `sshd`, but I don't think running `sshd` inside a software stack container and overmounting directory components it uses is something anybody should be doing). 

I would personally say "if a real-world program breaks since it does not follow links, but checks permissions / type before resolving links", that's likely not wanted behaviour, but a bug in the program instead. 

So I think this approach has big advantages, e.g. allowing to bind into crowded directories such as `/etc/` on read-only images. On the other hand, I don't believe it will break anything — famous last words :wink: .

lucaudill (approved):
LGTM

kchilleri (approved):
Looks good

PR #1798, opened 2023-12-19
`ch-run`: add `--quiet`

### LINKED ISSUE #1660: `ch-run`: add `--quiet`
#1636 adds a `--quiet` option to `ch-image` to suppress logging and writing to stdout. Originally, that PR was intended to also add the same option to `ch-run`, but we've decided that doing so is out of scope given the nature of the bug being addressed. Still, it would probably be a good idea to add `--quiet` to `ch-run` eventually. Take a look at [this commit](https://github.com/hpc/charliecloud/pull/1636/commits/7ff10650a0d809603deba228e0fc8122f3224d78) to see how I originally laid the groundwork for that.

lucaudill:
closes #1660

### DIFF
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -60,7 +60,7 @@ const struct argp_option options[] = {
    { "join-pid",       -5, "PID",  0, "join a namespace using a PID" },
    { "join-ct",        -3, "N",    0, "number of join peers (implies --join)" },
    { "join-tag",       -4, "TAG",  0, "label for peer group (implies --join)" },
-   { "test",  -17, "TEST", 0, "do test TEST" },
+   { "test",          -17, "TEST", 0, "do test TEST" },
    { "mount",         'm', "DIR",  0, "SquashFS mount point"},
    { "no-passwd",      -9, 0,      0, "don't bind-mount /etc/{passwd,group}"},
    { "private-tmp",   't', 0,      0, "use container-private /tmp" },
@@ -487,13 +487,12 @@ static error_t parse_opt(int key, char *arg, struct argp_state *state)
       exit(0);
       break;
    case -17: // --test
-      if (!strcmp(arg, "log")) {
+      if (!strcmp(arg, "log"))
          test_logging(false);
-      } else if (!strcmp(arg, "log-fail")) {
+      else if (!strcmp(arg, "log-fail"))
          test_logging(true);
-      } else {
+      else
          FATAL("invalid --test argument: %s; see source code", arg);
-      }
       break;
    case 'b': {  // --bind
          char *src, *dst;
--- a/bin/ch-run.c
+++ b/bin/ch-run.c
@@ -57,6 +57,7 @@ const struct argp_option options[] = {
    { "join-pid",       -5, "PID",  0, "join a namespace using a PID" },
    { "join-ct",        -3, "N",    0, "number of join peers (implies --join)" },
    { "join-tag",       -4, "TAG",  0, "label for peer group (implies --join)" },
+   { "logging",        -17, "FAIL", OPTION_ARG_OPTIONAL, "fooooooo" },
    { "mount",         'm', "DIR",  0, "SquashFS mount point"},
    { "no-passwd",      -9, 0,      0, "don't bind-mount /etc/{passwd,group}"},
    { "private-tmp",   't', 0,      0, "use container-private /tmp" },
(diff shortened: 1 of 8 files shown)


bin/ch-run.c
```diff
@@ -57,9 +57,11 @@ const struct argp_option options[] = {
    { "join-pid",       -5, "PID",  0, "join a namespace using a PID" },
    { "join-ct",        -3, "N",    0, "number of join peers (implies --join)" },
    { "join-tag",       -4, "TAG",  0, "label for peer group (implies --join)" },
+   { "logging-test",  -17, "FAIL", OPTION_ARG_OPTIONAL, "test log levels" },
```
reidpr, line 60: I wonder if this would be more future-proof as something like:
```suggestion
   { "test",  -17, "TEST", 0, "do test TEST" },
```
with valid `TEST` being `log` and `log-fail`?

bin/ch-run.c
```diff
@@ -466,13 +468,20 @@ static error_t parse_opt(int key, char *arg, struct argp_state *state)
...
          WARNING("this is warning %d!", i);
       exit(0);
       break;
-   case -15: // --set-env0
-      parse_set_env(args, arg, '\0');
+   case -17: // --logging-test
+         if (arg == NULL) {
+            logging_print(false);
+         } else {
+            logging_print((!strcmp(arg, "fail")));
```
reidpr, line 483: Does this mean all arguments other than `fail` are accepted but a no-op? That seems a bit user-unfriendly, but it might be fine for a test option.

bin/ch_misc.c
```diff
@@ -406,6 +406,19 @@ void log_ids(const char *func, int line)
    }
 }
 
+void logging_print(bool fail) {
```
reidpr, line 409: Maybe call it `test_logging`? Or perhaps just `kenny`.

bin/ch_misc.c
```diff
@@ -406,6 +406,19 @@ void log_ids(const char *func, int line)
...
+   TRACE("trace");
+   DEBUG("debug");
+   VERBOSE("verbose");
+   INFO("info");
+   WARNING("warning");
+   if (fail) {
+      FATAL("the program failed inexplicably");
```
reidpr, line 416: As much as I like this joke, I wonder if it will become confusing over time and the message should be more clear about what his happening.

Also you don't need braces on this `if`.

bin/ch_misc.c
```diff
@@ -406,6 +406,19 @@ void log_ids(const char *func, int line)
    }
 }
 
+void logging_print(bool fail) {
+   TRACE("trace");
+   DEBUG("debug");
+   VERBOSE("verbose");
+   INFO("info");
+   WARNING("warning");
```
reidpr, line 414: Add `ERROR`? Do we have `ERROR` in the C stuff?

doc/ch-run.rst
```diff
@@ -96,6 +96,11 @@ mounting SquashFS images with FUSE.
     bind-mounted into it. If this is specified, no such temporary files are
     created and the image’s files are exposed.
 
+  :code:`-q`, :code:`--quiet`
+    Be quieter; can be repeated. Incompatible with :code:`-v` and suppresses
+    :code:`--debug` regardless of option order. See the :ref:`FAQ entry on
```
reidpr, line 101: Does `ch-run` have `--debug`? I'm not seeing it.

doc/ch-run.rst
```diff
@@ -130,7 +135,8 @@ mounting SquashFS images with FUSE.
     Unset environment variables whose names match :code:`GLOB`.
 
   :code:`-v`, :code:`--verbose`
-    Be more verbose (can be repeated).
+    Print extra chatter; can be repeated. See the :ref:`FAQ entry on verbosity
+    <faq_verbosity>` for details.
```
reidpr, line 139: I discovered recently that you don't have to repeat the target title; you can just say:

```
:ref:`faq_verbosity`
```

and it will use the title of the target for the link title.

reidpr (changes requested):
Very solid. Some suggestions in-line.

bin/ch_misc.c
lucaudill: The C code doesn't have an `ERROR` logging function.
```
$ misc/grep -F '#define FATAL'
./bin/ch_misc.h:68:#define FATAL(...)   msg_fatal(      __FILE__, __LINE__, 0, __VA_ARGS__);
$ misc/grep -F '#define ERROR'
$
```

doc/ch-run.rst
lucaudill: oops, good catch

reidpr (approved):
Great work, thank you!

PR #1804, opened 2024-01-10
add tab completion to `ch-convert`

lucaudill:


lucaudill:
Looks like this PR is also removing a bug that makes `ch-run` [erroneously complete tarballs](https://github.com/hpc/charliecloud/blob/62d8693283a43a11a6cbf5fc52f97224d9c768c9/bin/ch-completion.bash#L520 ) as images that can be run. Oops.

PR #1815, opened 2024-01-19
fix scope handling

reidpr:
This PR fixes a bug where (1) two example Dockerfiles were missing scope and (2) this problem did not fail the tests.

The missing scope was interpreted as "quick". In particular this made CI quite a bit slower because the Spack image takes a long time to build.


lucaudill (approved):
Looks good to me!

PR #1818, opened 2024-01-23
tidy tar-related output

reidpr:
Sometime recently, Red Hat-distributed Pythons have started warning:

> `/usr/lib64/python3.6/tarfile.py:2221`: RuntimeWarning: The default behavior of tarfile extraction has been changed to disallow common exploits (including CVE-2007-4559). By default, absolute/parent paths are disallowed and some mode bits are cleared. See https://access.redhat.com/articles/7004769 for more details. RuntimeWarning)

As a nice touch, that URL is subscriber-only. But, the listed CVE is about path sanitization in tar archives, which we already do. Therefore, suppress the warning.

This PR also makes our warning about fixing absolute member paths actually work.

The specific feature appears to be a backport of a [3.12 feature](https://peps.python.org/pep-0706/). We should keep an eye on things as this may cause further disruption in the future.

### DIFF
--- a/lib/charliecloud.py
+++ b/lib/charliecloud.py
@@ -20,6 +20,7 @@ import subprocess
 import sys
 import time
 import traceback
+import warnings
 
 
 # List of dependency problems. This variable needs to be created before we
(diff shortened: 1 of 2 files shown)


lucaudill (approved):
LGTM

PR #1824, opened 2024-01-26
fix crash with unlinked worktrees

reidpr:
`ch-image` can get into a state where all commands crash with a traceback similar to:

```
$ ch-image build -t paraview -f ./examples/paraview/Dockerfile ./examples
  File "[...]/bin/ch-image", line 332, in <module>
    main()
  File "[...]/bin/ch-image", line 319, in main
    bu.init(cli)
  File "[...]/lib/build_cache.py", line 184, in init
    cache = Enabled_Cache(cli.cache_large)
  File "[...]/lib/build_cache.py", line 677, in __init__
    self.worktrees_fix()
  File "[...]/lib/build_cache.py", line 1329, in worktrees_fix
    assert (wt_gits == wt_actuals)
AssertionError
```

The crash is because one or more images think they are a Git worktree (i.e., `img/$FOO/ch/git` exists) but Git has no record of the worktree (i.e., `bucache/worktrees/$FOO` does not)

I haven’t been able to reproduce the situation itself. We think it has something to do with mixing `--cache` (usually the default) and `--no-cache`.

This PR converts the crash to a warning and requests help reproducing the underlying bug.

### DIFF
--- a/lib/build_cache.py
+++ b/lib/build_cache.py
@@ -1324,6 +1324,15 @@ class Enabled_Cache:
                                              // "*" // im.GIT_DIR)) }
       wt_gits =    { fs.Path(i).name
                      for i in glob.iglob("%s/worktrees/*" % self.root) }
+      # Unlink images that think they are in Git but are not. This should not
+      # happen, but it does, and I wasn’t able to figure out how it happened.
+      wt_gits_orphaned = wt_actuals - wt_gits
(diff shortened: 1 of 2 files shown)


j-ogas (approved):
LGTM. Unfortunately I went critical on the cache that produced this bug; I would have liked to try this on it.

PR #1825, opened 2024-01-30
add new log function for `printf(3)`-style debugging

reidpr:


kchilleri (approved):
This is going to be SOOO useful! Thanks!

PR #1826, opened 2024-02-01
`old-storage`: better feedback, fix AlmaLinux

reidpr:
This PR does two things:

1. Fix `test/old-storage`, which has been failing on AlmaLinux 8 due to GPG key problems (turns out the AlmaLinux people [lost their master key](https://almalinux.org/blog/2023-12-20-almalinux-8-key-update/) in December 2023).

2. Change `old-storage` to test all provided tarballs, rather than stopping after the first failure, the reasoning being that more complete feedback is actually better than a faster notification of failure.

At present, `old-storage` tests fail downloading `hello-2.10.tar.gz` because TripAdvisor is HTTP 500 (again). However, other than this the tests look good so I am going to declare victory anyway.


lucaudill (approved):
LGTM

PR #1827, opened 2024-02-01
use `kernel.org` for GNU Hello

reidpr:
Tests occasionally fail because we can’t download GNU Hello, and mirror `tripadvisor.com` is often the culprit. `ftp.gnu.org` itself has also been unreliable in the past (PR #1642). Therefore, use `kernel.org`.


lucaudill (approved):
LGTM

PR #1832, opened 2024-02-02
`ch-fromhost`: tidy

### LINKED ISSUE #732: ch-fromhost: ldconfig prints errors when certain paths don't exist
This is an issue reported by Dave Rich.

My thoughts:
`ldconfig` appears to assume that certain directories exist in an image and will attempt to `stat` them unsuccessfully resulting in errors. This is then passed to the user by `ch-fromhost` resulting in confusion as to whether or not `ch-fromhost` was successful when it was.

For example:
```
$ ch-fromhost -p /usr/lib/x86_64-linux-gnu/libpthread.so -v /var/tmp/images/centos8/
asking ldconfig for shared library destination
/sbin/ldconfig: Can't stat /libx32: No such file or directory
/sbin/ldconfig: Can't stat /usr/libx32: No such file or directory
shared library destination: /usr/lib64//bind9-export
injecting into image: /var/tmp/images/centos8/
  /usr/lib/x86_64-linux-gnu/libpthread.so -> /usr/lib64//bind9-export (inferred)
running ldconfig
```
Note: without the `-v` flag nothing besides the errors is printed making it appear as if the command was unsuccessful.

This should be documented somewhere and perhaps handled by `ch-fromhost` although I worry that if we aren't careful we may suppress legitimate errors with the latter.

### LINKED ISSUE #1830: `ch-fromhost`: incorrect arguments in man page

### LINKED ISSUE #1831: `ch-fromhost`: fix logging verbosity
PR #1814 inadvertently demoted some logging from VERBOSE to DEBUG. Fix that.

reidpr:
Closes #732. Closes #1830. Closes #1831.

### DIFF
--- a/bin/ch-fromhost
+++ b/bin/ch-fromhost
@@ -4,10 +4,10 @@
 # source:destination pairs separated by newlines, then walk through them and
 # copy them into the image.
 #
-# The colon separator is to avoid the difficulty of iterating through a sequence
-# of pairs with no arrays or structures in POSIX sh. We could avoid it by
-# taking action immediately upon encountering each file in the argument list,
-# but that would (a) yield a half-injected image for basic errors like
+# The colon separator is to avoid the difficulty of iterating through a
+# sequence of pairs with no arrays or structures in POSIX sh. We could avoid
+# it by taking action immediately upon encountering each file in the argument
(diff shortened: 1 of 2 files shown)


lucaudill (approved):
LGTM

Issue #1758, opened 2023-10-31
`ch-image` does notice incomplete blob downloads


reidpr:
If a blob download ends before it should, `ch-image` doesn’t notice and happily uses the file.

We could also validate the hashes of blobs.

PR #1833, opened 2024-02-06
Return error if blob download is incomplete by validating hash digest.

kchilleri:
Closes issue #1758

### DIFF
--- a/lib/registry.py
+++ b/lib/registry.py
@@ -400,7 +400,7 @@ class HTTP:
       sw.close()
 
       # Check for integrity of downloaded blob by validating its hash digest
-      if (str(digest) != str(m.hexdigest())):
+      if (digest != m.hexdigest()):
          ch.FATAL("Incomplete blob download.")
 
    def blob_upload(self, digest, data, note=""):
--- a/lib/registry.py
+++ b/lib/registry.py
@@ -4,6 +4,7 @@ import os
 import re
 import types
 import urllib
+import hashlib
 
 import charliecloud as ch
 
@@ -387,7 +388,7 @@ class HTTP:
       #    requests.head() does not follow redirects, and it seems like a
       #    weird status, so I worry there is a gotcha I haven’t figured out.
       url = self._url_of("blobs", "sha256:%s" % digest)
-      res = self.request("HEAD", url, {200,401,404})
+      res, m = self.request("HEAD", url, {200,401,404})
       return (res.status_code == 200)
 
    def blob_to_file(self, digest, path, msg):
@@ -395,8 +396,12 @@ class HTTP:
       # /v2/library/hello-world/blobs/<layer-hash>
       url = self._url_of("blobs", "sha256:" + digest)
       sw = ch.Progress_Writer(path, msg)
-      self.request("GET", url, out=sw)
+      res, m = self.request("GET", url, out=sw)
       sw.close()
+
+      # Check for integrity of downloaded blob by validating its hash digest
+      if (str(digest) != str(m.hexdigest())):
+         ch.FATAL("Incomplete blob download.")
 
    def blob_upload(self, digest, data, note=""):
       """Upload blob with hash digest to url. data is the data to upload, and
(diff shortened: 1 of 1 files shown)


lib/registry.py
```diff
@@ -387,17 +388,21 @@ def blob_exists_p(self, digest):
...
       "GET the blob with hash digest and save it at path."
       # /v2/library/hello-world/blobs/<layer-hash>
       url = self._url_of("blobs", "sha256:" + digest)
       sw = ch.Progress_Writer(path, msg)
-      self.request("GET", url, out=sw)
+      res, m = self.request("GET", url, out=sw)
       sw.close()
 
+      # Check for integrity of downloaded blob by validating its hash digest
+      if (str(digest) != str(m.hexdigest())):
```
reidpr, line 403: I'm pretty sure both of these are already strings? i.e. I think you want:
```suggestion
      if (digest != m.hexdigest()):
```

lib/registry.py
```diff
@@ -387,17 +388,21 @@ def blob_exists_p(self, digest):
...
+         ch.FATAL("Incomplete blob download.")
```
reidpr, line 404: 1. Does this correctly describe all the reasons this error could happen? If not, revise.
2. Re. style, we start errors with a lower-case letter and don't add sentence-ending punctuation, i.e.:
```suggestion
         ch.FATAL("incomplete blob download")
```

lib/registry.py
```diff
@@ -557,8 +563,9 @@ def request(self, method, url, statuses={200}, out=None, **kwargs):
          out.start(length)
          for chunk in res.iter_content(ch.HTTP_CHUNK_SIZE):
             out.write(chunk)
+            m.update(chunk)
       # Done.
-      return res
+      return res, m
```
reidpr, line 568: This interface change would change the basic contract of the method. Right now it cannot fail: either it returns a valid response, or it exits the program with an error. This change adds the possibility the method can fail, so every caller would have to deal with that. I suggest instead adding a keyword argument that is the expected digest, and `request()` does the checking.

There is also a subtle interaction with `out` to consider.

reidpr (changes requested):
Great work; you've solved the essence of the issue.

See in-line for some other comments.

Also — how might we test this? It's a fairly rare case for the error to come up naturally, so the new code path won't be exercised much unless we add test(s). I have some ideas that we can talk about, but first I'd like you to noodle on it some.

lib/registry.py
```diff
@@ -395,7 +396,7 @@ def blob_to_file(self, digest, path, msg):
       # /v2/library/hello-world/blobs/<layer-hash>
       url = self._url_of("blobs", "sha256:" + digest)
       sw = ch.Progress_Writer(path, msg)
-      self.request("GET", url, out=sw)
+      res = self.request("GET", url, out=sw, expdigest=digest)
```
reidpr, line 399: This return value looks left over from an earlier iteration of the PR?
```suggestion
      self.request("GET", url, out=sw, expdigest=digest)
```

lib/registry.py
```diff
@@ -518,7 +519,7 @@ def manifest_upload(self, manifest):
       self.request("PUT", url, {201}, data=manifest,
                    headers={ "Content-Type": TYPES_MANIFEST["docker2"] })
 
-   def request(self, method, url, statuses={200}, out=None, **kwargs):
+   def request(self, method, url, statuses={200}, out=None, expdigest=None, **kwargs):
```
reidpr, line 522: Document the new keyword argument in the docstring.

Also style — looks like this line may be more than 80 characters? If so please wrap.

lib/registry.py
```diff
@@ -557,6 +559,11 @@ def request(self, method, url, statuses={200}, out=None, **kwargs):
          out.start(length)
          for chunk in res.iter_content(ch.HTTP_CHUNK_SIZE):
             out.write(chunk)
+            m.update(chunk)
+         # Check for integrity of downloaded blob by validating its hash digest against expected hash digest
```
reidpr, line 563: Wrap comment to 80 characters.

lib/registry.py
```diff
@@ -557,6 +559,11 @@ def request(self, method, url, statuses={200}, out=None, **kwargs):
...
+         if (expdigest is not None):
+            if (expdigest != m.hexdigest()):
```
reidpr, line 565: This can be condensed to a single if statement:

```suggestion
         if (expdigest is not None and expdigest != m.hexdigest()):
```

lib/registry.py
```diff
@@ -557,6 +559,11 @@ def request(self, method, url, statuses={200}, out=None, **kwargs):
...
+               ch.FATAL("incomplete blob download")
```
reidpr, line 566: This error message should be specific about what we know and what we don’t know, and speculation about cause should be labeled clearly. In this case, one possibility would be (don’t take this as a given, think about it some):

```suggestion
               ch.FATAL("registry response has bad digest")
```

Specifically, note that at this level, we don’t know whether it’s a blob being downloaded or something else, and we don’t know why the digest did not match.

reidpr (changes requested):
Great work. The logic seems quite solid to me. I have some style recommendations in-line.

One oversight, unless I’ve missed something: `request()` only checks the digest if the response is going to a file, i.e. `out != None`. I think this is fine, but the arguments need to be checked. As written, if `out == None` and `expdigest != None`, the latter will be silently ignored. It should instead fail noisily. `assert` would be fine I think.

reidpr (approved):
Nice. I didn’t see an assert so I added one.

PR #1835, opened 2024-02-09
fix bad escape sequences

### LINKED ISSUE #1834: `ch-image build` issues syntax warnings with Python 3.12
Running `ch-image build` on a system using Charliecloud 0.36 and Python 3.12 one gets the following syntax warnings:
```
/usr/lib/charliecloud/build_cache.py:910: SyntaxWarning: invalid escape sequence '\g'
  "#\g<0>", text, flags=re.MULTILINE)
/usr/lib/charliecloud/force.py:187: SyntaxWarning: invalid escape sequence '\['
  "if ! grep -Eq '\[epel\]' /etc/yum.conf /etc/yum.repos.d/*; then "
/usr/lib/charliecloud/force.py:202: SyntaxWarning: invalid escape sequence '\['
  "if ! grep -Eq '\[epel\]' /etc/yum.conf /etc/yum.repos.d/*; then "
```
This issue was initially reported in [Debian bug 1063467](https://bugs.debian.org/1063467).

reidpr:
We were depending on unknown escape sequences passing through unchanged, which has been deprecated behavior since 3.6. In 3.12, it generates a `SyntaxWarning`, which causes some distro test suites to fail.

Closes #1834.

### DIFF
--- a/lib/build_cache.py
+++ b/lib/build_cache.py
@@ -907,7 +907,7 @@ class Enabled_Cache:
             i += 1
          text = "\n".join(lines)
          text = re.sub(r"^(D|M [0-7]+ [0-9a-f]+) \.(git|weirdal_)ignore$",
-                       "#\g<0>", text, flags=re.MULTILINE)
+                       r"#\g<0>", text, flags=re.MULTILINE)
          #fs.Path("/tmp/new").file_write(text)
          self.git(["fast-import", "--force"], input=text)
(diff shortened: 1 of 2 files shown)


reidpr:
@wiene, does this fix the bug?

wiene:
> @wiene, does this fix the bug?

Yes, it does. Many thanks for the prompt fix.

Issue #1844, opened 2024-02-21
configure: add `--with-libfuse3`


reidpr:
FUSE3 does not put its headers in a standard location, so we need an appropriate `-I`. Currently we get this from `pkg-config`, but it would be nice if it could also be specified manually with something like `--with-fuse3-headers` and `--with-fuse-libs`.

PR #1845, opened 2024-02-21
rm `pkg-config` dependency at `autogen.sh` time

reidpr:
PR #1784 introduced a dependency on `pkg-config` at `autogen.sh` time. Even though it’s not great that we always require `pkg-config` to find libfuse3 (see issue #1844), we definitely don’t want to require it unconditionally at `autogen.sh` time. This PR moves the dependency to configure time and only if `--with-libsquashfuse=yes` (the default) or a path.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -63,12 +63,6 @@ MKDIR_P=${MKDIR_P:-install -d -m 0755}
 
 AM_INIT_AUTOMAKE([1.13 -Wall -Werror foreign subdir-objects])
 
-# Check for “pkg-config”. It’s here because we use PKG_CHECK_MODULES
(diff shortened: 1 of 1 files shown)


kchilleri (approved):
This will be very useful. Looks good to me!

reidpr:
The error message if the dependency is not met is quite misleading:

```
 configure.ac:304: error: possibly undefined macro: AC_CHECK_LIB
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:504: error: possibly undefined macro: AC_CHECK_HEADER
autoreconf: error: /usr/bin/autoconf failed with exit status: 1
```

PR #1846, opened 2024-02-28
Adding Experiment Directory

laytonmccafferty:
Updates `lib/filesystem.py` to include an extra directory known as `experiment`, which will contain all of the data gathered from the build cache performance experiment.

### DIFF
--- a/lib/filesystem.py
+++ b/lib/filesystem.py
@@ -1096,6 +1096,10 @@ class Storage:
    @property
    def build_large(self):
       return self.root // "bularge"
+   
+   @property
+   def experiment_dir(self):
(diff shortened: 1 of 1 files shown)


reidpr (changes requested):
Thanks Layton!

A few comments:

1. At this level we want a generic direction called `experiment`. Your stuff would then be a sub-directory within that. The goal is that future experiments would then be other sub-directories.

2. Increment the storage version. Otherwise, older Charliecloud versions that don’t know about the experiment directory will fail, complaining about unexpected files.

3. Document the directory’s purpose somewhere (possibly `experiment_dir()` method?).

PR #1853, opened 2024-03-05
tutorial: fix missing hello.py in output

### LINKED ISSUE #1852: tutorial: missing `hello.py` in output
## Section [2.6.3. Pull and compare](https://hpc.github.io/charliecloud/tutorial.html#pull-and-compare)

In the last command:

```bash
ls ./hello.2
```

shouldn't we also see `hello.py` in its output? - the whole point is to compare that it's the same image as the one we pushed in the previous section, which had this file

VladimirFokow:
fix #1852


reidpr:
Merged, thanks for the PR!!

PR #1870, opened 2024-03-21
ch-run: run oci bundles

### LINKED ISSUE #1754: run OCI bundles
OCI bundles are pretty simple and `ch-run` should be able to run them.

kchilleri:
If option ```--oci``` or ```--o``` is provided/enabled and the image path is the path to an OCI bundle containing a directory named ```rootfs``` and a spec file ```config.json``` then parse the JSON file and distribute the arguments to the corresponding Charliecloud container arguments.

Example:
```
kchilleri@dev:/usr/local/src/charliecloud$ ls ~/examples/redis
config.json  rootfs
kchilleri@dev:/usr/local/src/charliecloud$ ls ~/examples/redis/rootfs/
bin  boot  data  dev  etc  home  lib  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
kchilleri@dev:/usr/local/src/charliecloud$ cat ~/examples/redis/config.json 
{
        "ociVersion": "1.0.2-dev",
        "process": {
                "terminal": true,
                "user": {}, 

                "args": [
                        "sh", "redis-server", "--bind", "0.0.0.0"
                ],
                "env": [
                        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
                        "TERM=xterm"
                ],
                "cwd": "/",
                "capabilities": {
                        "bounding": [
                                "CAP_AUDIT_WRITE",
                                "CAP_KILL",
                                "CAP_NET_BIND_SERVICE"
                        ],
...
kchilleri@dev:/usr/local/src/charliecloud$ ch-run --oci ~/examples/redis -- true
```

closes #1754


bin/ch-run.c
```diff
@@ -216,6 +216,12 @@ int main(int argc, char *argv[])
    case IMG_NONE:
       FATAL("unknown image type: %s", args.c.img_ref);
       break;
+   case OCI_BUNDLE:
+      if (args.c.newroot != NULL)  // --mount was set
+         WARNING("--mount invalid with directory image, ignoring");
+      args.c.newroot = realpath_(cat(args.c.img_ref, "/rootfs"), false);
+      img_directory_verify(args.c.newroot, &args);
```
reidpr, line 223: Three of the four lines here are identical to `IMG_DIRECTORY`, which isn’t DRY. Is there a way to re-work things to avoid repeated code?

bin/ch-run.c
```diff
@@ -216,6 +216,12 @@ int main(int argc, char *argv[])
    case IMG_NONE:
```
reidpr, line 216: I'd like to keep `IMG_NONE` last. Otherwise they should be alphabetical.

bin/ch_core.c
```diff
@@ -380,8 +382,15 @@ enum img_type image_type(const char *ref, const char *storage_dir)
    Zf (stat(ref, &st), "can't stat: %s", ref);
 
    // If ref is the path to a directory, then it’s a directory.
-   if (S_ISDIR(st.st_mode))
-      return IMG_DIRECTORY;
+   // If ref is the path to a directory that contains rootfs directory
+   // and config.json file, assume it is an oci bundle
```
reidpr, line 386: But this “if” is for both directory format and OCI bundle format images?

bin/ch_core.c
```diff
@@ -370,6 +370,8 @@ enum img_type image_type(const char *ref, const char *storage_dir)
    struct stat st;
    FILE *fp;
    char magic[4];  // four bytes, not a string
+   char *conff = cat(ref, "/config.json");
+   char *rootfs = cat(ref, "/rootfs");
```
reidpr, line 374: Use `path_join()` for this (you may see `cat()` in the code but that’s obsolete).

bin/ch_core.h
```diff
@@ -25,6 +25,7 @@ enum img_type {
    IMG_SQUASH,     // SquashFS archive file (not yet mounted)
    IMG_NAME,       // name of image in storage
    IMG_NONE,       // image type is not set yet
+   OCI_BUNDLE,     // contains rootfs img_directory and config.json file
```
reidpr, line 28: Call it `IMG_OCI_BUNDLE` to keep the consistent prefix.

reidpr (changes requested):
Nice work.

I’d like to also see documentation for this. In particular, as written, the PR ignores `config.json`. We’ll need to justify this choice. Looking briefly over the `config.json` spec, I do see a lot of things that correspond to `ch-run` arguments, as well as a lot of stuff that we can’t support, e.g. complex UID/GID mappings.  We’ll need to decide what to do with all that.

There are quite a few JSON libraries for C, so we could consider those. A Python wrapper to process `config.json` would also be a reasonable choice.

reidpr:
e.g. [this one](https://github.com/json-parser/json-parser) is BSD licensed and recommends “just drop `json.c` and `json.h` into your project”.

reidpr:
[this one](https://github.com/DaveGamble/cJSON) similarly

j-ogas:
FWIW, we should embrace JSON parsing sooner, rather than later. I suspect it would be very useful for things like oci-hooks for GPU injection.

reidpr:
@j-ogas do you have an opinion on whether it lives in `ch-run` or a Python wrapper?

bin/cjson-parser.c
```diff
@@ -0,0 +1,106 @@
...
+        fprintf(stderr, "Error: could not open file %s\n", filename);
+        return;
+    }
+
+    // Get the file length
+    fseek(file, 0, SEEK_END);
+    length = ftell(file);
+    fseek(file, 0, SEEK_SET);
+    // Allocate content buffer
+    buffer = (char *)malloc((size_t)length + sizeof(""));
```
reidpr, line 71: Nice, I was about to complain about a buffer overflow below.

`sizeof()` here I think it computing the length of the pointer to the string. `sizeof(char)` would be correct, though I think that’s defined to be 1 which would be idiomatic also.

bin/cjson-parser.c
```diff
@@ -0,0 +1,106 @@
...
+
+    // Parse the JSON data
+    cJSON *json = cJSON_Parse(buffer);
+    // Check if parsing was successful
+    if (json == NULL) {
+        const char *error_ptr = cJSON_GetErrorPtr();
+        if (error_ptr != NULL) {
+            fprintf(stderr, "Error before: %s\n", error_ptr);
+        }
+        cJSON_Delete(json);
```
reidpr, line 86: I liked the `goto` trick in the example to avoid repeating the freeing code. Linux uses that a lot too.

bin/cjson-parser.c
```diff
@@ -0,0 +1,106 @@
...
+
+int main() {
+    const char *filename = "config.json";
+    readJSONFile(filename);
+    return 0;
+}
+
+void iterJSONFile(cJSON *json) {
+ 
+    cJSON *proc_key = cJSON_GetObjectItemCaseSensitive(json, "process");
```
reidpr, line 18: I’d recommend either a `cJSON_GetObject` based approach or an iterative approach, but not a mixed approach like you currently have now.

I wonder also if you could have a table-based approach of some kind that maps keys to functions for a given element. E.g. (you’d need some supporting types and the nesting won’t work):

```
callback_map = [ [ "ociVersion", cj_ociVersion ],
                 [ "process", [ "env", cj_env ] ] ]
```

Notably, the `cJSON` type is singly linked and doesn’t include pointers to parent, which would be useful.

bin/cjson-parser.c
```diff
@@ -0,0 +1,106 @@
...
+    json = json->child;
+
+    while (json != NULL) {
+        if (strcmp(json->string, "ociVersion") == 0) {
+            // validate oci Version
+            printf("ociVersion: %s\n", json->valuestring);
+        } else if (strcmp(json->string, "process") == 0) {
+            // child: env - set environment to PATH value, cwd corresponds to --cd
+            cJSON *env_key = cJSON_GetObjectItemCaseSensitive(proc_key, "env");
+            printf("process env: %s\n", env_key->child->valuestring);
```
reidpr, line 30: This seems risky because `env_key->child` could be a null pointer if the `config.json` is malformed.

reidpr (commented):
Krishna, great work here! I think we might want to alter the approach a bit, and some suggestions in-line on that, but this is a solid start that avoids some common pitfalls.

bin/ch-run.c
```diff
@@ -63,6 +64,7 @@ const struct argp_option options[] = {
    { "test",          -17, "TEST", 0, "do test TEST" },
    { "mount",         'm', "DIR",  0, "SquashFS mount point"},
    { "no-passwd",      -9, 0,      0, "don't bind-mount /etc/{passwd,group}"},
+   { "oci",           'o', 0,      0, "run as an oci bundle" },
```
kchilleri, line 67: Add command line option to specify OCI bundle

bin/ch-run.c
```diff
@@ -208,6 +245,18 @@ int main(int argc, char *argv[])
       Tf (!args.c.writable || args.unsafe,
           "--write invalid when running by name");
       break;
+   case IMG_OCI_BUNDLE:
+      /* Parse config.json file -> containerize */
+      char *rootfs = path_join(args.c.img_ref, "rootfs");
+      char *conff = path_join(args.c.img_ref, "config.json");
+
+      Tf (path_exists(rootfs, NULL, true), "%s: No such file or directory", rootfs);
+      Tf (path_exists(conff, NULL, true), "%s: No such file or directory", conff);
```
kchilleri, line 254: Check OCI bundle path provided contains ```config.json``` file and ```rootfs``` directory. Make sure they both exist

bin/ch-run.c
```diff
@@ -646,3 +699,127 @@ char *storage_default(void)
 
    return storage;
 }
+
+void readJSONFile(const char *filename, struct args *args) {
```
kchilleri, line 703: Read and parse ```config.json``` file found

bin/ch-run.c
```diff
@@ -646,3 +699,127 @@ char *storage_default(void)
...
+            cJSON_ArrayForEach(elem, subtree)
+                dispatch(actions[i], elem, args);
+        } else {
+            dispatch(actions[i], subtree, args);
+        }
+    }
+}
+
+void oci_ociVersion(cJSON *json, struct args *args) {
+   // Need to validate this version. Not sure how just yet.
```
kchilleri, line 776: How should I validate the OCI version?

bin/ch-run.c
```diff
@@ -646,3 +699,127 @@ char *storage_default(void)
...
+   // TO DO: corresponds to --set-env
+   printf("oci_env: %s\n", json->valuestring);
+}
+
+
+void oci_cwd(cJSON *json, struct args *args) {
+   // corresponds to --cd
+   //printf("oci_cwd: %s\n", json->valuestring);
+   if (args->initial_dir == NULL) {
+      args->initial_dir = json->valuestring;
```
kchilleri, line 791: I assign the corresponding values found on the ```config.json``` file. But these values are not stored/saved to use outside of these functions. I believe it has something to do with the way I'm using pointers (most likely, incorrectly).

bin/ch-run.c
```diff
@@ -646,3 +699,127 @@ char *storage_default(void)
...
+}
+
+
+void oci_path(cJSON *json, struct args *args) {
+   //printf("oci_path: %s\n", json->valuestring);
+
+   args->c.newroot = json->valuestring;
+
+   if (!path_exists(args->c.newroot, NULL, true)) {
+      args->c.newroot = path_join(args->c.img_ref, json->valuestring);
```
kchilleri, line 804: This assignment works, but I believe it's because I'm using the function ```path_join()``` and it's storing the pointer values the right way.

bin/ch-run.c
```diff
@@ -646,3 +699,127 @@ char *storage_default(void)
...
+    if (cJSON_IsTrue(json)) {
+        bool_to_str = "true";
+    } if (cJSON_IsFalse(json)) {
+        bool_to_str = "false";
+    }*/
+
+    //printf("oci_readonly: %s\n", bool_to_str);
+
+   if (cJSON_IsFalse(json)) {
+      args->c.writable = true;
```
kchilleri, line 823: This assignment also works outside of this function. Could be because I'm assigning a boolean instead of a char?

reidpr (commented):
Nice progress! Discussed offline.

kchilleri:
TO DO: 
1. Add documentation
2. Perhaps more error checking

reidpr (commented):
Solid! Looks like the actual logic isn’t too complicated.

The major missing feature might be priority (defaults, environment variables, cmd line, now `config.json`). One way to approach that might be to create an arguments struct for each of those, then merge.

I think the path forward on this might be to table it for a little while until my CDI stuff is merged — there’s some duplicate code.

PR #1882, opened 2024-03-25
fix `$CFLAGS` duplication that we’ve had since 2019

reidpr:
We’ve had duplicate flags in `$CC` invocations for a long time, I *think* since the introduction of Autotools in PR 589.

This is because `ch_run_CFLAGS` is *additional* flags, which I didn’t realize. This PR removes the duplication.

We also state `$CFLAGS` in `configure` and remove an unused variable there.


lucaudill (approved):
LGTM

Issue #1847, opened 2024-02-28
Spack build of `charliecloud@0.35.0+squashfuse` fails to find `ll.h`


cameronrutherford:
Not really sure what's going on here, as I have had success installing charliecloud in the past, but this bug seems be new and persistent/reproducible:

<details>
  <summary>Spack concretization</summary>
  
```
==> Concretized charliecloud+squashfuse
 -   7y3yp4q  charliecloud@0.35%gcc@11.2.0~docs+squashfuse build_system=autotools arch=linux-centos7-zen2
[+]  2al52uo      ^autoconf@2.72%gcc@11.2.0 build_system=autotools arch=linux-centos7-zen2
[e]  3enz5ve          ^perl@5.26.0%gcc@11.2.0+cpanm+opcode+open+shared+threads build_system=generic patches=0eac10e,8cf4302 arch=linux-centos7-zen2
[+]  iq3awo2      ^automake@1.16.5%gcc@11.2.0 build_system=autotools arch=linux-centos7-zen2
[+]  3wkhnll      ^gcc-runtime@11.2.0%gcc@11.2.0 build_system=generic arch=linux-centos7-zen2
[e]  hymxaj5      ^git@2.37.3.dirty%gcc@11.2.0+man+nls+perl+subtree~svn+tcltk build_system=autotools arch=linux-centos7-zen2
[+]  ynvrnsk      ^gmake@4.4.1%gcc@11.2.0~guile build_system=generic arch=linux-centos7-zen2
[+]  y2ubeek      ^libtool@2.4.7%gcc@11.2.0 build_system=autotools arch=linux-centos7-zen2
[+]  jsju533          ^findutils@4.9.0%gcc@11.2.0 build_system=autotools patches=440b954 arch=linux-centos7-zen2
[+]  tyus3ia      ^m4@1.4.19%gcc@11.2.0+sigsegv build_system=autotools patches=9dc5fbd,bfdffa7 arch=linux-centos7-zen2
[+]  3ei4ld3          ^diffutils@3.9%gcc@11.2.0 build_system=autotools arch=linux-centos7-zen2
[+]  tp2gtwt              ^libiconv@1.17%gcc@11.2.0 build_system=autotools libs=shared,static arch=linux-centos7-zen2
[+]  s6uah56          ^libsigsegv@2.14%gcc@11.2.0 build_system=autotools arch=linux-centos7-zen2
[...]

PR #1883, opened 2024-03-26
Spack example: use external `packages.yaml`, add patch

j-ogas:
This PR adds a number of changes.

1. add `packages.yaml`; tell spack to make use of external (container supplied) packages, including libfuse for `+squashfuse1` variant.
2. use spack envrionments to make use of the parallel building feature.
3. add `libfuse.patch`, see https://github.com/spack/spack/pull/43374. This effectively resolves https://github.com/hpc/charliecloud/issues/1847.

This makes the Spack example take 4 minutes and 40 seconds from start to finish, with `--no-cache`.

### DIFF
--- a/examples/spack/Dockerfile
+++ b/examples/spack/Dockerfile
@@ -13,20 +13,29 @@ FROM almalinux:8
 # Spack’s m4, and that package is in PowerTools, which we enable using sed(1)
 # to avoid installing the config-manager DNF plugin.
 #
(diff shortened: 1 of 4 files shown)


j-ogas:
Since it now takes under 5 minutes we could change the scope to `standard`.

reidpr:
If we build against Spack develop we would catch problems much faster.

PR #1885, opened 2024-03-27
fix ObsPy test

reidpr:
ObsPy 1.4.0, which is the current version, is [not compatible](https://github.com/obspy/obspy/issues/3313#issuecomment-1818165937) with Python 3.12. The default Python version in Miniconda recently became Python 3.12, so this test stopped working.

This PR fixes the Miniconda version at 24.1.2 with Python 3.11 to get the test going again for now.


reidpr:
Internal tests pass in full scope.

lucaudill (approved):
LGTM

PR #1887, opened 2024-03-27
`configure.ac`: add missed `fnctl.h`

reidpr:
`open(2)` is [documented](https://man7.org/linux/man-pages/man2/open.2.html) to require `fnctl.h`. We weren’t including it explicitly, but it worked, I assume because we got `fnctl.h` indirectly. However this is apparently not true on Ubuntu 22.04, and regardless we should include the documented file.


lucaudill (approved):
LGTM

PR #1899, opened 2024-04-29
`ch-test`: build and unpack image only

### LINKED ISSUE #1748: add test feature that builds and unpacks one image
Many tests require that a specific image be present. If it’s not, then building and unpacking it can be a hassle. The new `build-images` phase from #1559 helps because it reduces the time to build *all* images, but it would be nice if we could simply "make this image ready for testing".

lucaudill:
closes #1748

### DIFF
--- a/bin/ch-test
+++ b/bin/ch-test
@@ -669,6 +669,10 @@ test_one_file () {
     bats "$one_file"
 }
 
+#build_one_image () {
+#
+#}
+
 test_rootemu () {
     if command -v ch-image > /dev/null 2>&1; then
         bats force-auto.bats
@@ -854,6 +858,11 @@ while [[ $# -gt 0 ]]; do
         ;;
     -h|--help)
         usage 0
+        ;;
+    -i)
+        phase=one-file
+        one_file="${TMP_}/build_auto.bats:$1"
+        shift
         ;;
     --img-dir)
         require_unset imgdir
--- a/test/common.bash
+++ b/test/common.bash
@@ -162,6 +162,28 @@ image_ok () {
 
 }
 
+# Function to determine whether first argument is in the subsequent expanded
+# array or sequence of words.
+
+# Example:
+#   $ is_elem "foo" "foo" "bar" "baz"
+#   0
+#
+#   $ is_elem "banana" "foo" "bar" "baz"
+#   1
+is_elem () {
+    local test_word="$1"
+    shift
+    local word_array=("$@")
+    for e in "${word_array[@]}"
+    do
+        if [[ "$e" == "$test_word" ]]; then
+            return 0
+        fi
+    done
+    return 1
+}
+
 localregistry_init () {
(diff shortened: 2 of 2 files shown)


bin/ch-test
```diff
@@ -88,6 +88,8 @@ EOF
 
 bats_vmin=1.2.0  # FIXME: get from configure somehow?
 
+ch_build_unpack_all="argenv quick copy almalinux_8ch debian_11ch hello multistage seccomp"
```
reidpr, line 91: Should derive this list from the tests we actually have; it should also be responsive to `--scope`.

bin/ch-test
```diff
@@ -197,6 +199,9 @@ dir_tar_check_file () {
     for f in "$@"; do
         if [[ -f ${CH_TEST_TARDIR}/${f} ]]; then
             return 0
+        elif [[ "$f" == *"chtest"* ]] && [[ -n "$ch_build_unpack_list" ]]; then
+            # chtest not needed for build-unpack phase
```
reidpr, line 203: I think this is not true — why treat `chtest` different from any other image?

bin/ch-test
```diff
@@ -945,6 +967,8 @@ while [[ $# -gt 0 ]]; do
     esac
 done
 
+export ch_build_unpack_list
```
reidpr, line 970: Subprocesses won't need this; are you sure it needs to be exported?

bin/ch-test
```diff
@@ -978,7 +1002,7 @@ if [[ $phase = one-file ]]; then
 fi
 
 printf "%-21s %s" 'phase:' "$phase"
-if [[ $phase = one-file ]]; then
+if [[ $phase = one-file || $phase = build-unpack ]]; then
```
reidpr, line 1005: I wonder if there's a reasonable way to make one-file interact with build-unpack. That is, rather than listing images with `-i`, build-unpack builds and unpacks all images *except* if `-f` is given. I think `-f` does have some smarts about which tests indicate an image?

bin/ch-test
```diff
@@ -1168,6 +1192,24 @@ case $phase in
         test_build
         win
         ;;
+    build-unpack)
```
reidpr, line 1195: I wonder if rather than making build-unpack a phase, could it be a modifier on build and run? Or, perhaps there's a new phase `unpack`, and `build-unpack` implies `build-images` and `unpack`? Or, maybe `build-images` goes away and is replaced by running `build` with the build-unpack modifier?

reidpr (commented):
Good start. Some ideas in-line.

bin/ch-test
lucaudill: It's my understanding that the `export` is necessary for this variable to be available to other files, e.g. `common.bash`, where we decide whether or not to skip a test. I'm basically copying what we do with `ch_one_test`.

Issue #1210, opened 2021-10-21
use `mksquashfs -quiet` if available


reidpr:
`mksquashfs(1)` is super noisy. Newer versions have `-quiet`. This issue is to use that switch if supported. We could probably grep `mksquashfs -?`.

PR #1920, opened 2024-09-09
Mksquashfs quiet

alstar555:
Fixes issue #1210 
Send output to /dev/null instead of using -quiet so that mksquashfs is quiet on all versions, not only 4.4 and above.


reidpr:
closing in favor of #1921 per discussion

PR #1921, opened 2024-09-09
Mksquashfs quiet using '-quiet' flag

alstar555:
Fixes issue #1210 
Uses the '-quiet' flag on mksquashfs. 
If the grep cmd is able to find the '-quiet' flag in the mksquashfs help menu, then we use the '-quiet' flag because this flag is only available on mksquashf version 4.4 or higher.

### DIFF
--- a/bin/ch-convert
+++ b/bin/ch-convert
@@ -116,8 +116,13 @@ cv_dir_squash () {
     # Exclude build cache metadata. 64kiB block size based on Shane’s
     # experiments.
     # shellcheck disable=SC2086
-    quiet mksquashfs "$1" "$2" $squash_xattr_arg -b 65536 -noappend -all-root \
+    if mksquashfs --help 2>&1 | grep "quieti"; then 
+       mksquashfs "$1" "$2" $squash_xattr_arg -b 65536 -noappend -all-root \
+                         -pf "$pflist" -e "$1"/ch/git -e "$1"/ch/git.pickle -quiet
+    else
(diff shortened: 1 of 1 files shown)


bin/ch-convert
```diff
@@ -116,8 +116,14 @@ cv_dir_squash () {
     # Exclude build cache metadata. 64kiB block size based on Shane’s
     # experiments.
     # shellcheck disable=SC2086
-    quiet mksquashfs "$1" "$2" $squash_xattr_arg -b 65536 -noappend -all-root \
-                         -pf "$pflist" -e "$1"/ch/git -e "$1"/ch/git.pickle
+    if mksquashfs --help 2>&1 | grep "quiet"; then
```
reidpr, line 119: Here I think you want to tighten up the `grep(1)` a bit, specifically:

```suggestion
    if mksquashfs --help 2>&1 | egrep -q '^-quiet'; then
```

the changes being:

1. `grep` → `egrep`: use “extended” regular expressions which really just means “horrible syntax” instead of “incomprehensibly horrible syntax”. In my view one should only ever `fgrep(1)` for string matching or `egrep(1)` for regex matching.
2. The carat means match only at the beginning of a line.
3. Match the option itself, without false positives if the help says the word “quiet” somewhere.
4. `-q`: Don’t print matching lines, since we only care about the presence or absence of a match.
5. Single quotes because we don’t want to expand shell variables.

reidpr (commented):
Discussed offline. Thanks Angelica!

PR #1924, opened 2024-09-16
Retire support for storage v6 

### LINKED ISSUE #1683: retire storage prior to 0.33 (v6) on 2024-06-10
Storage version 7 was introduced in version 0.33, released June 9, 2023. Per #1507, this issue is to remove support for storage v6.

alstar555:
Fixes #1683

### DIFF
--- a/lib/filesystem.py
+++ b/lib/filesystem.py
@@ -1175,7 +1175,7 @@ class Storage:
       if (v_found == STORAGE_VERSION):
          ch.VERBOSE("found storage dir v%d: %s" % (STORAGE_VERSION, self.root))
          self.lock()
-      elif (v_found in {None, 6}):  # initialize/upgrade
+      elif (v_found in {None}):  # initialize/upgrade
          ch.INFO("%s storage directory: v%d %s"
                  % (op, STORAGE_VERSION, self.root))
          self.root.mkdir()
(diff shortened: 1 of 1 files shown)


lib/filesystem.py
```diff
@@ -1175,7 +1175,7 @@ def init(self):
       if (v_found == STORAGE_VERSION):
          ch.VERBOSE("found storage dir v%d: %s" % (STORAGE_VERSION, self.root))
          self.lock()
-      elif (v_found in {None, 6}):  # initialize/upgrade
+      elif (v_found is None):  # initialize/upgrade
```
reidpr: I'd like to leave the set. Right now we're in an odd situation where literally no old storage versions are upgradeable, but I expect that will change in the future. So the set notation will make the intent clearer. Also my comment is kind of bad. For example:

```suggestion
      elif (v_found in {None,}):  # None if initializing, or upgraded version
```

lib/filesystem.py
```diff
@@ -1294,10 +1284,8 @@ def validate_strict(self):
                      % (msg_prefix, img), ch.BUG_REPORT_PLZ)
 
    def version_read(self):
-      # While support for storage v1 was dropped some time ago, let’s at least
-      # retain the ability to recognize it.
       if (not os.path.isfile(self.version_file)):
-         return 1
+         ch.FATAL('Expected file "%s" not found' % self.version_file)
```
reidpr: What if we instead omit this whole check, and add the version file to `valid_p` per #1147, which you would then add to this PR?

reidpr (changes requested):
Good job Angelica. A couple suggestions in-line.

